PR #35, opened 2017-12-19
Garden sentence and "encountered" gone

comradekingu:


app/src/main/res/values/strings.xml
```diff
@@ -19,13 +19,13 @@
...
     <string name="nc_display_name_stored">Display name stored</string>
-    <string name="nc_display_name_not_stored">Display name not stored, aborting</string>
+    <string name="nc_display_name_not_stored">Could not store display name, aborting</string>
 
     <string name="nc_never">Never joined</string>
     <string name="nc_search">Search</string>
 
     <string name="nc_certificate_dialog_title">Check out the certificate</string>
-    <string name="nc_certificate_dialog_text">Previously unknown SSL certificate encountered, issued by %1$s for %2$s valid from %3$s to %4$s. Trust it?</string>
+    <string name="nc_certificate_dialog_text">Do you trust hitherto unknown SSL certificate, issued by %1$s for %2$s, valid from %3$s to %4$s?</string>
```
mario: hiterto? What does that means?

tobiasKaminsky: Something like "previous" or "the one before that".

mario: I see. I put myself among highly proficient in English, yet I've never heard about this earlier. Every day is a good day to learn :D Can we however turn this into something more "simple"?

mario:
Thanks.

PR #36, opened 2017-12-19
Fix transifex

tobiasKaminsky:
Transifex is using values-es_419 for spanish "latin america", but this is not supported on android. This changes "values-es_419" to "values-es-rUS", which works.


mario:
Thanks.

PR #45, opened 2017-12-27
launcher icon including dotted background + source files

AndyScherzinger:
for regeneration the foreground logo need to be set to 80% in the asset generator.

..as discussed on IRC.


mario:
You are awesome, the best ever icon maker in the known universe.

PR #46, opened 2017-12-27
Display participants via flexbox layout

AndyScherzinger:
maybe a work in progress but I open the PR just in case so it is on the radar/visible during the time I can't further work on it in case somebody wants to pick up where I paused. :)

Details/documentation for the used flexbox layout: https://github.com/google/flexbox-layout

### DIFF
--- a/app/src/main/res/layout/activity_call.xml
+++ b/app/src/main/res/layout/activity_call.xml
@@ -26,15 +26,16 @@
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
                 android:fitsSystemWindows="true"
+                android:background="#000000"
                 tools:context=".activities.CallActivity">
 
     <com.google.android.flexbox.FlexboxLayout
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
@mario seems to be doing what you needed to be done:

device-2017-12-28-172850

device-2017-12-28-172926

Beware though that the flex box (like the linear layout before) has the own camera image (me in this case) hovering above the surface render items...

AndyScherzinger:
btw. I added some spacing via paddings which can just be removed in case you don't want them to be spaced but rather glued together.

AndyScherzinger:
@mario I think we can close this one, right?

mario:
@AndyScherzinger probably :-/ We need to see how to change the layout which is what I'll discuss with you via some kind of video chat or in person. Sorry about that!

AndyScherzinger:
All good @mario :+1: No worries!

PR #68, opened 2018-01-18
Odd newlines cleaned up

comradekingu:
Previously looked like this on TX:
"Permissions need to be granted to establish a video and/or audio call. Please
       click \"ALLOW\" in the upcoming system dialog."


mario:
Thanks.

PR #92, opened 2018-02-08
Fix spelling of LICENSE file

ThyMYthOS:


mario:
Meh. British vs US, I guess?

ThyMYthOS:
Right, just looked strange, but seems valid.

mario:
Actually, here's a better explanation:)

https://www.macquariedictionary.com.au/blog/article/100/

PR #123, opened 2018-03-22
Update README.md

Poussinou:
Hi,

This small and useless PR adds two buttons to get your app on F-Droid and Google Play 👍


mario:
Awesome! One question though: Why is the start of the Gplay button not aligned with the text above and below?

mario:
@Poussinou poke! :)

Poussinou:
Sorry I was off for some days xD I don't really know, it's allays like that, as you can see [here](https://github.com/nextcloud/android/blob/master/README.md) or [here](https://github.com/LawnchairLauncher/Lawnchair/blob/alpha/README.md) for example...

mario:
Thanks, you're awesome!

PR #125, opened 2018-03-27
Update README.md

jospoortvliet:
Update the text ;-)

Can make further changes if you think it makes sense, I just missed an answer to the "why is this so awesome" question :D


mario:
👍

Issue #100, opened 2018-02-21
Contacts list dividers


jancborchardt:
- They shouldn't be colored Nextcloud blue, it's too much and breaks the list. Instead just white
- The letter should left-align with the people's names, so move a bit to the right
- Can the divider be a bit transparent/blur elements behind, like on iOS or our server in the file list?

mario:
- The divider is never blue. The header of a contacts group is, however.
- The current design is standard. There's a reason why header text starts earlier. Just like when you do lists and sublists, you use indentation.
- What divider? The header or one that divides the items? In either case, not convinced since this is not standard across Android.

PR #147, opened 2018-04-26
redesign the contacts header/contact items

AndyScherzinger:
My take on  #100 by going for the Android Oreo contacts list style.
* Category left aligned 16dp
* Contact items further to the left
* Divider is only set for the categories, not between the items themselves
--> This way the Category will stay at the far left when being sticky :)

List at the top:
device-2018-04-26-104045

When being (slightly scrolled... don't have many contacts)
device-2018-04-26-104130


mario:
Thanks for the awesome contribution.

PR #148, opened 2018-04-26
extract new call item to layout

AndyScherzinger:
@mario like I said on IRC I can't compile the code anymore thanks to dagger failing. So this is a completely blind implementation :(


mario:
@AndyScherzinger thanks for working on this.

PR #149, opened 2018-04-27
trying to manually tint bottom sheet navigation items

AndyScherzinger:
unsuccessful....

Hi @mario here are my tries to get this done but, no way found to get it working... I can deactivate the auto-tintings, then I see the avatar (else it'll be just a grey/blue circle since it gets auto tinted...). but then no icon in the bottom bar gets tinted.

We had the same issue in the drawer for the files app which is why I implemented manual state/tinting handling but for whatever reason this doesn't work for the bottom navigation view, I tint the icons but it doesn't show up on the view ?!...

Atm I have no clue how to implement this design request :/


AndyScherzinger:
Housekeeping: I'll close the issue for now but leave the branch untouched

PR #150, opened 2018-04-27
Adding icons to the call menu

### LINKED ISSUE #99: Call menu design
- Can we place the menu just where the menu icon is instead of at the bottom? The menu shouldn't be in a different plce than where was clicked. Also compare to Files Android app.
- Then we can also drop the irrelevant header "What would you like to do today?"
- "Leave call" should always be the last item in the list, and colored red
- Can we get icons to the left of the items? Rename pencil, Share icon, Delete icon

AndyScherzinger:
Fixes #99

* icons added
* font size bumped to 18sp (typical bottom sheet font size)
* put "leave conversation" at the bottom

device-2018-04-27-182647


AndyScherzinger:
A) materialdesignicons.com
B) erm, where do I find that part? :)

mario:
CallMenuController, onItemClick :)

mario:
Thanks for the contribution. I fixed the actions and will merge this now.

jancborchardt:
Nice, thanks @AndyScherzinger! :) Two more things:
- Can we drop the unneeded header? Too much text which is not useful
- Is it not standard to have a cancel entry on the bottom too?

AndyScherzinger:
You're welcome @jancborchardt 

> * Can we drop the unneeded header? Too much text which is not useful

Sure thing

> Is it not standard to have a cancel entry on the bottom too?

Bottom sheets are alternative menus so they usually don't contain an explicit cancel. This can be done by clicking "somewhere else on the screen" or hitting the back button.

AndyScherzinger:
Created an issue for the header removal: #151

PR #154, opened 2018-04-30
Fix bottom sheet background color

AndyScherzinger:
now the items have a white background. For whatever reason setting the background color on any parent element doesn't work...


AndyScherzinger:
device-2018-04-30-145641

PR #194, opened 2018-06-13
add info about install script

enoch85:
Feedback is welcome!


nickvergessen:
This is the android app... I don't think we need general install instructions for the php app there.

mario:
I told him to do this back in the time. Did you add it to the spreed repo then?

enoch85:
> Did you add it to the spreed repo then?

Yup: https://github.com/nextcloud/spreed#installing

PR #196, opened 2018-06-14
revamp call notification screen 

AndyScherzinger:
with 
* dark theme and 
* constraint layout

call

Beware that the system bar will still be blue. In order to get that part in black one would need to set another theme like `NoActionBar`

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -17,7 +17,7 @@
     <color name="nc_light_grey">#E8E8E8</color>
     <color name="grey_600">#757575</color>
     <color name="nc_grey">#D5D5D5</color>
-    <color name="white30">#e9ffffff</color>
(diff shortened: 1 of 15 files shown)


app/src/main/res/values/colors.xml
```diff
@@ -17,5 +17,6 @@
     <color name="nc_light_grey">#E8E8E8</color>
     <color name="grey_600">#757575</color>
     <color name="nc_grey">#D5D5D5</color>
+    <color name="white30">#e9ffffff</color>
```
mario: Upper case.

AndyScherzinger:
lightblack

AndyScherzinger:
lighter black/grey tone

mario:
Thanks mate, as always!

mario:
Will wait for build to turn green and merge.

PR #198, opened 2018-06-18
Fix call notification layout for landscape

AndyScherzinger:
* slightly lower the call notification avatar size, 
* optimize for layout to work for portrait and landscape

The avatar is still quite large in landscape, so the `dimens` value could be lowered to `140dp` or `120dp` and should still look okay fort portrait.

portrait
landscape


mario:
140dp in portrait, and 120 in landscape? :D

AndyScherzinger:
> 140dp in portrait, and 120 in landscape? :dagger: 

Will that work with your avatar fetching?

AndyScherzinger:
We could of course declare portrait, landscape and fetching dimens values, with portrait==fetching size

mario:
If you load avatar in onAttach and set size depending on orientation.

PR #215, opened 2018-06-27
Fix wording details

jancborchardt:
When using the 3-dot-menu for a conversation, it only needs to show the conversation name in the title. "Configure conversation" before that is not necessary.

(Also detail fix »Change a password« → »Change password«)


mario:
*shrug* cc @AndyScherzinger

jancborchardt:
@mario would it really kill you to answer with a short friendly "thank you! :)" instead? ;)

PR #302, opened 2018-08-31
improve performance

o0lwj0o:
see https://stackoverflow.com/questions/44334233/why-is-the-stringbuilder-chaining-pattern-sb-appendx-appendy-faster-than-reg/44338950


mario:
Thanks for the contribution.

PR #331, opened 2018-10-22
Height of conversations in list should all be the same

### LINKED ISSUE #320: Height of conversations in list should all be the same
Currently in the conversation list there are 2 different kinds of height:
- Conversations with 1 row of text as message preview
- Conversations with 2 rows of text as message preview

This looks strange when scrolling when the avatars on the left are differently spaced. We should adjust it so the height is the same for all rows.

AndyScherzinger:
Resolves #320 - setting a fixed line number of 2

+ aligning avatar and unread count at the top
+ adding IDE preview rendering infos

Not 100% sure this is what has been described in #320 and if it covers all areas referred to in the issue so please have a look :)

_______

2018-10-22 12_19_30-talk-conversationheight c__dev_src_android_nextcloud_talk-conversationheight -

### DIFF
--- a/app/src/main/res/layout/rv_item_conversation_with_last_message.xml
+++ b/app/src/main/res/layout/rv_item_conversation_with_last_message.xml
@@ -24,6 +24,7 @@
   -->
 
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+                xmlns:tools="http://schemas.android.com/tools"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content">
 
@@ -42,7 +43,8 @@
             <ImageView
                 android:id="@id/dialogAvatar"
                 android:layout_width="@dimen/avatar_size"
-                android:layout_height="@dimen/avatar_size"/>
+                android:layout_height="@dimen/avatar_size"
+                tools:src="@drawable/ic_call_black_24dp"/>
 
             <ImageView
                 android:id="@+id/passwordProtectedRoomImageView"
@@ -71,7 +73,8 @@
             android:includeFontPadding="false"
             android:maxLines="1"
             android:textColor="@color/nc_incoming_text_default"
-            android:textSize="16sp"/>
+            android:textSize="16sp"
+            tools:text="Best conversation"/>
 
         <TextView
             android:id="@id/dialogDate"
@@ -95,8 +98,8 @@
                 android:id="@id/dialogLastMessageUserAvatar"
                 android:layout_width="@dimen/small_item_height"
                 android:layout_height="@dimen/small_item_height"
-                android:layout_centerVertical="true"
-                android:layout_marginEnd="8dp"/>
+                android:layout_marginEnd="8dp"
+                tools:src="@drawable/ic_call_black_24dp"/>
 
             <TextView
                 android:id="@id/dialogLastMessage"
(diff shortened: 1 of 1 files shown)


mario:
Thanks for your contribution.

jancborchardt:
> aligning avatar and unread count at the top

Vertically centering those on the text would look a bit better. :)

AndyScherzinger:
> Vertically centering those on the text would look a bit better. :)

Not sure how that could be implemented right away since the vertical alignment can only be made in relation to the size of the textview which now is 2-lines, so it would only work if the text fills up 2 lines, If the text is only 1 line the icons will hang at the bottom 50% below the text.
It has been like that be 2-lines-size has been introduced so this is basically the trade-off to have the conversation items having the same size :/

AndyScherzinger:
To make my point a bit clearer here are two screenshots to show what I mean:

**single line message / icons centered**
conv1_centered

**with this merged PR: single line message / icons aligned top**
conv1


I am fine either way, I just need to know and can create another PR for it, no big deal

AndyScherzinger:
I opened #332 just in case you decide to go for vertically centered

jancborchardt:
@AndyScherzinger is there no way to have it without the trade-offs? Cause the original issue was about the line-height of the whole row – and it was fixed by forcing 2 rows on the preview.

What if instead you force a height on every row, and the content stays the same (vertically centered of course)?

AndyScherzinger:
> What if instead you force a height on every row, and the content stays the same (vertically centered of course)?

hmm, kind of yes, that could be established with some trade-offs... 
I can't exactly calculate the height of a two line textview for various reasons like sp/dp, text-alignments/overflows so I'd have to choose a height that definitely fits the whole thing so the bottom spacing of the item might not be the super-exact-pixel-perfect spacing at the bottom we have for the top spacing. If that is fine then I can give that one a try :)

jancborchardt:
@AndyScherzinger yeah, sounds hacky too. :D Just another thought: For simplicity we could also just go for 1 line of preview text? cc @mario

mario:
@jancborchardt in most cases 1 line of preview text doesn't help at all as it's usually just 2-3 words, depending on the language. I've evaluated this closely and looked at the other apps and took users feedback into consideration, and I'd really like it to stay at 2.

How you decide to format it is a different story - I'm more than open to your suggestions (and @AndyScherzinger will implement :P) :)

AndyScherzinger:
> @AndyScherzinger yeah, sounds hacky too. :D Just another thought: For simplicity we could also just go for 1 line of preview text? cc @mario

Yes, technically a single line is easy to achieve if that is what we want.

AndyScherzinger:
Argh, answered too late.

Looking at the standard SMS app:
* items with unread messages are sorted at the top with up to 3 lines of message content and bold
* items that are completely read are sorted below with single line message content and not-bold

Other than that, just to completely ague against vertically centered, looking at the types section of https://material.io/design/components/lists.html#usage multi line items are basically "top aligned" when it comes to the content, so we not just top align everything, especially the left, large avatar like:

**2-line** 
2018-10-22 12_19_30-talk-conversationheight c__dev_src_android_nextcloud_talk-conversationheight -

**single line**
conv1

**I could even argue for this:**

 The bottom text is the last message, the bubble show the amount of unread messages, so not directly related to the distinct message show, just to the type of info (messages). While I could argue that the counter is a meta-information of the conversation itself thus belonging to the headline/conv-name ;)

PR #378, opened 2018-12-10
Fix a few crashes

mario:
Signed-off-by: Mario Danic <mario@lovelyhq.com>

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
@@ -161,15 +161,17 @@ public class ConversationInfoController extends BaseController {
     }
 
     private void setupAdapter() {
-        if (adapter == null) {
+        if (adapter == null && getActivity() != null) {
             adapter = new FlexibleAdapter<>(recyclerViewItems, getActivity(), true);
         }
 
-        SmoothScrollLinearLayoutManager layoutManager =
-                new SmoothScrollLinearLayoutManager(getActivity());
-        recyclerView.setLayoutManager(layoutManager);
-        recyclerView.setHasFixedSize(true);
-        recyclerView.setAdapter(adapter);
+        if (recyclerView != null && getActivity() != null) {
+            SmoothScrollLinearLayoutManager layoutManager =
+                    new SmoothScrollLinearLayoutManager(getActivity());
+            recyclerView.setLayoutManager(layoutManager);
+            recyclerView.setHasFixedSize(true);
+            recyclerView.setAdapter(adapter);
+        }
     }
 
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
👍 code wise it looks good to me :)

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
```diff
@@ -161,15 +161,17 @@ protected void onViewBound(@NonNull View view) {
     }
 
     private void setupAdapter() {
-        if (adapter == null) {
+        if (adapter == null && getActivity() != null) {
```
tobiasKaminsky: You could assign getActivity to a variable and check it only once in preceding if.
This might be a bit better to read:

```
Activity activity = getActivity();
if (activity != null) {
 	if (adapter == null) {}

	if (recyclerView != null) {}
}
```

tobiasKaminsky:
:+1:

tobiasKaminsky:
@mario drone is failing?

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
mario: I thought about this. What if Activity disappears between the first place where I need it and the second one? :)

Unlikely, sure, but possible? :)

mario:
@tobiasKaminsky one warning more in this PR. Will check and fix.

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
tobiasKaminsky: As there are no time intensive operation in it, I think there is a very little chance to have this, but of course it could happen. 
I still would bet that you never see a crash regarding this ;-)

AndyScherzinger:
👍

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
mario: It's not about it being time intensive, but the user might interrupt the app. Anyway, I do agree it's unlikely like I mentioned above. Will change it.

AndyScherzinger:
@mario it is an extra warning due to the dependencies:
```
../../../build.gradle:111: A newer version of android.arch.work:work-runtime than 1.0.0-alpha11 is available: 1.0.0-alpha12
```

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.java
mario: Done post-merge!

PR #403, opened 2019-01-07
Add screensharing support

mario:
Signed-off-by: Mario Danic <mario@lovelyhq.com>


mario:
- [ ] - handle nick changes with MCU
- [ ] - handle screenUnshare with MCU

PR #458, opened 2019-02-13
Update strings.xml

Valdnet:
I'm proposing new names. I think that names are more clever for users.


mario:
The problem is that the "One-to-one calls" sound really only gets in effect when one person calls you. Group and public calls would fall under "Other" so I'm not sure if "Calls" makes sense since it's not really true :/

Valdnet:
> Group and public calls would fall under "Other" so I'm not sure if "Calls" makes sense since it's not really true :/

In my opinion, this is unclear to the user. The options are very misleading.

"Other" - only the short notification sound is used for these calls.
At group and public calls, this sound is inappropriate - too short and misleading with notifications.

It can change the settings:
1. Sound
* Calls
* One-to-one calls
* Notifications

or
1. Sound
* Calls
* Notifications
2. One-to-one sound
* Calls
* Notifications

Please, think about it.

mario:
Thank you for your contribution!

PR #493, opened 2019-03-05
Improve mentions

mario:


Ivansss:
screenshot? :)

mario:
screenshot_20190306_104618_com nextcloud talk2 debug

mario:
@Ivansss once you click on a suggested autocomplete, it's replaced by a name. When you try to edit that name in any way (remove a char, add a char, whatever), you CAN, but then it's not a mention anymore and BOLD is removed from it.

When you send a message to the server, "Maser Walter" will be transformed magically to "@maser" in this case.

PR #519, opened 2019-04-09
Adds mention chips

mario:


mario:
Screenshot_20190409_214835_com nextcloud talk2 debug

cc @jancborchardt ^_^

mario:
Screenshot_20190410_142651_com nextcloud talk2 debug

jancborchardt:
Some feedback, sorry it’s a bit late:
- The text in the mention chips is a bit lower than the message text. The baseline should be the same.
- What’s the logic on the coloring? I assume "you" are "Allen Molly" in these examples? So your chip should be the one colored blue, not the others.
- Especially not the chips in the new message input, the blue background is very harsh there as it’s overused. Better would be white with a slight grey border.
- Similar for the white chips in the outgoing messages, the white on blue is too harsh. Better to have a transparent background and grey/white border (partially transparent).

mario:
@jancborchardt 

- sometimes, yes (in multi-line messages) - a bit complex to fix, so I opened an issue for later
- some things I've already changed, but I can look at it further tomorrow

jancborchardt:
@mario cool, thanks! Especially the coloring is important: "blue bubble" always means it’s you, whether it’s a message bubble, a notification @ bubble in the list, or a mention chip (yes, the mention chips of others when _you_ mention them are technically blue, but only because they are in your blue bubble).

PR #522, opened 2019-04-10
utils: use try-with-resources to properly close Closable.

ardevd:
Connections, streams, files, and other classes that implement the Closeable interface or its super-interface, AutoCloseable, needs to be closed after use. Further, that close call must be made in a finally block otherwise an exception could keep the call from being made. Preferably, when class implements AutoCloseable, resource should be created using "try-with-resources" pattern and will be closed automatically.

Failure to properly close resources will result in a resource leak.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/utils/PushUtils.java
+++ b/app/src/main/java/com/nextcloud/talk/utils/PushUtils.java
@@ -134,13 +134,17 @@ public class PushUtils {
     private int saveKeyToFile(Key key, String path) {
         byte[] encoded = key.getEncoded();
 
-        try (FileOutputStream keyFileOutputStream = new FileOutputStream(path)) {
-            keyFileOutputStream.write(encoded);
-
+        try {
             if (!new File(path).exists()) {
-                new File(path).createNewFile();
+                if (!new File(path).createNewFile()) {
+                    return -1;
+                }
             }
-            return 0;
+
+            try (FileOutputStream keyFileOutputStream = new FileOutputStream(path)) {
+                keyFileOutputStream.write(encoded);
+                return 0;
+            }
         } catch (FileNotFoundException e) {
             Log.d(TAG, "Failed to save key to file");
         } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/utils/PushUtils.java
```diff
@@ -128,14 +133,13 @@ public SignatureVerification verifySignature(byte[] signatureBytes, byte[] subje
 
     private int saveKeyToFile(Key key, String path) {
         byte[] encoded = key.getEncoded();
-        FileOutputStream keyFileOutputStream = null;
-        try {
+
+        try (FileOutputStream keyFileOutputStream = new FileOutputStream(path)) {
+            keyFileOutputStream.write(encoded);
```
mario: Hm, shouldn't this go AFTER creating a file?

ardevd: Agreed. Also, we should probably ensure we could actually create the file by checking the return value.

mario: Yes, in theory, but it should mostly work because it's internal storage? :) But feel free to add additional safeguards! :)

ardevd: done

mario: @ardevd you pushed a wrong commit? :)

ardevd:
omg, no idea how I managed to screw that one up. fixed now.

mario:
Sorry to be nitpicking, but any chance of squashing this into one sane commit and force pushing? :) 

Then I can merge! :)

ardevd:
> Sorry to be nitpicking, but any chance of squashing this into one sane commit and force pushing? :)
> 
> Then I can merge! :)

Can't you squash when you merge the PR? :)

mario:
I can. Thanks @ardevd.

PR #529, opened 2019-04-17
controllers: removed identical sub-expression in if statement.

ardevd:
Using the same value on either side of a binary operator is almost always a mistake. Either there is supposed to be a null check for something else in there or the second sub expression should be removed.


mario:
Ha ha, thanks.

PR #532, opened 2019-04-26
gradle: replace use of obsolete 'compile' configuration.

ardevd:
'compile' is obsolete and has been replaced with 'implementation' and 'api'.


mario:
Thanks! :)

PR #546, opened 2019-05-16
Make the unread counter more usable

nickvergessen:


mario:
It's a bit more complex, I don't think 99+ shows properly. But I'll take a look.

mario:
Moved to elipsis, rebased and fix a few issues. Not perfect, but yea...

PR #588, opened 2019-08-03
FIDO U2F Support

dschuermann:
As requested in https://github.com/nextcloud/android/pull/4286

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.java
@@ -22,6 +22,7 @@ package com.nextcloud.talk.controllers;
 
 import android.annotation.SuppressLint;
 import android.content.pm.ActivityInfo;
+import android.graphics.Bitmap;
 import android.net.http.SslCertificate;
 import android.net.http.SslError;
 import android.os.Build;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.java
```diff
@@ -169,6 +175,8 @@ protected void onViewBound(@NonNull View view) {
         webView.clearHistory();
         WebView.clearClientCertPreferences(null);
 
+        webViewFidoBridge = WebViewFidoBridge.createInstanceForWebView((AppCompatActivity) getActivity(), webView);
```
mario: Maybe we should do a check if getActivity() != null?

dschuermann: Hm, if getActivity() is null a lot of other lines would also fail (such as the next line with ``CookieSyncManager.createInstance(getActivity());``).
What about an early exit for the whole ``onViewBound()`` method if getActivity is null?

mario: I don't disagree with you. There's a need to rewrite this whole thing to make it better. So it's probably fine for now :)

mario:
Thanks. One more thing before I can merge it - it looks like your commits are not signed. Could you please do it?

Sorry for the hassle!

dschuermann:
> Thanks. One more thing before I can merge it - it looks like your commits are not signed. Could you please do it?
> 
> Sorry for the hassle!

As the maintainer of OpenKeychain, I concur ;)

mario:
Thank you so much!

jancborchardt:
@dschuermann by the way, we have a "Talk development chat" for all contributors of the app. If you like, we can add you there with a guest account? :)

PR #617, opened 2019-09-04
Signed-off-by: Peter Edens <petere@conceiva.com>

Peterede:
Testing for null on getLobbyState.  Using an older version of nextcloud talk on the server the app will crash on getLobbyState because it returns null and equals is called on it without testing for null.


mario:
Hey, thanks for the PR but this was already fixed a bit differently :) Could you please do a git pull and check if that works for you?

PR #670, opened 2019-11-12
fix CI

tobiasKaminsky:
@nickvergessen pointed out that CI is not working anymore.
This should fix it.


nickvergessen:
Seems to not work completely

tobiasKaminsky:
This was a problem fetching dependencies. I restarted CI

PR #746, opened 2020-03-07
Fix NullPointerException on (first) account creation with client certificates (#195)

stephanritscher:
I was able to reproduce and fix (at least for me) the problem described in #195 

Signed-off-by: Stephan Ritscher <no3pam@gmail.com>

Issue #195, opened 2018-06-13
attempt to invoke virtual method java.lang.String com.nextcloud.talk.models.database.UserEntity.getClientCertificate() on a null object reference


ClCfe:
Hello

i just updated to 2.0.0 from play store
I ve got this error as soon as I enter the server address that has ssl client cert enabled

attempt to invoke virtual method java.lang.String com.nextcloud.talk.models.database.UserEntity.getClientCertificate() on a null object reference

I have android 8.1.0

I have no stack trace (nothing in logcat)

Maybe this is occuring here:

https://github.com/nextcloud/talk-android/blob/1b728a7c4d1928f808420b03745666b7583ee306/app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.java#L233

ClCfe:
Yes cert  is imported in android and i can connect to my nextcloud with chrome
No i haven't configured any user in the app

mario:
Thanks for the clarification.

ClCfe:
Hello

Bug is still here, but is definitely related to not having any user configured

if I connect to a server without client certs, then I can add the 2nd account targetting my server with certs 

Maybe the bug is either in controllers/SettingsController.java
Or
utils/ssl/MagicKeyManager.java

IMbiea:
I'm still seeing this issue as well.

Same error.  I have my CA installed, and client certs (same custom CA) which works with Android/chrome
But doesn't work with the latest client (3.1.1)

intika:
still not solved...

mario:
Should be fixed now :)

PR #747, opened 2020-03-07
Handle client certificate alias if currentUser is null (relates to #195)

stephanritscher:
Unfortunately I couldn't verify the problem or test the fix on this branch. The login GUI seems to be somewhat broken currently due to refactoring process (no option for client certificates showing).
But looking at the code, the master branch has a similar problem in the logic as v.8.0.x which I addressed in PR #746.

Signed-off-by: Stephan Ritscher <no3pam@gmail.com>


mario:
Thanks!

Issue #239, opened 2018-07-16
Add screenshots to readme


jancborchardt:
The readme would look much nicer with some screenshots below the short description. Simply taking the same ones as from the app store and showing them in one row would be great.

(A great example of a nice readme is the Mail app: https://github.com/nextcloud/mail/#nextcloud-mail )

giulianograssi:
Hey there, 

i created a table with titles of the screenshots and linked the images directly to the urls received from the img in the playstore. i guess those links can get broken if you ever update the playstore version or the images itself. i heard NC has a file server where you have all screenshots stored. i think linking to those screenshots would be the better way to do this. 

Questions:
1. where can i find the file server?
2. is it publicly available?
3. do you have another view on solving this?
4. how can i change the status of the issue to "on progress" while im working on the issue?

the current situation can be found here:
https://github.com/giulianograssi/talk-android/tree/239-add-screenshots-to-readme

thank you for your answer

giulianograssi:
i just talked with @mario and the result is 
- There is no public available file server for screenshots
- Store the Screenshots into the docs folder
- Wait on the new screenshots used for the new version

just ping me when the new version is out and i will continue working on it.

mario:
@giulianograssi yup - will do. Thanks for contributing.

giulianograssi:
@mario please add ggrassi6 (at) gmail.com to the beta test group

1. i will then install the beta and make screenshots for the languages
- German
- English
- French
- Italian
2. i will upload them to talk-android/fastlane/metadata/android/xx-YY/phoneScreenshots/
3. i will use the en-US screenshots to get displayed in the README.md

mario:
@giulianograssi my apologies, I've just seen this but I think you became part of the beta group at the conference.

https://play.google.com/apps/testing/com.nextcloud.talk2

In the mean time 3.0 has been released as stable.

mario:
@giulianograssi you around? :)

PR #797, opened 2020-04-29
Add screenshots to README.md

stefan-niedermann:
Used the screens from the `fastlane` folder like @giulianograssi suggested in #239

@mario you probably should update the screenshots in the Play Store as well - they look different from what i can see (at least in the F-Droid version) :wink:.


mario:
Thanks. v9.0 is way different so I'll hopefully, eventually, any year now get the time to update the screenshots :)

Issue #877, opened 2020-07-24
Fix datachannel support with latest Janus


fancycode:
Adding here for the Android-related fixes to support the datachannel changes in the latest Janus (actually since 0.7).

See https://github.com/nextcloud/spreed/issues/3937 for further information.

danxuliu:
It was fixed in #907. As the _v8.0.x_ branch became the _master_ branch [it is already in master too](https://github.com/nextcloud/talk-android/blob/0bc3be1b6e9a42256a9ff1a65baabbb963e5b61f/app/src/main/java/com/nextcloud/talk/webrtc/MagicPeerConnectionWrapper.java#L423).

Issue #830, opened 2020-05-08
Full chat history is loaded when entering a conversation the first time


nickvergessen:
It should just start at the current readmarker instead.
load the 200 from before there and then the future.

ctilluma:
What ideal behavior are we looking for here? Load last 200 messages, and if there are more we can scroll back? I understand load time, but I also thought I read that Talk 9 could remember messages offline as well, although I could have misunderstood what was read. 

Ideally, we want a fast load time, and then the ability to access entire conversations - first online, then offline. So perhaps we do a quick load in the focus, and async a load of the entire conversation?

dan0xii:
I think #895 fixes this because I can't replicate it when running that branch. The chat window opens at the bottom and as you scroll up the additional text is loaded.

nickvergessen:
At least still happening when using "Join with link" feature

AndyScherzinger:
Chat loading was rewritten with the offline-first repository (#217) and again in #5635, so history is no longer fetched the way this describes. Closing as stale — please open a fresh report if entering a conversation still pulls the full history on 25.x.

PR #882, opened 2020-08-11
Fixes for #877 and #830

ctilluma:
Updated channel support for Janus as per #877 
Limited downloading conversation to last 200 messages, as per #830


nickvergessen:
@ctilluma do you need help to figure out how to split the things so there is 1 PR per change

PR #904, opened 2020-10-02
[v8.0.x] Fix some issues related to guest participants

danxuliu:
This pull request adds avatars for guests in the call view, as well as fixing other related issues. For details please refer to the individual commit messages :-)

It will probably need to be forwardported, although I have not checked.

Also note that the avatar is currently shown only for guests for which there is a peer connection (as the guest avatar depends on the nick, and the nick for guests when the HPB is used is currently sent through data channels, which require a peer connection). None of the changes from https://github.com/nextcloud/spreed/pull/4182 are applied here.

## How to test (scenario 1)

- Setup the HPB
- Create a public conversation
- With the Android app, join the public conversation
- With a browser, join the public conversation as a guest

Nothing strange happens.

The web socket of the Android app gets reconnected.

## How to test (scenario 2)

- Setup the HPB
- Create a public conversation
- With a browser, join the public conversation as a guest and start a call
- With the Android app, join the public conversation and join the call
- In the browser, change the guest name

The guest name is updated.

The guest name is not updated. Also, checking the errors in Android's logcat it can be seen that NullPointerExceptions are thrown when the guest sends a `nickChanged` event (which happens every second).

## How to test (scenario 3)

- This change applies both when the HPB is used and when it is not used. *Important:* I have tested it only with the HPB; it should work without HPB too, but testing is very much welcome :-)
- Create a public conversation
- With a browser, join the public conversation as a guest and start a call
- With the Android app, join the public conversation and join the call
- In the browser, change the guest name

The guest avatar is initially shown and then updated when the guest name is changed.

The guest avatar is not shown.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -2031,16 +2031,11 @@ public class CallController extends BaseController {
         }
     }
 
-    private void gotNick(String sessionOrUserId, String nick, boolean isFromAnEvent, String type) {
-        if (isFromAnEvent && hasExternalSignalingServer) {
-            // get session based on userId
-            sessionOrUserId = webSocketClient.getSessionForUserId(sessionOrUserId);
-        }
-
-        sessionOrUserId += "+" + type;
+    private void gotNick(String sessionId, String nick, boolean isFromAnEvent, String type) {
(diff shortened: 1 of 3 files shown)


tobiasKaminsky (approved):

dasbiswajit (approved):
Perfect tested all scenarios.

PR #912, opened 2020-10-09
Chat during call

tobiasKaminsky:
This supports to toggle between chat and call during a video/audio call.


nickvergessen (approved):
Worked good (if this is the code you gave me)

PR #963, opened 2021-01-20
ignore .idea in .gitignore / remove .idea from index

mahibi:
executed
git rm -r --cached .idea
according to
https://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore


mahibi:
this has already been done in V8. Now it's done in master, so AndroidStudio should stop to complain about "missing modules" when branches are switched.

tobiasKaminsky (approved):

tobiasKaminsky:
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

If you need help setting up your gpg, just ping me.

PR #964, opened 2021-01-22
Bump master to 11.0.0 Alpha 01

tobiasKaminsky:
Master should be always ahead of all stable branches.
Current latest stable release is 8.2.0, from v8.0.x, so master is 11.0.0 Alpha 01.


app/build.gradle
```diff
@@ -34,13 +34,12 @@ android {
...
         applicationId "com.nextcloud.talk2"
-        versionName version
         minSdkVersion 21
         targetSdkVersion 29
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 
-        versionCode 80020090 // 0-50=Alpha / 51-98=RC / 90-99=stable
-        versionName "8.2.0"
+        versionCode 110000000 // 0-50=Alpha / 51-98=RC / 90-99=stable
+        versionName "11.0.0 Alpha 01"
```
mahibi, line 42: is it common to call it Alpha 01 when the last digits of the version code are "00"?
should it always be shifted by 1 ??

tobiasKaminsky: Off by one error? :-D
You are right, for the same reason 51 is RC1, otherwise it confused me.
But with 1 = Alpha 1 we would have no 0, which is also strange.

If at all, only version name matters for users, so I am fine with anything you decide :-)

PR #973, opened 2021-01-28
Delombok everything

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/models/json/chat/ChatMessage.java
+++ b/app/src/main/java/com/nextcloud/talk/models/json/chat/ChatMessage.java
@@ -43,10 +43,8 @@ import java.util.List;
 import java.util.Map;
 
 import androidx.annotation.Nullable;
-import lombok.Data;
 
 @Parcel
-@Data
 @JsonObject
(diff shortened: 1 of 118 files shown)


app/src/main/java/com/nextcloud/talk/models/json/chat/ChatMessage.java
```diff
@@ -251,6 +252,245 @@ public String getSystemMessage() {
...
+
+    public void setReadStatus(Enum<ReadStatus> readStatus) {
+        this.readStatus = readStatus;
+    }
+
+    public void setMessageTypesToIgnore(List<MessageType> messageTypesToIgnore) {
+        this.messageTypesToIgnore = messageTypesToIgnore;
+    }
+
+    public boolean equals(final Object o) {
```
mahibi, line 391: creepy... ;) i think delombok should be the last emergency solution...

tobiasKaminsky: Yes :/ 
Maybe it is better to move those classes one by one to Kotlin data classes…

mahibi:
closing this PR because delombok would add too much boilerplate. 
Lombok problems will have to be solved by moving to kotlin data classes..

Issue #817, opened 2020-05-06
Changing the description "Privacy"


Valdnet:
I suggest changing the "Privacy" description in the "About" app to a different one, e.g. "Privacy and Legal Policy"

In my opinion, the same names in two different options is confusing to the user.

![obraz](https://user-images.githubusercontent.com/47037905/81199212-a8677000-8fc2-11ea-9ff4-53f73cdc52e4.png)

mario:
Will be in beta5.

PR #990, opened 2021-02-02
Change text strings #458 , #817

Valdnet:
Changes that appeared in version 9.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -81,10 +81,10 @@
     <string name="nc_settings_no_talk_installed">Talk app is not installed on the server you tried to authenticate against</string>
     <string name="nc_settings_account_updated">Your already existing account was updated, instead of adding a new one</string>
     <string name="nc_account_scheduled_for_deletion">The account is scheduled for deletion, and cannot be changed</string>
-    <string name="nc_settings_notification_sounds">Notification sounds</string>
-    <string name="nc_settings_call_ringtone">One-on-One Calls</string>
+    <string name="nc_settings_notification_sounds">Sound</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -142,7 +142,7 @@
     <string name="nc_new_password">New password</string>
     <string name="nc_wrong_password">Wrong password</string>
     <string name="nc_about">About</string>
-    <string name="nc_privacy">Privacy</string>
+    <string name="nc_privacy">Privacy and Legal Policy</string>
```
mahibi, line 145: As far as is see there's no need to rename "Privacy" because the menu for master doesn't contain any other clickable entry like this? (like it would have been for v9)

i like the clarification of the items for the sound settings!

mahibi:
i like the clarification of the items for the sound settings! makes sense

app/src/main/res/values/strings.xml
Valdnet: You're right. I didn't notice it. I returned to the previous description 😉.

Valdnet:
> i like the clarification of the items for the sound settings!

This was a topic that was discussed with @mario for a long time. It was hard to convince him to make this change, but it finally made it.

app/src/main/res/values/strings.xml
Valdnet: Alternatively, I can change to a more professional name which is: `Privacy policy`.

mahibi: i think "Privacy" is fine. Files app also named it like this.

mahibi (approved):

mahibi:
thanks @Valdnet :)

PR #999, opened 2021-02-07
Feature/956/call screen

mahibi:


app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -1958,6 +1970,7 @@ private void setupAvatarForSession(String session) {
                         avatarImageView.setController(draweeController);
                     }
                 }
+                setCallState(CallStatus.IN_CONVERSATION);
```
mahibi, line 1973: - [ ] discuss with @tobiasKaminsky if this is allowed here and if deleting in line 1588 was okay

tobiasKaminsky: As far as I understand it, you postpone setting call state "in_conversation" a bit later, so that you are sure that already a proper connection to at least one recipient is available.
I think it is fine :+1:

app/src/main/res/layout/controller_call.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?><!--
   ~ Nextcloud Talk application
   ~
-  ~ @author Mario Danic
-  ~ Copyright (C) 2017-2018 Mario Danic <mario@lovelyhq.com>
+  ~ @author Marcel Hibbe
+  ~ Copyright (C) 2021 Marcel Hibbe <dev@mhibbe.de>
```
mario, line 5: I'd suggest you add yourself, rather than completely eliminating me :P

mahibi: oh sorry mario, i will do that!
I guess i finally have to adjust my copyright template for Android Studio so that I don't always have to change everything manually. I suspect this happened accidentally when I tried to manually replace my name for completely new files.
no bad intention ..

- [ ] update copyright

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
mahibi: at least deleting in line 1588 was not okay and introdcued new bugs. anyway, busy with other things right now....

app/src/main/res/layout/controller_call.xml
mario: No bad intention was assumed ;)

app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java
```diff
@@ -127,6 +127,10 @@ public static String getRoom(String baseUrl, String token) {
         return baseUrl + ocsApiVersion + spreedApiVersion + "/room/" + token;
     }
 
+    public static String getRoomV3(String baseUrl, String token) {
```
tobiasKaminsky, line 130: Is this also in older Talk versions?

tobiasKaminsky (approved):

app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java
tobiasKaminsky: There is a check, great work!

tobiasKaminsky (approved):

mahibi:
fix #956

PR #1010, opened 2021-02-15
Login Textfield design

AndyScherzinger:
Original, corresponding Android Files client PR based on the design review by @jancborchardt and others (https://github.com/nextcloud/android/pull/7971): 

* Using container-style for text fields (initially just for the login screen since it has a specific set of colors)

### Screenshots

| Filled Address  | Error Text | Checking Server |
| ------------- | ------------- | ------------- |
|  device-2021-02-15-153947 | device-2021-02-15-155501 |  device-2021-02-15-154229 |

### Sidenote
* atm the talk client simply states the same error (see screenshot) no matter the error (e.g. server doesn't even exists Vs. talk app not installed). This could be changed but is not related to this UI PR and can be added in the future by anybody, the UI is in pace
* talk client has a icon opacity logic for enabled/disabled which the files client lacks atm (so either remove or add to the files client in the future)
* like with the files client the helper-text is brand-able

Other than that
* Hi @mahibi 👋 
* cc @tobiasKaminsky for the opacity thing 😃

PR #1014, opened 2021-02-17
Drone signs its comments

tobiasKaminsky:


tobiasKaminsky:
Otherwise PRs like https://github.com/nextcloud/talk-android/pull/1010 will always fail

tobiasKaminsky:
Lint is up because of:
1 more obsolete gradle dependency
2 more TypographyEllipsis: Ellipsis string can be replaced with ellipsis character --> I would limit this to english only

mahibi (approved):

tobiasKaminsky:
> Lint is up because of:
> 1 more obsolete gradle dependency
> 2 more TypographyEllipsis: Ellipsis string can be replaced with ellipsis character --> I would limit this to english only

This is done in #1015, so we can merge this.

PR #1021, opened 2021-02-22
Use valid email address

tobiasKaminsky:


mahibi (approved):

tobiasKaminsky:
Ok, This just does not work, an I have no idea why.

So if there is a drone commit reducing lint count, we have to either approve DCO manually or use admin rights to merge the PR.

PR #1040, opened 2021-02-25
fix missing deck card links / always show links if available in message

### LINKED ISSUE #1025: deck card link is show as {object}
Browser:
image

App:
image

mahibi:
fix #1025 


app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+
+package com.nextcloud.talk.models.json.chat
+
+import java.util.*
+
+class ChatUtils {
+    companion object {
+        fun getParsedMessage(message: String?, messageParameters: HashMap<String?, HashMap<String?, String?>>?):
+                String? {
+            var message = message
```
tobiasKaminsky, line 31: That's uncommon, or?
If you want to you can change parameter to "var message" directly.

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+            if (messageParameters != null && messageParameters.size > 0) {
+                for (key in messageParameters.keys) {
+                    val individualHashMap = messageParameters[key]
+                    val type = individualHashMap?.get("type")
+                    if (type == "user" || type == "guest" || type == "call") {
```
tobiasKaminsky, line 36: type can be null if individualHashMap is null.
When you use "user" == type then it will not crash with NPE.

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.nextcloud.talk.models.json.chat
+
+import java.util.*
```
tobiasKaminsky, line 25: try to avoid this generic imports

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+        fun getParsedMessage(message: String?, messageParameters: HashMap<String?, HashMap<String?, String?>>?):
+                String? {
+            var message = message
+            if (messageParameters != null && messageParameters.size > 0) {
+                for (key in messageParameters.keys) {
+                    val individualHashMap = messageParameters[key]
+                    val type = individualHashMap?.get("type")
+                    if (type == "user" || type == "guest" || type == "call") {
+                        message = message?.replace("{$key}", "@" + individualHashMap["name"])
+                    } else if (individualHashMap?.containsKey("link") == true) {
```
tobiasKaminsky, line 38: == true is unnecessary.

mahibi: grafik

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
mahibi: agree.
in fact is was even not needed

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
mahibi: for function parameters "var" is not allowed in kotlin (https://discuss.kotlinlang.org/t/method-parameters-and-mutable-variables/333). I guess you were thinking about primary constructors for classes where it is allowed.

`var message = message`  is called "shadowing" (https://stackoverflow.com/questions/49680040/why-kotlin-allows-to-declare-variable-with-the-same-name-as-parameter-inside-the), but maybe it's not best practice here.
so i will go for
`var resultMessage = message`
Not sure if this is best practice so i'm open to other ideas..

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
mahibi: this will not throw a NPE. 
This would resolve to  `null == "user"` which is "false"
I guess you were thinking about java where it's better to check for
`"mystring".equals(value)`
in case of
`value.equals("mysting")` (-> NPE if value is null)
?

tobiasKaminsky: Right, bad Java habits ;-)

app/src/main/java/com/nextcloud/talk/models/json/chat/ChatUtils.kt
tobiasKaminsky: True 🤣
I overlooked "?."

tobiasKaminsky (approved):

jakobroehrl:
@mahibi @stefan-niedermann 
Is it possible to open the card in the Android App if installed?

stefan-niedermann:
@jakobroehrl yes, there are two options:

1. talk directly checks whether deck is installed and launches an intent
2. Deck catches the browser-open action and opens itself (https://github.com/stefan-niedermann/nextcloud-deck/issues/498)

PR #1046, opened 2021-03-01
fix calls

mahibi:


mahibi:
- [ ] load avatars again in CallNotificationController (they seemed to block further processes when in debug mode)

mahibi:
closed. will be split into 2 braches..

PR #1049, opened 2021-03-04
Fix calls

mahibi:


jakobroehrl:
@mahibi 
Could you tell, what should PR fix/do?

mahibi:
> @mahibi
> Could you tell, what should PR fix/do?

sorry overseen that @jakobroehrl 
this branch was originally to fix a bug with the call connections and to get an understanding how everything fits together so there are many debug commands.
Meanwhile some time has passed and 
- i can't reproduce the bug anymore 😆
- had some troubles with rebasing
i will delete this branch. copied the debug log commands to https://github.com/nextcloud/talk-android/pull/1096 in case they are still needed

Issue #1064, opened 2021-03-22
make differentiation between audio/video call if currentConversation is passed to CallNotificationController


mahibi:
right now differentiation is only done in "handleFromNotification".
But if currentConversation was already passed in, there has to be a differentiation between audio/video too.

```
        if (currentConversation == null) {
            handleFromNotification();
        } else {
            runAllThings();    // handleFromNotification not called so there is no differentiation. audio/video differentiation might have to be passed as another parameter into CallNotificationController
        }
```

AndyScherzinger:
`CallNotificationController` no longer exists — call notifications are handled by `NotificationWorker` and `CallNotificationActivity`, which do differentiate audio and video calls via `KEY_ROOM_ONE_TO_ONE` and the call flags. Closing as obsolete.

PR #1065, opened 2021-03-22
set application name in call screens from String nc_app_name

mahibi:

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -2174,7 +2174,7 @@ public class CallController extends BaseController {
                         }
                         callConversationNameTextView.setText(conversationName);
 
-                        callVoiceOrVideoTextView.setText(isVoiceOnlyCall ? R.string.nc_voice_call : R.string.nc_video_call);
+                        callVoiceOrVideoTextView.setText(getDescriptionForCallType());
 
                         if (callStateView.getVisibility() != View.VISIBLE) {
                             callStateView.setVisibility(View.VISIBLE);
@@ -2197,7 +2197,7 @@ public class CallController extends BaseController {
                     handler.post(() -> {
                         hangup(false);
                         callStateTextView.setText(R.string.nc_call_timeout);
-                        callVoiceOrVideoTextView.setText(isVoiceOnlyCall ? R.string.nc_voice_call : R.string.nc_video_call);
+                        callVoiceOrVideoTextView.setText(getDescriptionForCallType());
                         if (callStateView.getVisibility() != View.VISIBLE) {
                             callStateView.setVisibility(View.VISIBLE);
                         }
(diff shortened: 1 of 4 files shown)


mahibi:
by the way i realized https://github.com/nextcloud/talk-android/issues/1064 but this should be done later

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -2360,6 +2360,15 @@ private void setCallState(CallStatus callState) {
         }
     }
 
+    private String getDescriptionForCallType() {
+        String appName = Objects.requireNonNull(getResources()).getString(R.string.nc_app_name);
```
tobiasKaminsky, line 2364: If you think this might be null, then you also need to handle resulting NPE.

app/src/main/res/values-de/strings.xml
```diff
@@ -166,9 +166,9 @@
     <string name="nc_permissions_settings">Einstellungen öffnen</string>
 
     <!-- Call -->
-    <string name="nc_voice_call">Nextcloud Talk Sprachanruf</string>
```
tobiasKaminsky, line 169: This will be overwritten again via Transifex, so please do not commit this.
Best practice is not to change an already existing string, but use a new one.
Otherwise you might have strange UX if a language is not updated.
With having a new string, worst what could happen is that it is English…

app/src/main/java/com/nextcloud/talk/controllers/CallNotificationController.java
```diff
@@ -343,11 +343,15 @@ private void runAllThings() {
         showAnswerControls();
     }
 
-    @SuppressLint("LongLogTag")
+    @SuppressLint({"LongLogTag", "SetTextI18n"})
```
tobiasKaminsky, line 346: Why is this added?

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
mahibi: ah forgot to change this. actually we don't check anywhere for NPE for resources. so i stick with this. or any concerns?

app/src/main/res/values-de/strings.xml
mahibi: ah sure. renamed the keys and changed the values

app/src/main/java/com/nextcloud/talk/controllers/CallNotificationController.java
mahibi: not needed anymore

app/src/main/java/com/nextcloud/talk/controllers/CallNotificationController.java
```diff
@@ -343,11 +343,15 @@ private void runAllThings() {
...
 
-    @SuppressLint("LongLogTag")
+    @SuppressLint({"LongLogTag"})
     @Override
     protected void onViewBound(@NonNull View view) {
         super.onViewBound(view);
 
+        String callDescriptionWithoutTypeInfo =
+                getResources().getString(R.string.nc_app_name) + " " + getResources().getString(R.string.nc_call_unknown);
+        incomingCallVoiceOrVideoTextView.setText(callDescriptionWithoutTypeInfo);
```
tobiasKaminsky, line 353: String concatenation are normally done as a separate string with "%s1 and %s2" so that left to right languages work.

mahibi: but here i combine two translated strings. it would be possible to make another string that has the value "%1$s %2$s" so for other languages it could be changed to "%2$s %1$s", but i don't think anyone will understand that?
or do i get it wrong?

tobiasKaminsky: True, this is hard to understand on translator side.
Maybe then have "%s call", where %s is app name?

mahibi: ah sure 🤦‍♂️ 😆

PR #1068, opened 2021-03-23
Fix null name in phonebook

mahibi:


app/src/main/res/xml/auth.xml
```diff
@@ -23,4 +23,4 @@
     android:accountType="com.nextcloud.talk2"
     android:icon="@mipmap/ic_launcher"
     android:smallIcon="@mipmap/ic_launcher"
-    android:label="Nextcloud Talk"/>
+    android:label="@string/nc_app_name"/>
```
mahibi, line 26: this should reslove the "null" strings

app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
```diff
@@ -68,11 +69,17 @@ class ContactAddressBookWorker(val context: Context, workerParameters: WorkerPar
...
+    private lateinit var accountName: String
+    private lateinit var accountType: String
+
     override fun doWork(): Result {
         sharedApplication!!.componentApplication.inject(this)
 
         val currentUser = userUtils.currentUser
 
+        accountName = context.getString(R.string.nc_app_name)
+        accountType = BuildConfig.APPLICATION_ID
```
mahibi, line 81: @tobiasKaminsky 
this is how applicationId is taken from build.gradle
But auth.xml and syncadapter.xml contain hardcoded "com.nextcloud.talk2"!
Because of this, `addAccountExplicitly` will fail with SecurityException when Brander has replaced the applicationId
(just like it's failing when in debug mode, because then "debug" is added to the applicationId
--> `java.util.concurrent.ExecutionException: java.lang.SecurityException: uid 10192 cannot explicitly add accounts of type: com.nextcloud.talk2.debug`)

So we need a way to replace the hardcoded entries in auth.xml and syncadapter.xml right? (https://stackoverflow.com/questions/57074999/get-application-id-in-xml-dynamically)
So for example if the applicationId is set as a string in strings.xml, Brander would still have to be adjusted to replace this string...?

app/build.gradle
```diff
@@ -78,9 +78,9 @@ android {
     }
 
     buildTypes {
-        debug {
-            applicationIdSuffix '.debug'
-        }
+//        debug {
+//            applicationIdSuffix '.debug'
+//        }
```
tobiasKaminsky, line 83: Please do this in a new PR.

app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
mahibi: Using resource-strings in auth.xml and syncadapter.xml didn't work. App somehow didn't react to onChanged events anymore 🤷‍♂️
Now brander replaces the applicationId in auth.xml and syncadapter.xml  directly..

app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
```diff
@@ -68,11 +69,19 @@ class ContactAddressBookWorker(val context: Context, workerParameters: WorkerPar
...
+        Log.d(TAG, "accountName: " + accountName)
+        Log.d(TAG, "accountType: " + accountType)
```
tobiasKaminsky, line 83: I guess this can be deleted.

PR #1073, opened 2021-03-24
Show avatars for incoming pictures in group chats

starypatyk:
I have noticed that avatars are not displayed when pictures from NC server are posted in a group chat.

Proposed change seems to fix this. The avatars are displayed correctly - only in group chats.

In one-to-one conversations this additional UI element is hidden by code already present in MagicPreviewMessageViewHolder.onBind method.

It seems that the SimpleDraweeView (messageUserAvatar) has been accidentally deleted in this commit 2613c1f0749df0230d9032b732e19104841b3904

### DIFF
--- a/app/src/main/res/layout/item_custom_incoming_preview_message.xml
+++ b/app/src/main/res/layout/item_custom_incoming_preview_message.xml
@@ -28,13 +28,21 @@
     android:layout_marginEnd="16dp"
     android:layout_marginBottom="2dp">
 
+    <com.facebook.drawee.view.SimpleDraweeView
+        android:id="@id/messageUserAvatar"
+        android:layout_width="24dp"
+        android:layout_height="24dp"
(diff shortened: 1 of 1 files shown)


mahibi:
Hi @starypatyk 
Thanks for your PR!! thats great! 👍👍
I had some problems to run the app on your branch on android studio (but it was due to android studio and not your branch i think, so all good).
However i copied your fix to another PR (https://github.com/nextcloud/talk-android/pull/1081) so this fix will be available in the next beta. Hope that's okay for you for this time.
I suggest @tobiasKaminsky adds you to our android nextcloud group so if you want you will be able to create branches in our repositories without to fork ;)
Or Tobias gives me the right to add members ;)

starypatyk:
Hi @mahibi,
I am glad that it got included - regardless how. :wink:
Thanks!

PR #1100, opened 2021-04-06
On Recyclerview everything needs to be set explicitly

tobiasKaminsky:
On current master
- have only one unread conversation in bold
- scroll up/ down
- see other unrelated conversation also in bold

--> on recyclerview items are recycled, thus the state also persists, so we have to set them explicitly.


AndyScherzinger (approved):

AndyScherzinger:
Should then also be back ported since the original change introducing bold text style is on v11 and v12 branches AFAIK. But @mahibi might know best 👍

mahibi:
will be backported

PR #1101, opened 2021-04-06
Remove "Nextcloud" hardcoding

mahibi:


mahibi:
- [ ] Brander should differentiate between files and talk..?

app/src/main/java/com/nextcloud/talk/ui/dialog/AttachmentDialog.kt
```diff
@@ -49,11 +49,13 @@ class AttachmentDialog(val activity: Activity, var chatController :ChatControlle
         super.onCreate(savedInstanceState)
         val view = layoutInflater.inflate(R.layout.dialog_attachment, null)
         setContentView(view)
-
         window?.setLayout(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
-
         unbinder = ButterKnife.bind(this, view)
 
+        attachFromCloud?.text = chatController.resources?.let {
+            String.format(it.getString(R.string.nc_upload_from_cloud), it.getString(R.string.nc_server_product_name))
```
tobiasKaminsky, line 56: You can use ThemingCapability -> name for this.
Then it is even more personalized, as it then does not show "share from Nextcloud", but maybe "share from Tobi's personal cloud".

tobiasKaminsky (approved):

app/src/main/java/com/nextcloud/talk/ui/dialog/AttachmentDialog.kt
mahibi: done

PR #1113, opened 2021-04-09
l10n: Correct spelling

Valdnet:


AndyScherzinger:
@mahibi I checked the server and this would align with their way (also this is basically what @Valdnet described 👍)

AndyScherzinger (approved):

PR #1114, opened 2021-04-09
Fix own peer handling when HPB is used

danxuliu:
When processing the participants in a room the signaling sessions of the participants were compared against the Nextcloud session of the local participant to find out which were the remote sessions. However, when the HPB is used the signaling session is not the same as the Nextcloud session, so the signaling session of the local participant never matched her Nextcloud session, so it was always seen as a remote participant.

This caused the call state to be changed to `in conversation` (which, for example, stopped the calling sound), when only the local participant was in the call.

## How to test

- Setup the HPB
- Open the Talk Android app from scratch (otherwise currently there could be dangling sessions, something to be fixed later)
- Open a conversation
- Start a call

The calling sound plays.

The calling sound does not play (or plays briefly and then stops).

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -1707,7 +1707,7 @@ public class CallController extends BaseController {
         }
 
         for (String sessionId : newSessions) {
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
❗ rebased to run latest Github actions and other checks / CI ❗

mahibi:
According to @danxuliu this PR is ready to review but needs to be tested without HPB

mahibi (approved):

mahibi:
tested both with hpb and without.
thanks @danxuliu !

PR #1119, opened 2021-04-12
🍱 Grid view for calls

### LINKED ISSUE #871: Grid View for video Calls
like in jitsi would be nice. 
This time it's hard to see all people in meetings > 5

mahibi:
fix #871 

work on this branch might be paused for a few days. beside cleaning up the following tasks will be continued later:

- [x] shrink gridView for audio calls
   - [x] simplify height calculation for grid in audio calls (display height - header should be enough?)
- [x] improve grid design
   - [x] landscape vs. portrait view
   - [x] for tablets
- [x] handle rotation of screen
   - [x] reset picture in picture view
- [x] display muted micropones etc
- [x] click listener for grid when there are no cells?
- [x] show avatar for participants who disabled video
- [x] picture in picture view should be freely movable (otherwise it would always overlap participant in the right corner)
   - [ ] add rounded corners 
       - seems not possible: https://stackoverflow.com/questions/36564025/how-to-make-a-rounded-surfaceview/36583173#36583173
       - wrapping with CardView fails 
- [x] bug: calculation of grid height fails when there is the animation for call infos
- [ ] design improvements?
- [x] test with MCU?
- [x] fix view for 1 participant in horizontal layout
- [x] tidy up / lint errors?
- [x] no TODO's?
   - [x] "gotNick"
- [x] add changelog entry

used https://stackoverflow.com/a/19467717 to have a HashMap for grid adapter. might be there are some better solutions..


AndyScherzinger:
increased findbugs warnings 😢

mahibi:
> increased findbugs warnings cry

sure will reduce it once it's ready

AndyScherzinger:
This is how the files client handles list/grid views for files (which is also just a item layout + the right layout manages)
https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java#L1367-L1383

jancborchardt:
@mahibi let me know when/if you have screenshots for review :)

mahibi:
> @mahibi let me know when/if you have screenshots for review :)

i will, thx @jancborchardt 
probably today or tomorrow..

nickvergessen:
Build locally and worked nicely

nickvergessen:
Dont mind the shade  appearing I'm too stupid to create a screenshot without tabbing the screen so it always pops up like this.

= | II
---|---
Screenshot_2021-05-05-16-08-40-144_com nextcloud talk2 qa | Screenshot_2021-05-05-16-08-26-954_com nextcloud talk2 qa

app/src/main/res/layout/call_item.xml
```diff
@@ -23,53 +25,38 @@
...
+
+    <org.webrtc.SurfaceViewRenderer
+        android:id="@+id/surface_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
         android:visibility="invisible" />
 
     <ImageView
-        android:id="@+id/remote_video_off"
+        android:id="@+id/remote_audio_off"
```
nickvergessen, line 54: you removed the `remote_video_off` option to disable a remote persons video and the audio thing replaces the video one?

jancborchardt:
Thanks @nickvergessen! :) And @mahibi just some detail design feedback:
- [ ] The mute indicator would be better next to the name, like it is in the web interface.
- [ ] The "Nextcloud Talk video call" title on the top can be removed, it just causes the title to overlap more of the participants. :)
- [ ] The call title text seems to be a bit large? It could be like in the chat view, same font size and bold (but centered is good here).
- [ ] The bottom controls bar is blue, which distracts a bit. Would be nicer just black?

What do you think?

AndyScherzinger:
> The bottom controls bar is blue, which distracts a bit. Would be nicer just black?

There is already a PR for this https://github.com/nextcloud/talk-android/pull/1173 which needs some more logic since the navigation bar color needs to be reset to the dark/light theme depending from where the call has been started (app launched and incomming call, app closed and incoming call) This is due to the splash screen which sets a blue navigation bar while in the app the light theme would set a light nav bar, so the call activity should rather set a custom theme which doesn't theme the nav bar and reset it to the default theme when terminating the call.

nickvergessen:
> The "Nextcloud Talk video call" title on the top can be removed, it just causes the title to overlap more of the participants. :)

This only appears on touch btw, its what I meant with "I'm to stupid to make a screenshot", but I agree it can be removed and we use that size for the conversation name then

AndyScherzinger:
@mahibi @nickvergessen here is a rebase dversion of this branch https://github.com/nextcloud/talk-android/tree/gridViewForCalls_Rebase while it issue mentioned by Joas remains (the turn off remote video option is gone)

app/src/main/res/layout/call_item.xml
mahibi: the `remote_video_off` icon was shown whenever a participant disabled it's own video. so at the same moment his video is replaced with his avatar. So the `remote_video_off` icon made no sense at all?

mahibi:
>     * The mute indicator would be better next to the name, like it is in the web interface.

@jancborchardt  aggree (updatet this in our doc, btw). 
i will remove the allocated place for the nickname when video is on. otherwise there would be a lot of space if i align the muted icon right to it but the nick is invisible. But it's wanted that the nickname is not shown when video is on right? otherwise i wouldn't need to remove any allocated space...

app/src/main/res/layout/call_item.xml
nickvergessen: Right, I see it now. So okay for now and a feature to add later.

app/src/main/res/values/strings.xml
```diff
@@ -284,8 +284,7 @@
     <!-- When translating to German, please use non-formal variant -->
     <string name="nc_formatted_message_you">You: %1$s</string>
     <string name="nc_message_read">Message read</string>
-    <string name="nc_message_sent">Message sent</string>
-    <string name="nc_remote_video_off">Remote video off</string>
+    <string name="nc_message_sent">Message sent</string>nc_add_attachment
```
nickvergessen, line 287:
```suggestion
    <string name="nc_message_sent">Message sent</string>
```

opsy

AndyScherzinger: Whoopy Daisy... fixed 👍

nickvergessen:
Somehow the QR build didn't work.

AndyScherzinger:
@tobiasKaminsky might be able to check next week

mahibi:
grafik

grafik

AndyScherzinger:
Nice work @mahibi 👍

nickvergessen (approved):

PR #1130, opened 2021-04-14
Accessibility participants list and grey tones

### LINKED ISSUE #1129: Optimize grey tones according to WCAG
At the moment the participants list is "greying-out" items for participants who are offline. It basically sets alpha to ```0.38``` which matches Material Design's value for disabled text. Unfortunately the accessibility scanner doesn't white like that (contrast ratio to low 2.x ...) So the question is do we want to use "disabled text color" (like material) or do we rather use the "secondary / medium-emphasis test color which WCAG would be fine with.

So calling @jancborchardt to the rescue :)

|now|medium emphasis text colored|
|---|---|
|device-2021-04-14-131825|device-2021-04-14-150555|

### LINKED ISSUE #1131: low contrast text color on mention chip-views on dark theme
see screenshots

|dark|light|
|---|---|
|device-2021-04-14-185814|device-2021-04-14-185856|

AndyScherzinger:
Resolves #1129

Also 
* removes the dark-blue-grey headline color and replaces any occurrences with high-emphasis text color except for the chat bubbles, they have been kept as is for now (might change with a separate PR)
* Sorts the participants list in a hierarchical way: ownership, online, moderator, displayname (like the webUI)
* WCAG optimizations regarding inactive icons, hitbox size of min=48dp
* don't make toolbar menu item text all-caps

|Settings|User profile|
|---|---|
|device-2021-04-14-184344|device-2021-04-14-231748|

Also resolves #1131
|before|after|after|
|---|---|---|
|device-2021-04-14-185814|device-2021-04-14-232544|device-2021-04-14-232755|

|before|after|after|
|---|---|---|
|device-2021-04-14-185856|device-2021-04-14-232631|device-2021-04-14-232711|

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.kt
@@ -78,7 +78,6 @@ import com.yarolegovich.mp.MaterialStandardPreference
 import com.yarolegovich.mp.MaterialSwitchPreference
 import eu.davidea.flexibleadapter.FlexibleAdapter
 import eu.davidea.flexibleadapter.common.SmoothScrollLinearLayoutManager
-import eu.davidea.flexibleadapter.items.AbstractFlexibleItem
 import io.reactivex.Observer
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.Disposable
@@ -142,8 +141,8 @@ class ConversationInfoController(args: Bundle) : BaseController(args), FlexibleA
     private var databaseStorageModule: DatabaseStorageModule? = null
     private var conversation: Conversation? = null
 
-    private var adapter: FlexibleAdapter<AbstractFlexibleItem<*>>? = null
-    private var recyclerViewItems: MutableList<AbstractFlexibleItem<*>> = ArrayList()
+    private var adapter: FlexibleAdapter<UserItem>? = null
+    private var recyclerViewItems: MutableList<UserItem> = ArrayList()
 
     private var saveStateHandler: LovelySaveStateHandler? = null
 
@@ -376,6 +375,7 @@ class ConversationInfoController(args: Bundle) : BaseController(args), FlexibleA
             }
         }
 
+        Collections.sort(recyclerViewItems, UserItemComparator())
 
         if (ownUserItem != null) {
             recyclerViewItems.add(0, ownUserItem)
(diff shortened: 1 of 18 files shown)


jancborchardt (approved):
Looks grrrrreat! :)

AndyScherzinger:
@mahibi: the latest changes after the approval have been discussed on the talk channel between Jan and me :)

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.kt
```diff
@@ -684,4 +684,23 @@ class ConversationInfoController(args: Bundle) : BaseController(args), FlexibleA
 
         private const val ID_DELETE_CONVERSATION_DIALOG = 0
     }
+
+    /**
+     * Comparator for participants, sorts by online-status, moderator-status and display name.
```
nickvergessen, line 689: > Sorts the participants list in a hierarchical way: ownership, online, moderator, displayname (like the webUI)

Web sorts:
* _Later:_ actor type (only with api v4) =>participants with actor type group last
* participant has a session (_Later:_ user status === offline forces false) => participants with sessions first
* participant in call flag is !== 0 => participant with !== 0 first
* can moderate (check participant type [OWNER, MODERATOR, GUEST_MODERATOR]) => moderator powers first
* _Later:_ user status is dnd => dnd last
* Display name => alphabetically case-insensitive

https://github.com/nextcloud/spreed/blob/master/src/components/RightSidebar/Participants/CurrentParticipants/CurrentParticipants.vue#L139-L201

app/src/main/java/com/nextcloud/talk/controllers/ProfileController.java
```diff
@@ -809,15 +809,15 @@ public void afterTextChanged(Editable s) {
...
+                    holder.scope.setAlpha(0.87f); // active - high emphasis
                 } else {
                     holder.text.setEnabled(false);
                     holder.text.setFocusableInTouchMode(false);
                     holder.text.setEnabled(false);
                     holder.text.setCursorVisible(false);
                     holder.text.setBackgroundTintList(ColorStateList.valueOf(Color.TRANSPARENT));
                     holder.scope.setOnClickListener(null);
-                    holder.scope.setAlpha(0.4f);
+                    holder.scope.setAlpha(0.6f); // incative - medium emphasis
```
nickvergessen, line 820:
```suggestion
                    holder.scope.setAlpha(0.6f); // inactive - medium emphasis
```

AndyScherzinger: DONE 👍

app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.kt
AndyScherzinger: Most of the infos aren't available on the Android client yet :/

So I changed the sorting to 
* online status
* moderator type(s) (moderator, owner; guest_moderator is not known to the client yet or rather not processed)
* Display name => alphabetically case-insensitive

mahibi (approved):

mahibi:
thx @AndyScherzinger !

AndyScherzinger:
@mahibi this PR is depending on UI changes present on master but not on v11 (UI rewrite with new search bar etc.)

mahibi:
yep realized that 😄

PR #1142, opened 2021-04-20
Support delete-multiple push notification

nickvergessen:
🧑‍🦯 (blind) coding here. I still fail to compile and run the app locally and sending push notifications also is rather troublesome locally.
Also I don't know kotlin and googled the foreach of java, but I want to help moving forward to fix a problem with dying push notifications in iOS by implementing https://github.com/nextcloud/notifications/issues/936


app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
```diff
@@ -162,6 +162,10 @@ class MagicFirebaseMessagingService : FirebaseMessagingService() {
                             cancelExistingNotificationWithId(applicationContext, signatureVerification!!.userEntity, notificationId)
                         } else if (deleteAll) {
                             cancelAllNotificationsForAccount(applicationContext, signatureVerification!!.userEntity)
+                        } else if (deleteMultiple) {
+                            notificationIds.forEach {
+                                cancelAllNotificationsForAccount(applicationContext, signatureVerification!!.userEntity, it)
```
AndyScherzinger, line 167: Compile error
```
> Task :app:compileGplayDebugKotlin FAILED
e: /drone/src/app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt: (167, 122): Too many arguments for public final fun cancelAllNotificationsForAccount(context: Context?, conversationUser: UserEntity): Unit defined in com.nextcloud.talk.utils.NotificationUtils
```

@nickvergessen I guess you would rather want to use ```cancelExistingNotificationWithId``` but I don't know anything about the notification part in Talk...

nickvergessen: Fixed, yeah was a copy mistake

AndyScherzinger (approved):
Looks good to me code-wise (as in "untested")

app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
```diff
@@ -162,6 +162,10 @@ class MagicFirebaseMessagingService : FirebaseMessagingService() {
                             cancelExistingNotificationWithId(applicationContext, signatureVerification!!.userEntity, notificationId)
                         } else if (deleteAll) {
                             cancelAllNotificationsForAccount(applicationContext, signatureVerification!!.userEntity)
+                        } else if (deleteMultiple) {
+                            notificationIds.forEach {
```
tobiasKaminsky, line 166: I am not sure how json is handling this is if server response is empty/null.
Is notificationIds (long[]) then also null, then it will fail.
Or is deleteMultiple only true if there is really an array for ids?

AndyScherzinger: `deleteMultiple` is a boolean value within the json response, so I'd say "false if not present" so the loop won't be reached when the (also new) flag isn't there _and_ true.

AndyScherzinger:
@nickvergessen any info how this can be tested? :)

nickvergessen:
It could be tested with https://github.com/nextcloud/notifications/pull/939
But NC22 talk doesnt work with the mobile apps yet.

app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
nickvergessen: https://github.com/nextcloud/notifications/pull/939/files#diff-5f4fa2d752904832306c56f7832e2126a1e12db8dd1679abd71e526b77989821R544-R558

When it's empty it's delete-all, when its one entry its delete when it's multiple its delete-multiple

AndyScherzinger: Scenarios are
```
{"delete-all":true}
{"delete":true,"nid":1234}
{"delete-multiple":true,"nids":[1,2,3,4]}
```

AndyScherzinger (approved):

mahibi (approved):

PR #1152, opened 2021-04-22
QA Build & GHActions for Quality Checks

AndyScherzinger:
Adding QA builds to the Talk Android app like we have them at the files app repo too

* [x] Add secrets etc to repo so build/publish works @tobiasKaminsky 

Also
* align dependabot config with files client
* add funding info (also see files app)
* gradle-wrapper check
* ktlint & detekt checks
* Assembly for gplay/generic

Beware that ktlint can't be "configured" so with a large number of formatting issues it'll fail until we reformat some classes which might be possible over time after me merged most of the open PRs.


mahibi (approved):

mahibi:
🥳

.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Close stale issues'
```
mahibi, line 1: no-response.yml   vs  stale.yml 
see https://github.com/nextcloud/talk-android/pull/1168/files#diff-564029bc456818e4cbe6e9e4ebf8f5761753f232ac68297598c6283cc7c5f7e5

@AndyScherzinger @nickvergessen any thoughts which one to delete?

AndyScherzinger: And the other Android App uses stale.yml ;)

PR #1156, opened 2021-04-23
v4️⃣ - Add option for APIv4 compatibility

nickvergessen:
So there are quite some breaking changes with the upcoming APIv4 of the Talk server for Nextcloud 22, most of all it's multiple sessions per participant.

So I wanted to check and help migrating to APIv4, and now I wonder how we want to do this even.

As you can see in this PR there is one place which has duplicate logic for APIv3 and APIv1 already.

to deduplicate some logic and e.g. not have getRoom, getRoomV3, getRoomV4 functions and further on multiplying all methods in ApiUtils.java like e.g. getUrlForParticipants, I wonder if the version should be an (or even the first) argument to all the methods instead.

Also instead of code like:
```java
boolean isConversationApiV3 = userBeingCalled.hasSpreedFeatureCapability("conversation-v3");
        boolean isConversationApiV4 = userBeingCalled.hasSpreedFeatureCapability("conversation-v4");
        if(isConversationApiV4 || isConversationApiV3) {
```
I would add a method in the `userBeingCalled` class which returns the API we should use, so we don't duplicate those capabilities checks all the time everywhere.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -441,7 +441,14 @@ public class CallController extends BaseController {
     }
 
     private void handleFromNotification() {
-        ncApi.getRooms(credentials, ApiUtils.getUrlForGetRooms(baseUrl))
+        Integer apiVersion = ApiUtils.getApiVersion(conversationUser, "conversation", new int[] {1});
+
+        if (apiVersion == null) {
+            Log.e(TAG, "No supported API version found", new Exception("No supported API version found"));
+            return;
+        }
+
+        ncApi.getRooms(credentials, ApiUtils.getUrlForRooms(apiVersion, baseUrl))
                 .retry(3)
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
@@ -1234,9 +1241,16 @@ public class CallController extends BaseController {
     private void joinRoomAndCall() {
         callSession = ApplicationWideCurrentRoomHolder.getInstance().getSession();
 
+        Integer apiVersion = ApiUtils.getApiVersion(conversationUser, "conversation",
+                                                    new int[] {1});
+        if (apiVersion == null) {
+            Log.e(TAG, "No supported API version found", new Exception("No supported API version found"));
+            return;
+        }
+
         if (TextUtils.isEmpty(callSession)) {
-            ncApi.joinRoom(credentials, ApiUtils.getUrlForSettingMyselfAsActiveParticipant(baseUrl,
-                                                                                           roomToken), conversationPassword)
+            ncApi.joinRoom(credentials, ApiUtils.getUrlForParticipantsActive(apiVersion, baseUrl, roomToken),
+                           conversationPassword)
                     .subscribeOn(Schedulers.io())
                     .observeOn(AndroidSchedulers.mainThread())
                     .retry(3)
@@ -1648,7 +1662,14 @@ public class CallController extends BaseController {
     }
 
     private void leaveRoom(boolean shutDownView) {
-        ncApi.leaveRoom(credentials, ApiUtils.getUrlForSettingMyselfAsActiveParticipant(baseUrl, roomToken))
+        Integer apiVersion = ApiUtils.getApiVersion(conversationUser, "conversation",
+                                                    new int[] {1});
+        if (apiVersion == null) {
+            Log.e(TAG, "No supported API version found", new Exception("No supported API version found"));
+            return;
+        }
+
+        ncApi.leaveRoom(credentials, ApiUtils.getUrlForParticipantsActive(apiVersion, baseUrl, roomToken))
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe(new Observer<GenericOverall>() {
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -441,7 +441,7 @@ public class CallController extends BaseController {
     }
 
     private void handleFromNotification() {
-        Integer apiVersion = ApiUtils.getApiVersion(conversationUser, "conversation", new int[] {1});
+        Integer apiVersion = ApiUtils.getConversationApiVersion(conversationUser, new int[] {1});
 
         if (apiVersion == null) {
             Log.e(TAG, "No supported API version found", new Exception("No supported API version found"));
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java
```diff
@@ -127,7 +129,11 @@ public static String getRoom(String baseUrl, String token) {
     }
 
     public static String getRoomV3(String baseUrl, String token) {
-        return baseUrl + ocsApiVersion + "/apps/spreed/api/v3" + "/room/" + token;
+        return baseUrl + ocsApiVersion + spreedApiVersion3 + "/room/" + token;
```
AndyScherzinger, line 132: `"room"` should likely be extracted into a constant (private static final)

app/src/main/java/com/nextcloud/talk/controllers/CallNotificationController.java
```diff
@@ -261,8 +260,13 @@ public void onComplete() {
...
     private void handleFromNotification() {
         boolean isConversationApiV3 = userBeingCalled.hasSpreedFeatureCapability("conversation-v3");
-        if(isConversationApiV3) {
-            ncApi.getRoom(credentials, ApiUtils.getRoomV3(userBeingCalled.getBaseUrl(), roomId))
+        boolean isConversationApiV4 = userBeingCalled.hasSpreedFeatureCapability("conversation-v4");
+        if(isConversationApiV4 || isConversationApiV3) {
+            String apiUrl = ApiUtils.getRoomV4(userBeingCalled.getBaseUrl(), roomId);
+            if (!isConversationApiV4) {
+                apiUrl = ApiUtils.getRoomV3(userBeingCalled.getBaseUrl(), roomId);
+            }
```
AndyScherzinger, line 268: could be an if/else so only one url-concatenation is triggered

AndyScherzinger:
> to deduplicate some logic and e.g. not have getRoom, getRoomV3, getRoomV4 functions and further on multiplying all methods in ApiUtils.java like e.g. getUrlForParticipants, I wonder if the version should be an (or even the first) argument to all the methods instead.

I'd think so too

> I would add a method in the userBeingCalled class which returns the API we should use, so we don't duplicate those capabilities checks all the time everywhere.

should work for now, yes

I think @nickvergessen's approach is fine/pragmatic.

AndyScherzinger (approved):
just some minor comments/thoughts, in general fine with me 👍

app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java
nickvergessen: So as the v-version is bound to the capability and specific to the controller, I wondered i we should use conversationApiVersion4, etc. which then is `/apps/spreed/api/v4/room/`, because e.g. chat is still on v1, settings is v2, conversation and participants v4, etc.

AndyScherzinger: That would also be fine too imho. In general it is just that unique strings used in several places should be made a constant to prevent typos etc. in one/some place. Declaring larger parts of the url for the different endpoints would be fine too.

nickvergessen:
I now introduced a pattern like the following in java files:
https://github.com/nextcloud/talk-android/blob/f787cb6b716b50b8c4364e82eab509aaf0180dc4/app/src/main/java/com/nextcloud/talk/jobs/LeaveConversationWorker.java#L90-L100

However I'm unable to transfer the nullable value and logging to kt files e.g.
https://github.com/nextcloud/talk-android/blob/9e587a4385907d1e9d48fc430985deb915c65507/app/src/main/java/com/nextcloud/talk/controllers/ConversationInfoController.kt#L311

Can someone help me add the same thing there?

Api calls that don't check for a v4 first also log an error now. My idea (after the kt stuff is fixed) is that we can merge this and add v4 handling individually on each call going forward when we tested each call.

nickvergessen:
Also there are some places where no capabilties object seems to be available and I don't know how to get it:
https://github.com/nextcloud/talk-android/blob/bd8edbc17fb8c79bafe66658daef8e4321dc53d7/app/src/main/java/com/nextcloud/talk/controllers/CallController.java#L436

This might also be relating to the "join public conversations via links" which means we don't have a "UserEntity"?

nickvergessen:
Should be ready now, it does not actually implement APIv4 anywhere, just allows to specify the version and goes with 1 and 3 where it was used already

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -433,7 +433,14 @@ private void basicInitialization() {
     }
 
     private void handleFromNotification() {
-        ncApi.getRooms(credentials, ApiUtils.getUrlForGetRooms(baseUrl))
+        Integer apiVersion = ApiUtils.getApiVersion(conversationUser, "conversation", new int[] {1});
+
+        if (apiVersion == null) {
+            Log.e(TAG, "No supported API version found", new Exception("No supported API version found"));
```
AndyScherzinger, line 439: There is no reason to add an exception the message is enough since this only leads to printing the exception that offers no info that the original message didn't :)

app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java
```diff
@@ -100,41 +105,102 @@ public static RetrofitBucket getRetrofitBucketForContactsSearchFor14(String base
...
+    public static Integer getApiVersion(UserEntity capabilities, String apiName, int[] versions) {
+         if (apiName.equals("conversation")) {
+             boolean hasApiV4 = false;
+             for (int version : versions) {
+                 hasApiV4 |= version == 4;
+             }
+
+             if (!hasApiV4) {
+                 Exception e = new Exception("Api call did not try conversation-v4 api");
+                 Log.d(TAG, e.getMessage(), e);
```
AndyScherzinger, line 121: No need to create an exception just to log it in the next line. Also exceptions tend to be expensive especially when you throw them.

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
nickvergessen: Apart from the stack trace which helps you too find where the problem comes from?

nickvergessen: Since it's also kind of fatal and the code will not continue we could also throw?

AndyScherzinger: Of course it will continue, you didn't call throw exception!

nickvergessen: but we return :)

nickvergessen: Okay, so I add a throw everywhere. Might be saver anyway

AndyScherzinger: > Okay, so I add a throw everywhere. Might be saver anyway

If you throw it then don't log it ;) - A "typical" Java pattern would be "log or throw" as in if you "handle the exception" then fell free to log it (where you handle it)

> but we return :)

That is hacky ;)

> Okay, so I add a throw everywhere. Might be saver anyway

Yeah, that is the nicer way to handle it while "Exception" is very generic and will create warnings, so we might consider creating our own (or find a suiting one)

nickvergessen:
Not sure whats wrong with Codacy, it looks all good there, no new issues and stuff, but the action here is read...

nickvergessen:
Rebased after #1119

mahibi (approved):

mahibi:
👍👍

PR #1164, opened 2021-04-24
Enable enlarging the image more than default 3x maximumScale.

starypatyk:
I have just tested the new image preview - works great... :smiley: Thanks!

However I have found that sometimes it would be good to be able to enlarge the image more than the default 3x set in the PhotoView library. This applies esp. when posting photos directly from the phone.

This change enables zooming the image 6x rather than just 3x. Since the default double-tap handler uses the max/medium scale parameter, I have also adjusted the mediumScale to make double-tab more consistent/predictable.


mahibi (approved):

mahibi:
yes thats better, thanks @starypatyk

PR #1173, opened 2021-04-27
make navigationbar for videocall black

mahibi:

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/MagicCallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/MagicCallActivity.kt
@@ -58,6 +58,7 @@ class MagicCallActivity : BaseActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
+        setTheme(R.style.CallTheme)
 
         requestWindowFeature(Window.FEATURE_NO_TITLE)
         window.addFlags(
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -42,6 +42,10 @@
         <item name="android:navigationBarColor">@color/bg_default</item>
     </style>
 
+    <style name="CallTheme" parent="AppTheme">
+        <item name="android:navigationBarColor">@color/grey950</item>
+    </style>
+
     <style name="ErrorAppearance" parent="@android:style/TextAppearance">
         <item name="android:textColor">@color/nc_darkRed</item>
         <item name="android:textSize">12sp</item>


AndyScherzinger:
The better way would be to set "the right" theme, see
https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/activities/MainActivity.kt#L115

which would need to be applied right here
https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/activities/MagicCallActivity.kt#L60

This would apply the overall app's main theme which is (unfortunately) dark/light aware so it might also be an idea to create a separate definition just for the call activity which is "all black", else we could think about also make the call activity light in case of the light theme. Maybe the designer have an opinion on that too. :)

AndyScherzinger:
In Addition to my above comment it would need to be check what happens in case the app has already been launched and a call gets triggered from the outside (and ended) an dalso I I start a call within the app (and end it). To be checked: how do the other screens look after the call since the theme would need to be "reset/re-set" to the original/main one else the navigation bar would likely stay black which is not intended.

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -349,6 +349,8 @@ private void createCameraEnumerator() {
     protected void onViewBound(@NonNull View view) {
         super.onViewBound(view);
 
+        DisplayUtils.applyColorToNavigationBar(getActivity().getWindow(), Color.BLACK);
```
AndyScherzinger, line 352: looking at some screenshots it might me `#121212` which is the default dark theme background color and might be the one used for the screen too.

AndyScherzinger: Ich checkd, atm the background of the call screen is `#111111` or better `@color/grey950`

mahibi:
had a quick look without the commit from this PR... but i couldn't even figure out why the navigation bar currently is gone for outgoing audio calls, but it exists for video calls 🤔

AndyScherzinger:
> the navigation bar currently is gone for outgoing audio calls, but it exists for video calls

seems to be hidden sometimes/most-of-the-times (vide/audio) for me ?!

mahibi:
when receiving a call by notification the navbar is always gone (because )

> > the navigation bar currently is gone for outgoing audio calls, but it exists for video calls
> 
> seems to be hidden sometimes/most-of-the-times (vide/audio) for me ?!

for incoming calls by notification the navigation bar is always gone (which is okay). But for outgoing it's always gone for audio calls and always visible for video calls.

Anyway, i made a CallTheme which sets the navigationBarColor. And because there no Launcher defined for this theme the blue primary color and logo don't appear anymore. Guess that should work for now?

nickvergessen (approved):

AndyScherzinger:
> Anyway, i made a CallTheme which sets the navigationBarColor. And because there no Launcher defined for this theme the blue primary color and logo don't appear anymore. Guess that should work for now?

The launcher screen is shown rather as a glimpse so for triggered calls I'd also say we are just fine without it 👍

AndyScherzinger (approved):

PR #1210, opened 2021-05-02
Show search bar when closing bottom sheet

### LINKED ISSUE #1172: Searchbar gets replaced by toolbar
## Steps to reproduce
1. long press on a conversation until bottommenu appears
2. tap somewhere else

### Expected behaviour
searchbar should not change

### Actual behaviour
searchbar shows title "Nextcloud Talk" instead of searchbar and user avatar

AndyScherzinger:
Fixes #1172


AndyScherzinger:
@mahibi see the change, it'll basically always show the searchbar now. _If_ the controller can be called from anywhere and thus needs to offer a back navigation via the back arrow then we should add it to the searchbar and handle the state in the BaseController. Basically then detecting the state like in the deleted lien of code by checking the backstack or adding a method to set the state like the toolbar would allow too.

mahibi (approved):

mahibi:
great, thanks!

PR #1215, opened 2021-05-04
Add a warning when the server version is (almost) EOL

nickvergessen:
(padding was adjusted after taking screenshots)

Version | Bright | Dark
---|---|---
X < 4 (Talk 4 is Nextcloud 14 which is Summer 2018 and EOL since 2 years) | Bildschirmfoto von 2021-05-04 09-37-25 | Bildschirmfoto von 2021-05-04 09-37-30
4 <= X < 8 (Talk 8 is Nextcloud 18 which is beginning 2020 and EOL since march) | Bildschirmfoto von 2021-05-04 09-37-54 | Bildschirmfoto von 2021-05-04 09-37-46


nickvergessen:
Bildschirmfoto von 2021-05-04 10-05-26

Added a warning to the conversations list as well:
* Cancel opens the account switcher
* Remove deletes the account *surprise pikatchu meme*

app/src/main/java/com/nextcloud/talk/controllers/SettingsController.java
```diff
@@ -144,6 +144,8 @@
     EmojiTextView displayNameTextView;
     @BindView(R.id.base_url_text)
     TextView baseUrlTextView;
+    @BindView(R.id.server_age_warning_text)
```
nickvergessen, line 147: Lint increase comes from this:
> Resource IDs will be non-final in Android Gradle Plugin version 5.0, avoid using them as annotation attributes

Can we ignore it for now?

mahibi: ignore it by using
@SuppressLint("NonConstantResourceId") @BindView(R.id.server_age_warning_text)

these lint warnings will be gone when we switched to view binding https://developer.android.com/topic/libraries/view-binding

AndyScherzinger: Nah, please raise the score in the ```lint-results.txt``` file so the issue gets reported but doesn't break the build. We will migrate to native view bindings after 12.0 :)

AndyScherzinger (approved):
Looks good code-wise, while like mentioned I'd suggest raising the lint-score in the referenced file, see other comment.

app/src/main/java/com/nextcloud/talk/controllers/SettingsController.java
mahibi: otherwise raising the lint-score would most likely mean to often have another round of CI build..? But yeah, fine for me

AndyScherzinger: > otherwise raising the lint-score would most likely mean to often have another round of CI build..? But yeah, fine for me

Ignore would mean it is hidden in the reports, but might be fine, since compiler will terminate anyways as soon as we remove Butterknife dependency. You will always have a new round since changeign the results txt or the java/kotlin file is a commit/push ;)

mahibi:
>     * Cancel opens the account switcher

account switcher is blank if there is no other account. maybe disable the cancel button when there's no other account?

nickvergessen:
> account switcher is blank if there is no other account. maybe disable the cancel button when there's no other account?

Thinking of non tech people I would prefer to close the app or something in that case, because if they have a dialog with one button they will press it, instead of knowing they can "drop out to another app" or something?

mahibi:
> > account switcher is blank if there is no other account. maybe disable the cancel button when there's no other account?
> 
> Thinking of non tech people I would prefer to close the app or something in that case, because if they have a dialog with one button they will press it, instead of knowing they can "drop out to another app" or something?

yes should work. or maybe call it something like "upgrade instructions" and open https://nextcloud.com/install/#instructions-server in a browser?

nickvergessen:
Cancel now exists the app when there is only the EOL account.

AndyScherzinger:
How about some UI sugar-coating @nickvergessen ?

Version | Bright | Dark
---|---|---
X < 4 (Talk 4 is Nextcloud 14 which is Summer 2018 and EOL since 2 years) | error_light | error_dark
4 <= X < 8 (Talk 8 is Nextcloud 18 which is beginning 2020 and EOL since march) | warning_light | warning_dark

AndyScherzinger (approved):

PR #1216, opened 2021-05-04
A11y content description

AndyScherzinger:
```$subject```, basically adding content descriptions (static/dynamic) to any image-button/-view where it makes sense (and else set ```@null``` so a screen reader skips it like icons for visual support of text-fields that already have a descriptive hint text which would otherwise duplicate this info for a reader).

### CODE REVIEW

File: app/src/main/res/values/strings.xml [hunk 1]
```diff
@@ -128,7 +128,8 @@
     <string name="nc_settings_link_previews_desc">Allows previews of content from received links for supported services</string>
     <string name="nc_settings_link_previews_key" translatable="false">link_previews</string>
     <string name="nc_settings_read_privacy_key" translatable="false">read_privacy</string>
-    <string name="nc_locked">Tap to unlock</string>
+    <string name="nc_locked_tap_to_unlock">Tap to unlock</string>
+    <string name="nc_locked">locked</string>
```
jancborchardt, line 132: Should this be `Locked`? Or `translatable="false"`, unsure. :)

app/src/main/res/values/strings.xml
```diff
@@ -232,6 +236,7 @@
     <string name="nc_ok">OK</string>
     <string name="nc_call_name">Conversation name</string>
     <string name="nc_proceed">Proceed</string>
+    <string name="nc_add_emojis">Add Emojis</string>
```
jancborchardt, line 239:
```suggestion
    <string name="nc_add_emojis">Add emojis</string>
```
"emojis with lowercase e :)

app/src/main/res/values/strings.xml
```diff
@@ -398,6 +409,7 @@
     <string name="scope_published_title">Published</string>
     <string name="scope_published_description">Synchronize to trusted servers and the global and public address book</string>
     <string name="scope_toggle">Scope toggle</string>
+    <string name="scope_toggle_description">Scope toggle for %1$s (%2$s)</string>
```
jancborchardt, line 412: In the server we use a less technical wording here: "Change privacy level of […]"
Ref https://github.com/nextcloud/server/pull/26548/files#diff-c1bbc91dbbd993cc8d0ee7bcef864eb9290be2fa29dd4810d92f1f3a19449183

app/src/main/res/values/strings.xml
AndyScherzinger: It should be ```Locked``` 👍

app/src/main/res/values/strings.xml
AndyScherzinger: Done!

app/src/main/res/values/strings.xml
AndyScherzinger: DONE, thanks for the hint&link 👍

mahibi (approved):

PR #1218, opened 2021-05-05
[apiv4] Remove props which are not there anymore

nickvergessen:
- [x] Based on #1156 
- [x] count prop is deprecated and removed in apiv4 => unused => removed on Conversation object
- [x] numGuests prop is deprecated and removed in apiv4 => unused => removed on Conversation object
- [x] guestList prop is deprecated and removed in apiv4 => unused => removed on Conversation object
- [x] participants prop is deprecated and removed in apiv4
    - [x] still used in various places


AndyScherzinger (approved):
looks good to me code-wise

PR #1221, opened 2021-05-06
Properly parse encryption/favorite flags

AndyScherzinger:
encryption and favorite property are delivered via 0/1 values not TRUE/FALSE thus the parser needs to check for that.

To test simply have files on your Nextcloud marked as favorite and try to add a file via Nextcloud in the chat. 
Before: No start icon
After: star icon


nickvergessen (approved):
As discussed

mahibi (approved):

PR #1229, opened 2021-05-07
[apiv4] ☎️ Calls

nickvergessen:
- [x] Join call
- [x] Leave call
- [ ] ~~Rejoin call~~ => broken also on master => https://github.com/nextcloud/talk-android/issues/1232
- [x] Can do an audio call
- [x] Can do a video call (works sometimes)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/CallController.java
@@ -1273,7 +1273,7 @@ public class CallController extends BaseController {
             inCallFlag = (int) Participant.ParticipantFlags.IN_CALL_WITH_AUDIO_AND_VIDEO.getValue();
         }
 
-        int apiVersion = ApiUtils.getCallApiVersion(conversationUser, new int[] {1});
+        int apiVersion = ApiUtils.getCallApiVersion(conversationUser, new int[] {ApiUtils.APIv4, 1});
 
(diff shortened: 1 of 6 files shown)


nickvergessen:
Not sure what's up with video calls, they seem equally unreliable in master as they do on this branch. Rest is fixed or reported.
Can go in from my POV.

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -2044,11 +2050,11 @@ private void setupVideoStreamForLayout(@Nullable MediaStream mediaStream, String
...
         }
 
-        String userId;
+        String userId = "";
         if (hasMCU) {
             userId = webSocketClient.getUserIdForSession(session);
-        } else {
-            userId = participantMap.get(session).getUserId();
+        } else if (participantMap.get(session).getActorType() == Participant.ActorType.USERS) {
+            userId = participantMap.get(session).getActorId();
```
AndyScherzinger, line 2057: is there a general fall back? because right now the userId can be an empty string if none of the ifs turns out to be true.

AndyScherzinger:
The code seems (mostly) fine whiel I keep failing to understand hwo this works againt a non-v4 endpoint (like v1)

app/src/main/java/com/nextcloud/talk/controllers/CallController.java
nickvergessen: before getUserId fell back and now there is a default value on the initialization

PR #1235, opened 2021-05-10
[apiv4] 👤🏷️ Fix API calls for guests

nickvergessen:
- [x] Based on #1229 
- [x] Fix opening a public conversation in `OperationsMenuController`
- [x] Drop initiateCall in `OperationsMenuController`
- [x] Fix making a call as guest
- [x] Guests posting messages are not detected as "self"
- [ ] Adjust checkCapabilities() in `CallController`

Best reviewed as https://github.com/nextcloud/talk-android/pull/1235/files?w=1


nickvergessen:
There is still a double checking of capabilities between the chatcontroller and the callcontroller, but I think we can merge it like this for now and fix the rest in another PR. It's just about getting rid of one more request, but I broke my brain 3 times this week on the circular calling already.

mahibi (approved):

PR #1237, opened 2021-05-10
Explicitely set header color

AndyScherzinger:
Conversation name visibility fix backported from #1236


AndyScherzinger:
Please ignore codacy which seems to miss a config so it also watches stable branches...

nickvergessen (approved):

PR #1240, opened 2021-05-11
run drone on stable-11 branch

### LINKED ISSUE #1238: Align stable branch naming pattern with files app
I added https://github.com/nextcloud/talk-android/pull/1237/commits/2d72c198aebf10dccede27c0c80a3fce4977d293 for the `v11.x` branch so all our Github actions also run on PRs targeting a stable branch. It didn't work before since (as you can see) the naming pattern on the Android talk repo is different from the files app.

I order to make it easier to re-use solutions we have in palce regarding automation to adapt the files app's naming schema.

So all stable branches should be renamed, e.g. `v11.x ` --> `stable-11`
You might spot the slight difference since the files app also adds the minor version number. This is due to the fact that the files app doesn't follow a version scheme that must match the server version. It also doesn't matter for our automation since `stable-*` will work just fin in both scenarios. 😄 

cc @mahibi @nickvergessen and (due to a possible impact on brander app) @tobiasKaminsky (just to be sure/save)

AndyScherzinger:

Resolves #1238

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,7 +9,7 @@ steps:
 
 trigger:
   branch:
-    - master
+    - stable-11
   event:
     - pull_request
     - push
@@ -26,7 +26,7 @@ steps:
 
 trigger:
   branch:
-    - master
+    - stable-11
   event:
    - push
    - pull_request
@@ -75,11 +75,11 @@ steps:
     status:
       - failure
     branch:
-      - master
+      - stable-11
 
 trigger:
   branch:
-    - master
+    - stable-11
   event:
     - push
     - pull_request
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
> @tobiasKaminsky will this show up as soon as this PR gets merged? [#1240 (comment)](https://github.com/nextcloud/talk-android/pull/1240#issuecomment-838499167)

Yes, it should(tm).
Upon a valid merge the summary is created and uploaded.

nickvergessen:
The translation update is "broken", it removes strings still used in stable-11 but not in master.

AndyScherzinger:
> The translation update is "broken", it removes strings still used in stable-11 but not in master.

@nickvergessen still is solved for the files app and we can port this whenever we have transifex updates for stable branches afaik we don't push transifex changes to the stable branches but only to master, no?

mahibi (approved):

PR #1242, opened 2021-05-11
[backport] set proper colors for unread-messages for edge case of theme switch

AndyScherzinger:
Manual backport and theme switch follow-up PR.

Besides the conversation name also the unread messages bubble/text haven't refreshed properly (the unread, but no mentions bubble as-in grey, not blue)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
@@ -23,6 +23,7 @@
 package com.nextcloud.talk.adapters.items;
 
 import android.content.Context;
+import android.content.res.ColorStateList;
 import android.graphics.Color;
 import android.graphics.Typeface;
 import android.graphics.drawable.BitmapDrawable;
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -159,8 +159,6 @@
     <string name="nc_make_call_private">Unterhaltung privat machen</string>
     <string name="nc_delete_call">Unterhaltung löschen</string>
     <string name="nc_delete">Löschen</string>
-    <string name="nc_delete_conversation_default">Bitte bestätigen Sie, dass Sie diese Unterhaltung entfernen möchten.</string>
```
nickvergessen, line 162: those strings are still used in 11 😕

AndyScherzinger: Don't know really, linter warns about a couple of unused string though we might remove them in the near future (for v12 then)

nickvergessen: I removed them in https://github.com/nextcloud/talk-android/pull/1218/files#diff-2f516f16cffd61f534bc36fffec911d1b2909323ee3717ddbeec5db40dd4264cL138 which is not going to be backported

AndyScherzinger: rebased and fixed

PR #1245, opened 2021-05-12
✉️ & ⭕ Allow to add emails and circles as particpants

nickvergessen:
1. Start a conversation
2. Go to conversation info
3. Tab "Add participant"
4. Search for an email address or circle
5. Select it and confirm with "Done"
6. Email is invited to the conversation

Closing another checkbox in the iOS <=> Android comparison

PR #1246, opened 2021-05-12
↖️ Reply privately

nickvergessen:
1. Long press on a chat message of another user in a group conversation
2. Select "Reply privately"
3. The one-to-one conversation with that user is being opened like in web and iOS


app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -1615,6 +1620,66 @@ class ChatController(args: Bundle) :
...
                         true
                     }
+                    R.id.action_reply_privately -> {
+                        val apiVersion = ApiUtils.getConversationApiVersion(conversationUser, intArrayOf(ApiUtils.APIv4, 1))
+                        val retrofitBucket = ApiUtils.getRetrofitBucketForCreateRoom(
+                            apiVersion,
+                            conversationUser?.baseUrl,
+                            "1",
+                            message?.user?.id?.substring(6),
+                            null
```
nickvergessen, line 1630: This conflicts with #1245 
When that is merged first we need to add a `null` before the userid

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -1615,6 +1620,66 @@ class ChatController(args: Bundle) :
...
+                                        .observeOn(AndroidSchedulers.mainThread())
+                                        .subscribe(object : Observer<RoomOverall> {
+                                            override fun onSubscribe(d: Disposable) {}
+                                            override fun onNext(roomOverall: RoomOverall) {
+                                                bundle.putParcelable(
+                                                    KEY_ACTIVE_CONVERSATION,
+                                                    Parcels.wrap(roomOverall.getOcs().getData())
+                                                )
+                                                remapChatController(
+                                                    router, conversationUser!!.id,
```
mahibi, line 1664: better to use one from the 3 other suggestions:
grafik

only use !! when 100% sure it can't be null

nickvergessen: It can not be null as only logged in users should see the option and guests can't end up in here as also the createRoom would fail for them when they would call it.

But I fixed an issue that didn't hide the option when "Join with a link"

mahibi (approved):

nickvergessen:
Bildschirmfoto von 2021-05-14 13-54-08

Now properly shows the icon as well

AndyScherzinger:
> Now properly shows the icon as well

...on Android 10+ only 😉

PR #1249, opened 2021-05-14
🎙️ Fix participant list call state icon

nickvergessen:
Before | After
---|---
1 icon for audio 1 icon for video ment to overlay the avatar like a favorite icon | Most important state is shown: 1. SIP dialin 2. Video otherwise Mic
Bildschirmfoto von 2021-05-12 18-35-15 | Bildschirmfoto von 2021-05-14 09-51-09


app/src/main/java/com/nextcloud/talk/adapters/items/UserItem.java
```diff
@@ -186,56 +187,27 @@ public void bindViewHolder(FlexibleAdapter adapter, UserItemViewHolder holder, i
...
+            Long inCallFlag = participant.getInCall();
+            if ((inCallFlag & InCallFlags.WITH_PHONE) > 0) {
+                holder.videoCallIconView.setBackground(
+                        ResourcesCompat.getDrawable(resources, R.drawable.ic_call_grey_600_24dp, null));
+                holder.videoCallIconView.setVisibility(View.VISIBLE);
+                holder.videoCallIconView.setContentDescription(
+                        resources.getString(R.string.nc_call_state_with_phone, participant.displayName));
+            } else if ((inCallFlag & InCallFlags.WITH_VIDEO) > 0) {
+                holder.videoCallIconView.setBackground(
+                        ResourcesCompat.getDrawable(resources, R.drawable.ic_videocam_grey_600_24dp, null));
```
nickvergessen, line 199: before this used things like: `shape_video_bubble`
Is it better to use the ic_ directly vs. the shape_ ?

AndyScherzinger: In this case, we shouldn't use the bubble which has likely been used since the icon would have been on the avatar and depending on the avatar quite hard to read/see. Since you put it to the right of the whole item we are fine since it is on the backgound (`#ffffff`or `#121212`) so we are fine 👍

AndyScherzinger (approved):

AndyScherzinger (approved):

PR #1251, opened 2021-05-14
Use the stun urls array

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


app/src/main/java/com/nextcloud/talk/controllers/CallController.java
```diff
@@ -1146,12 +1146,14 @@ public void onNext(SignalingSettingsOverall signalingSettingsOverall) {
...
                                      i++) {
                                     iceServer = signalingSettingsOverall.getOcs().getSettings().getStunServers().get(i);
-                                    if (TextUtils.isEmpty(iceServer.getUsername()) || TextUtils.isEmpty(iceServer
-                                                                                                                .getCredential())) {
-                                        iceServers.add(new PeerConnection.IceServer(iceServer.getUrl()));
-                                    } else {
-                                        iceServers.add(new PeerConnection.IceServer(iceServer.getUrl(),
-                                                                                    iceServer.getUsername(), iceServer.getCredential()));
+                                    for (int j = 0; j < iceServer.getUrls().size(); j++) {
+                                        if (TextUtils.isEmpty(iceServer.getUsername()) || TextUtils.isEmpty(iceServer
```
AndyScherzinger, line 1150: I'd probably flip the logic and check for username/credentials first else you do this in each iteration on the loop while the result is always the same. The downside is having two nearly identical loops.

nickvergessen: Actually stun never has credentials and turn always. Also we will use a v3 api for it, currently discussing in the server pr, so more changes incoming here

AndyScherzinger (approved):
Tested and works fine for me on `sermo` instance

PR #1252, opened 2021-05-15
add ability to "share to" files to nextcloud talk app

### LINKED ISSUE #308: "Share with ..." function (Android)
Hi
I was wondering why I can't find Nextcloud Talk in my list when I try to share sth. like in my Brave Browser (LOS 15.1)? Did I miss something or this right?
I would be really happy if there'll be this feature to simply share links directly with my contacts.
Thanks a lot!

mahibi:
resolve #308


nickvergessen:
Did a test against sermo, works nicely

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
```diff
@@ -764,6 +833,117 @@ public void onItemLongClick(int position) {
...
+                    Log.e(TAG, "Something went wrong when extracting data from intent");
+                }
+            }
+        }
+    }
+
+    private void upload() {
+        if (selectedConversation == null) {
+            Log.e(TAG, "not able to upload any files because conversation was null.");
+            return;
```
tobiasKaminsky, line 871: Is there any info to user?

tobiasKaminsky (approved):

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
mahibi: will add a toast 👍

PR #1264, opened 2021-05-17
Remove custom password animation and use default/component one

AndyScherzinger:
Try to join a conversation via link with a passowrd and check the animation on the visibility toggle of the password field.

The material component (as of today) handles this quite fine on its own so there is no need to "hack" it's animation/icons (anymore).

PS: I could have removed the tools-namespace declaration but left it in so @MorrisJobke's script for translation-merge-magic doesn't get into trouble. While ultimately the magical adding of the namespace declaration could actually be removed after this PR has been merged.


MorrisJobke:
> PS: I could have removed the tools-namespace declaration but left it in so @MorrisJobke's script for translation-merge-magic doesn't get into trouble. While ultimately the magical adding of the namespace declaration could actually be removed after this PR has been merged.

You can remove it from here, as we add it directly in the translation script itself. If you backport this change also to stable-11 or once that branch is not in the translation script anymore then we can also remove the prefix from the translation script.

AndyScherzinger:
> You can remove it from here, as we add it directly in the translation script itself. If you backport this change also to stable-11 or once that branch is not in the translation script anymore then we can also remove the prefix from the translation script.

Removed it within this PR and flagged it for backport to v11

nickvergessen (approved):
Ha, I was confused about that animation in #1235 already

mahibi (approved):

PR #1283, opened 2021-05-21
fix displaying of large images

### LINKED ISSUE #1190: Not displaying images
## Steps to reproduce

1. Receive the image in the conversation.
2. Tap on an image to view it in full size.

The image is not displayed. Only a black background with three dots appears (kebab menu).

I also noticed that after tapping on the image there is no reaction for a long time. This confuses the user. Doesn't know if the application worked properly. In my opinion, there should be a loading wheel or other information about the image loading.

## Device information

**Android version:** 10

**Talk version:** 12.0.0 Alpha 08

mahibi:
fix #1190 

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -186,6 +186,7 @@ dependencies {
     implementation "androidx.work:work-runtime:${workVersion}"
     implementation "androidx.work:work-rxjava2:${workVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
+    implementation 'androidx.exifinterface:exifinterface:1.3.2'
     androidTestImplementation "androidx.work:work-testing:${workVersion}"
     implementation 'com.google.android:flexbox:2.0.1'
     implementation ('com.gitlab.bitfireAT:dav4jvm:2.1.2', {
--- a/app/src/main/java/com/nextcloud/talk/utils/BitmapShrinker.kt
+++ b/app/src/main/java/com/nextcloud/talk/utils/BitmapShrinker.kt
@@ -1,0 +1,112 @@
+/*
+ * Nextcloud Talk application
+ *
+ * @author Marcel Hibbe
+ * Copyright (C) 2021 Marcel Hibbe <dev@mhibbe.de>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.nextcloud.talk.utils
+
+import android.graphics.Bitmap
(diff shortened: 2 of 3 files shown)


nickvergessen:
Couldn't reproduce an issue before, but code is in the files app, so should be fine.

Tested with https://eoimages.gsfc.nasa.gov/images/imagerecords/73000/73751/world.topo.bathy.200407.3x5400x2700.jpg

app/src/main/java/com/nextcloud/talk/utils/BitmapShrinker.kt
```diff
@@ -0,0 +1,119 @@
...
+                0,
+                0,
+                bitmap.getWidth(),
+                bitmap.getHeight(),
+                matrix,
+                true
+            )
+            return rotatedBitmap
+        } catch (e: IOException) {
+            Log.e(TAG, "error while rotating image")
```
AndyScherzinger, line 115: I'd add the exception as a 3rd parameter

mahibi: > Couldn't reproduce an issue before, but code is in the files app, so should be fine.

infact i went for an other solution: https://developer.android.com/topic/performance/graphics/load-bitmap

app/src/main/java/com/nextcloud/talk/utils/BitmapShrinker.kt
```diff
@@ -0,0 +1,119 @@
...
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.graphics.Matrix
+import android.media.ExifInterface
+import android.util.Log
+import java.io.IOException
+
+/**
+ * https://developer.android.com/topic/performance/graphics/load-bitmap.
+ * https://stackoverflow.com/a/15341203
```
AndyScherzinger, line 32: Stack overflow uses https://stackoverflow.com/help/licensing while I don't know if this leads to any issues in combination with GPLv3+ and also I don't know which license android.com uses. So I guess it can't be re-licensed as GPLv3  (both sources) and I don't know if GPLv3 and CC-BY can code exist.

Looping in @schiessle for some licensing feedback ❤️

mahibi:
> Couldn't reproduce an issue before, but code is in the files app, so should be fine.

infact i went for an other solution: https://developer.android.com/topic/performance/graphics/load-bitmap

app/build.gradle
```diff
@@ -186,6 +186,7 @@ dependencies {
     implementation "androidx.work:work-runtime:${workVersion}"
     implementation "androidx.work:work-rxjava2:${workVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
+    implementation 'androidx.exifinterface:exifinterface:1.3.0-rc01'
```
AndyScherzinger, line 189: @mahibi Any reason for using ```1.3.0-rc01```? I'd advice to use latest which according to https://developer.android.com/jetpack/androidx/releases/exifinterface is ```1.3.2```

AndyScherzinger (approved):
Looks good just one minor comment regarding the lib-version

AndyScherzinger:
@mahibi and dete3kt isn't happy... -> https://github.com/nextcloud/talk-android/pull/1283/checks?check_run_id=2782243685

app/build.gradle
mahibi: ah thx. came in via auto suggestion

mahibi:
> @mahibi and detekt isn't happy... -> https://github.com/nextcloud/talk-android/pull/1283/checks?check_run_id=2782243685

fixed the detect issues for the 2 files of this PR

PR #1284, opened 2021-05-21
Feature/1136/share location

### LINKED ISSUE #1136: Share a location
Server ref https://github.com/nextcloud/spreed/issues/1478

It should be possible to share a location from e.g. Google Maps to a Talk chat
Also locations shared by others should be displayed somehow and allow to open in a Map app

mahibi:
fix #1136 

remaining improvements / known bugs:
- [x] replace current-location icon
- [x] avoid to make contents of location description (e.g. zip codes) clickable as telephone number
   - i tried to suppress that zip codes are shown as clickable telephopne numbers. but
     android:autoLink="none"
    and
     messageText!!.linksClickable = false
    don't seem to work.
    as a workaround i disabled the text field completely. so the links are still underlined but not clickable. and it's also not possible  now to mark text and get the context menu.
- [x] improve design of  share location button / location description 
- [x] check if server supports location sharing (hide option in bottom menu)
- [x] copyrights
- [ ] fix context menu of selected text in geocoding search bar 
  - grafik
- [ ] fix search result display/map being reset on screen rotation
- [x] remove delete from message popup for location shares, see https://github.com/nextcloud/spreed/pull/5739
- [x] fix darkmode textcolor for geocoding results
- [x] view binding
- [x] fix lint
   - most should be done by view binding
   - Duplicate Platform Classes ?


mahibi:
@nickvergessen @Ivansss 
right now i show the location on the map if it was selected from the geocoded results. Sharing without to see it on the map (like it was planned in the mockup) felt quite wrong for me. What's your opinion on this? 

@Ivansss if you have the possibility to test, please take a moment to check if feature parity is okay..

mahibi:
> Does this include the pin not being vertically centered when there is no description/address?

yep

mahibi:
and as you might have noticed @nickvergessen :
when selecting a location by dragging, the reverse geocoding is only done when the location is actually shared.
so there are no reverse geocoding requests for every dragging to avoid stressing nominatim. you're okay with that?

AndyScherzinger (approved):

Issue #1029, opened 2021-02-24
display error when using emojis


theoneandonly-vector:
typing works correctly using the keyboard, but as soon as I select emojis it looks like this (you don't see what you type):
![Screenshot_20210224-101120](https://user-images.githubusercontent.com/10380051/108977713-1fbcdc00-7689-11eb-94ee-7f4a1b94c05e.jpg)

this doesn't happen in other apps I use.

it may happen because of this setting I use:
wm overscan 0,0,0,-70
(this removes the keyboard-gap on oneplus-devices)

mahibi:
hi @theoneandonly-vector 
which version of the app are you using? was it also wrong with older versions of the app?

theoneandonly-vector:
hi @mahibi 
I use the latest release-version (11.0.0), but the issue also appears on older verisons.

jakobroehrl:
On my device I can't use the groups, nothing happens on click. It seams the android keyboard settings are overlaying them:
![image](https://user-images.githubusercontent.com/24530680/109287961-8d513f80-7824-11eb-93f8-2211468109ea.png)

mahibi:
@jakobroehrl what's 
your device 
/android version
/app version?

dorpauli:
@mahibi 
I got the same problem like @jakobroehrl but i can still navigate through emojis with switching. 

Device: OnePlus 7T
Android Version: 10 (Oxygen Os 10.0.16.HD65BA)
App Version: 11.0.0

mahibi:
i will close this one as it is the same error as https://github.com/nextcloud/talk-android/issues/1900
didn't find any solution yet..

PR #1285, opened 2021-05-22
fix design when keyboard is shown

### LINKED ISSUE #992: Top bar hides partially under status bar when typing
In 11.0.0 Alpha 04, I get a strange behaviour. The top bar moves half way up and hides under the status bar when I tap on the input field.

Phone: Oneplus 8
Android: 11
OS: Oxygen OS 11.0.3.3.IN21BA

Screenshot_20210202-164651_2

mahibi:
fix #992
might also be a fix for #1029 

basically with this PR issue #992 is fixed by setting `windowSoftInputMode` to `adjustResize` and deleting the `KeyboardUtils`.
This in turn caused conflicts with https://github.com/Kennyc1012/BottomSheetMenu which was now replaced with androids [BottomSheetDialog](https://developer.android.com/reference/com/google/android/material/bottomsheet/BottomSheetDialog)
The replacing of the bottom dialog lib led to quite some refactoring.

For now, while using the BottomSheetDialog there are some workarounds (=add more routers) to keep using the controllers from conductor. 

fixed design in conversation | conversation bottom dialog | input dialog moved over keyboard  
------- | -------- | -------- 
grafik | grafik | grafik


mahibi:
deleting the KeyboardUtils will lead to a bug that the BottomSheet (https://github.com/Kennyc1012/BottomSheetMenu) is hidden under the keyboard (for example when entering a name for a new group).

I guess the best solution is to get rid of the BottomSheet lib. It even doesn't support customized views any more in newer versions, see https://github.com/Kennyc1012/BottomSheetMenu#upgrading-to-3x
The bug with the keyboard hiding the BottomSheet was known (see https://github.com/Kennyc1012/BottomSheetMenu/issues/55) but there was never a solution by this lib, so that's where the KeyboardUtils made sense i guess.

So we should try to replace this lib with other solutions like android's BottomSheetDialogFragment 
( https://developer.android.com/reference/com/google/android/material/bottomsheet/BottomSheetDialogFragment )

I remember we had the idea to unify the "creation of groups" between iOS and android at the hackweek meeting @nickvergessen 
So if there are anyway any other ideas instead to have a menu at the bottom it might be a win win to also get rid of the bug that the Actionbar is partially hidden whenever keyboard is shown..

AndyScherzinger:
@mahibi looking good, I did add some color improvements over the weekend for the bottom sheets for dark/light theming and improved the detekt/ktlint scores. Hope it helps. 💙

mahibi:
> @mahibi looking good, I did add some color improvements over the weekend for the bottom sheets for dark/light theming and improved the detekt/ktlint scores. Hope it helps. blue_heart

cool, thanks @AndyScherzinger :+1:

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/EntryMenuController.java
```diff
@@ -143,54 +144,49 @@ protected void onAttach(@NonNull View view) {
...
 
             getRouter().pushController(RouterTransaction.with(new OperationsMenuController(bundle))
                     .pushChangeHandler(new HorizontalChangeHandler())
                     .popChangeHandler(new HorizontalChangeHandler()));
-        } else if (operationCode != 7 && operationCode != 10 && operationCode != 11) {
-            eventBus.post(new BottomSheetLockEvent(false, 0, false, false));
+        } else if (operation != ConversationOperationEnum.SHARE_LINK && operation != ConversationOperationEnum.GET_AND_JOIN_ROOM && operation != ConversationOperationEnum.INVITE_USERS) {
             bundle = new Bundle();
-            if (operationCode == 4 || operationCode == 6) {
+            if (operation == ConversationOperationEnum.CHANGE_PASSWORD || operation == ConversationOperationEnum.SET_PASSWORD) {
```
timkrueger, line 162: Can you add an on-demand static import for `ConversationOperation` enum?

```
import static com.nextcloud.talk.controllers.bottomsheet.ConversationOperationEnum.*;
```

In my opinion it will be much more readable.

timkrueger (approved):

mahibi:
> @mahibi looks good, one thing though (at least on Android 12 on my OP9Pro): i.e. when marking a conversation as a favorite, after the "success animation" the bottom sheet isn't closed automatically anymore. This has been working before, so maybe that part didn't get migrated?

indeed that was not migrated! fixed now

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/EntryMenuController.java
mahibi: done

AndyScherzinger (approved):
@mahibi works like a charm, just needs a single ktlint fix 👍

mahibi:
> @mahibi works like a charm, just needs a single ktlint fix +1

@AndyScherzinger well the klint warning is exactly about the wildcard import suggested by @timkrueger 
https://github.com/nextcloud/talk-android/pull/1285#discussion_r807041353

So should we
- avoid wildcard imports?
- change klint config to allow them?
- yolo and merge? :D

timkrueger:
> import every single enum as static?

I would prefer to import every single enum value as static.

AndyScherzinger:
Yeah, no wildcards, also 'mark as read' in the conversation list bottom sheet can't be clicked @mahibi

mahibi:
we need to get rid of:

```
Could not load custom lint check jar file /root/.gradle/caches/transforms-2/files-2.1/e0e0d00b63718648d7473e1d0cc948ee/jetified-conductor-3.1.3/jars/lint.jar
--

java.lang.UnsupportedClassVersionError:
 com/bluelinelabs/conductor/lint/ControllerIssueDetector has been 
compiled by a more recent version of the Java Runtime (class file 
version 55.0), this version of the Java Runtime only recognizes class 
file versions up to 52.0

```

https://drone.nextcloud.com/nextcloud/talk-android/3014/4/2

PR #1300, opened 2021-05-28
Rebase action: /rebase as comment

AndyScherzinger:
shameless c&p from https://github.com/nextcloud/android/pull/8488 where it has been shamelessly copied from the desktop repo ;)

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,0 +1,18 @@
+on:
+    issue_comment:
+        types: [ created ]
+name: Automatic Rebase
+jobs:
+    rebase:
+        name: Rebase
+        if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/rebase')
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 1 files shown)


mahibi (approved):

szaimen:
cc @skjnldsv maybe this could be something for the server repository, too?

AndyScherzinger:
@szaimen @skjnldsv it then be nice if somebody could take a look into https://github.com/nextcloud/android/pull/8488/files#r640384338 and upgrade the GH action to use the latest version. Being lazy I just c&p-ed it while mentioning that it should be updated at some point.

skjnldsv:
Damn, yeah!
But those github actions are getting complicated to maintain.
We really need a template repo and a script to copy/update some basics (like fixup check, node, etc etc)

I guess the script would create a PR updating the file if there is a new update from the template repo and the filename matches? :thinking: 

i'm sure other would love it

AndyScherzinger:
And on the template repo we could run an updater action like: https://github.com/marketplace/actions/github-actions-version-updater

PR #1331, opened 2021-06-16
Feature/1137/voice messages

### LINKED ISSUE #1137: Add voice messages
Server issue https://github.com/nextcloud/spreed/issues/1576

mahibi:
moved to https://github.com/nextcloud/talk-android/pull/1340 (because of rebase problems)


app/src/main/java/com/nextcloud/talk/jobs/ShareOperationWorker.java
```diff
@@ -70,12 +75,16 @@ public ShareOperationWorker(@NonNull Context context, @NonNull WorkerParameters
...
+        String metaData = "{\"messageType\":\"voice-message\"}";
+
         for (int i = 0; i < filesArray.size(); i++) {
             ncApi.createRemoteShare(credentials,
                     ApiUtils.getSharingUrl(baseUrl),
                     filesArray.get(i),
                     roomToken,
-                    "10")
+                    "10",
+                     metaData)
```
nickvergessen, line 87: This is only done for voice-messages, right? Or are you flagging all as such?

app/src/main/res/values/strings.xml
```diff
@@ -380,6 +382,10 @@
     <string name="nc_share_this_location">Share this location</string>
     <string name="nc_location_current_position_description">Your current location</string>
 
+    <!-- voice messages -->
+    <string name="nc_voice_message_filename">Talk recording from</string>
```
nickvergessen, line 386: I would make the date and name placeholder part of the string, in case it is needed for a translation.
See https://github.com/nextcloud/spreed/pull/5610/files#diff-59506255a83039f3102821486034814a784ad7022780a1ebdb22146b184d3970R282 for inspiration

nickvergessen (changes requested):
- [ ] Leaving the chat keeps playing the message. But if oyu go back in there is no way to pause it. I think we should stop playing when the back button is used to leave the chat?
- [x] Control is not visible for own messages (blue on blue with a blue progress bar)
Screenshot_2021-06-17-08-11-30-492_com nextcloud talk2 qa
- [x] The play/pause button should be white in dark mode:



---

Optional:
- My wife asked for a "lock" function, so you don't need to keep pressing the button. Whatsapp does it on slide up, everyone else by default.

app/src/main/java/com/nextcloud/talk/jobs/ShareOperationWorker.java
mahibi: right this was just for testing. changed this now that only voice messages are flagged

AndyScherzinger:
@mahibi @nickvergessen I put any UI/code-check fixes/improves onto #1340

PR #1342, opened 2021-06-22
Show usage of interface for communication between slider position and chat controller

tobiasKaminsky:


.idea/codeStyles/Project.xml
```diff
@@ -212,9 +212,8 @@
       <option name="KEEP_BLANK_LINES_BEFORE_RBRACE" value="0" />
       <option name="ALIGN_MULTILINE_PARAMETERS" value="false" />
       <indentOptions>
-		<option name="INDENT_SIZE" value="4" />
-		<option name="CONTINUATION_INDENT_SIZE" value="4" />
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
```
tobiasKaminsky, line 215: This is due to a different Android Studio version, please ignore it :)

AndyScherzinger:
I'd say we are trying to work around a design issue here. Imho the mediaPlayer shouldn't be handled by the view holder, it should be a singleton/service originally provided by the ChatController, and provided to the Adapter, etc. So the media player doesn't get "null-ed" at each completion and re-created for every voice message play. This way the Controlelr is in charge of the resource and can stop/kill the mediaPlayer when the Controller gets paused/destroyed.

What do you think? Having the player as a service or handled by the controller is a bigger change though.

mahibi:
> I'd say we are trying to work around a design issue here. Imho the mediaPlayer shouldn't be handled by the view holder, it should be a singleton/service originally provided by the ChatController, and provided to the Adapter, etc. So the media player doesn't get "null-ed" at each completion and re-created for every voice message play. This way the Controlelr is in charge of the resource and can stop/kill the mediaPlayer when the Controller gets paused/destroyed.
> 
> What do you think? Having the player as a service or handled by the controller is a bigger change though.

As discussed 1:1 with @AndyScherzinger : 
Yes absolutely, the mediaPlayer shouldn't be inside the ViewHolder!
Already had a look on how move download and mediaPlayer logic to controller.. will continue with this in the next days i guess.
What Tobias did in this PR is a proof of concept how communication between ViewHolders and controller could be done..

PR #1362, opened 2021-06-27
Update command-rebase.yml from template

skjnldsv:
See https://github.com/nextcloud/.github


.github/workflows/command-rebase.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout the latest code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      
+      - name: Automatic Rebase
+        uses: cirrus-actions/rebase@1.5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AndyScherzinger, line 27: > NOTE: To ensure GitHub Actions is automatically re-run after a successful rebase action use a [Personal Access Token](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) for `actions/checkout@v2` and `cirrus-actions/rebase@1.4`. See the following [discussion](https://github.community/t/triggering-a-new-workflow-from-another-workflow/16250/37) for more details.

@skjnldsv starting with v1.4 there seems to be a slight change with the tokens. Are we already using personal access tokens?

AndyScherzinger (approved):

AndyScherzinger:
Merged, so let's just see if it works (tomorrow). We can always roll-back :)

.github/workflows/command-rebase.yml
skjnldsv: The change was here already, thi sis about github, not `cirrus-actions/rebase`. Github action does not trigger ci. You can close/open to trigger again. If you want to rebase as an external user, you need to have a PAT. Let's wait for our security expert to give an opinion on this on https://github.com/nextcloud/.github/pull/3#discussion_r658707307

AndyScherzinger: Sounds good, so let's see wait Lukas is thinking 👍

PR #1380, opened 2021-06-30
Shortened launcher label

AndyScherzinger:
shorten app label Nextcloud Talk -> Talk while keeping "Nextcloud Talk" as a label within the app itself

**TODO**

_Brander needs to adapt to additional string in ```setup.xml```_ 
ping @tobiasKaminsky for details on the brander.

❗ **Should only be backported to 12.1 _if_ the brander is going to support it within the next days** ❗

I suggest to postpone this to 12.2 (at least), also cc @nickvergessen since this PR creates a change demand for brander - to be managed/timed

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/AccountVerificationController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/AccountVerificationController.java
@@ -244,8 +244,10 @@ public class AccountVerificationController extends BaseController {
                             fetchProfile(credentials);
                         } else {
                             if (getActivity() != null && getResources() != null) {
-                                getActivity().runOnUiThread(() -> progressText.setText(String.format(getResources().getString(
-                                        R.string.nc_nextcloud_talk_app_not_installed), getResources().getString(R.string.nc_app_name))));
+                                getActivity().runOnUiThread(() -> progressText.setText(
+                                        String.format(
+                                                getResources().getString(R.string.nc_nextcloud_talk_app_not_installed),
+                                                getResources().getString(R.string.nc_app_product_name))));
                             }
 
                             ApplicationWideMessageHolder.getInstance().setMessageType(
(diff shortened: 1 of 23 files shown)


tobiasKaminsky:
> _Brander needs to adapt to additional string in `setup.xml`_
> ping @tobiasKaminsky for details on the brander.
> 
> exclamation **Should only be backported to 12.1 _if_ the brander is going to support it within the next days** exclamation

I can do this today or tomorrow, but only once it is merged and backported :)

AndyScherzinger:
So @nickvergessen and @mahibi what do you think? Also I think the brander can auto generate the new value by combined app_name value and server_product_name value.

nickvergessen:
From my POV 3 strings are better than 2 which are then used alone or combined. Already see customers asking for different order of the strings (RTL) and stuff, so yeah no concatenation is better.

mahibi (approved):

mahibi:
i'm ok with merging to 12.1

tobiasKaminsky:
On brander side we currently only have one name for Talk, which is then used for both (old) nc_app_name and nc_server_product_name.

nc_server_product_name is not really needed in my opinion for brander:
- most of customer have enforced domain, so they can directly authenticate and do not see the server selection page
- attachmentDialog can use it from capabilities
- it is used in error text on server selection controller, which can be reworded, I guess.

AndyScherzinger:
- ✔️ most of customer have enforced domain, so they can directly authenticate and do not see the server selection page
- ⚠️ attachmentDialog can use it from capabilities -> we do and the value is used as a fallback if the capabilities don't return a value ;)
- ✔️ it is used in error text on server selection controller, which can be reworded, I guess. (we can replace it with "server", which makes it even clearer regarding the messages I guess.

So with the brander we have only the issue about the attachment dialog fallback scenario which is wrong even today for branded clients. So I'd keep it and maybe even not so one name for nc_app_name and nc_server_product_name. For example "Sharing with Ubercloud Talk" would be "wrong" since we share it with/via the Nextcloud server, not the server's talk app.
I'd be fine to leave this as is but still add the newly introduced string to give customers the possibility to change the launcher's label.

AndyScherzinger:
The PR has been merged and backported so whatever you decided is fine, if we drop one of the 3 strings available (and to be filled) now that just give me a ping so we don't have to change this again for brander _and_ app.

PR #1387, opened 2021-07-02
Bugfix/1308/improve grid view design

### LINKED ISSUE #1308: improve grid view design
## Steps to reproduce
1. join call with many participants

### Expected behaviour
- nothing overlaps.
- for audio call, the buttons at the bottom don't overlap the participants

### Actual behaviour
- the grid looks quite compressed
- Names and microphone-muted icons are cut off and overlap avatars
- buttons at the bottom don't overlap the participants (at least not in audio call. for video call that might be okay because the buttons disappear after some seconds when there was no tap). this was originally implemented but seems to be broken.

screenshots are at cnc because of privacy reasons (don't know if everyone is ok with avatar beeing shown here..)

### TODO:
- 3 colums are too much for smartphones? limit the shown avatars? make participants scrollable (or better check if it works.. it should already..)

mahibi:
fix #1308 

- for voice calls, the call controls don't overlap the participants anymore (for viceocalls they indeed overlap, but there the controls disappear after a few seconds)

before|after
---|---
before|after

- defined a minHeight for a participant cell so that for too many participants the grid is not compressed but it's scrollable. 
- reduce number of possible columns when many participants are in call to avoid design issues
- cut off very long participant names and append "..."


AndyScherzinger:
/stable-12.1

AndyScherzinger (approved):

PR #1418, opened 2021-07-08
Add missing .editorconfig

timkrueger:
If an developer have a own '.editorconfig' in some location above the repostires
directory it will used in the gradle build and formatting checks can be fail.

To avoid this an get an more uniform code style between Nextclouds Android apps
the '.editorconfig' from the Nextcloud Android app is used.

Because of defining 'root=true' the developers own '.editorconfig' is not used
anymore.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,41 @@
+# see http://EditorConfig.org
+
+# This is the file in the root of the project.
+# For sub folders you can have other files that override only some settings.
+# For these, this settings should be false.
+root=true
+
+[*]
+max_line_length=120
+# use spaces, not tabs.
+indent_style=space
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

nickvergessen (approved):

.editorconfig
```diff
@@ -0,0 +1,41 @@
...
+
+# This is the file in the root of the project.
+# For sub folders you can have other files that override only some settings.
+# For these, this settings should be false.
+root=true
+
+[*]
+max_line_length=120
+# use spaces, not tabs.
+indent_style=space
```
nickvergessen, line 11: my eyes hurt thou

PR #1457, opened 2021-07-12
remove logo dimensions for sever selection screen

mahibi:
(to allow other ratios and sizes for customer logos. adapt to files app which allows all sizes)

### DIFF
--- a/app/src/main/res/layout/controller_server_selection.xml
+++ b/app/src/main/res/layout/controller_server_selection.xml
@@ -41,9 +41,10 @@
         <ImageView
             android:id="@+id/image_logo"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
+            android:layout_height="96dp"
             android:layout_marginBottom="@dimen/standard_margin"
             android:contentDescription="@string/nc_app_product_name"
+            android:scaleType="centerInside"
             app:srcCompat="@drawable/ic_logo" />
 
         <LinearLayout
--- a/app/src/main/res/layout/controller_server_selection.xml
+++ b/app/src/main/res/layout/controller_server_selection.xml
@@ -40,11 +40,10 @@
 
         <ImageView
             android:id="@+id/image_logo"
-            android:layout_width="96dp"
-            android:layout_height="96dp"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@mahibi I don't think this is a good idea. This will scale the view purely based on the images size and it will strech square images to screen width. I'd rather go for:

```
<ImageView
            android:id="@+id/image_logo"
            android:layout_width="match_parent"
            android:layout_height="96dp"
            android:layout_marginBottom="@dimen/standard_margin"
            android:contentDescription="@string/nc_app_product_name"
            android:scaleType="centerInside"
            app:srcCompat="@drawable/ic_logo" />
```

This would perform image scaling but limit the height to 96dp so it won't influence the overall layout due to an image that is very large or in a portrait format.

mahibi:
> @mahibi I don't think this is a good idea. This will scale the view purely based on the images size and it will strech square images to screen width. I'd rather go for:
> 
> ```
> <ImageView
>             android:id="@+id/image_logo"
>             android:layout_width="match_parent"
>             android:layout_height="96dp"
>             android:layout_marginBottom="@dimen/standard_margin"
>             android:contentDescription="@string/nc_app_product_name"
>             android:scaleType="centerInside"
>             app:srcCompat="@drawable/ic_logo" />
> ```
> 
> This would perform image scaling but limit the height to 96dp so it won't influence the overall layout due to an image that is very large or in a portrait format.

square images are not streched to screen width, but anyway i think you solution is better. 
@tobiasKaminsky so also might want to do this for the files app 
https://github.com/nextcloud/android/blob/adbdcd267065a4ec1e7d69ad976c80a1a8fe881c/src/main/res/layout/account_setup.xml#L39-L45

tobiasKaminsky (approved):

AndyScherzinger:
> square images are not streched to screen width, but anyway i think your solution with the fixed height and centerInside is better.

Right, because you also removed android:scaleType="fitXY" 👍 
@tobiasKaminsky I agree with @mahibi that we should adopt this for the files client too, since even our logo (at least for talk) is larger then the original 96dp we used here for the high. This way we lock the vertical size of the logo which keeps the overall layout stable.

AndyScherzinger (approved):

AndyScherzinger:
@mahibi looks good, merge at will 💯

PR #1472, opened 2021-07-14
Remove obsolete setting link preview

timkrueger:
The setting is obsolete since the functionality for media detection in
text messages was removed.

See: commit 8468535a


nickvergessen:
Do we need an update to remove the setting from a database or something? So in case we ever use it again some people don't have it preset

timkrueger:
> Do we need an update to remove the setting from a database or something? So in case we ever use it again some people don't have it preset

Good you say it. I didn't thought about it. By my feeling it should be removed. 

Is this a manual step or is already some :mage: will do it?

//cc @mahibi  @AndyScherzinger

AndyScherzinger:
> Is this a manual step or is already some 🧙 will do it?

No magic, you'd have to manually remove it.

AndyScherzinger:
@timkrueger translations are handled by the transifex bot and will be overridden, so you only need to remove the strings from the master file `values/strings.xml`

timkrueger:
I think I destroyed something during git rebase. I accidentally included commits from the master. 

I'll create a new branch and make a new PR.

AndyScherzinger (approved):
LGTM while it be nice to have a TODO comment in the code to remove it with 13.0.0 (for example)

app/src/main/java/com/nextcloud/talk/utils/preferences/AppPreferences.java
```diff
@@ -253,9 +251,7 @@
     @KeyByString("phone_book_integration") 
     void setPhoneBookIntegration(boolean value);
 
-    @KeyByString("link_previews")
-    void setLinkPreviewsAllowed(boolean value);
-
+    // TODO: Remove in 13.0.0
```
AndyScherzinger, line 254: ❤️

timkrueger:
Backported manually in #1484

PR #1507, opened 2021-08-02
Dav test

tobiasKaminsky:
dav X5 library parses last modified date wrong, found by by @AndyScherzinger 
I debugged it and added two tests to reproduce it.
It only fails in instrumented test, not in unit tests

Bug report upstream: https://gitlab.com/bitfireAT/dav4jvm/-/issues/1


AndyScherzinger:
@tobiasKaminsky the bug report link leads to a 404. :(

AndyScherzinger:
❤️ Trying to loop in @rfc2822 if this mention reaches you 🙏

tobiasKaminsky:
> @tobiasKaminsky the bug report link leads to a 404. :(

Yeah, as this is not public, but only for registered users.
If there is an update, I'll post/inform here.

AndyScherzinger:
> If there is an update, I'll post/inform here.

Thanks

> Yeah, as this is not public, but only for registered users.

Actually I am logged in :/

tobiasKaminsky:
> Actually I am logged in :/

Then it is even more restricted to users in their org.

PR #1589, opened 2021-09-17
Show lobby timer and description in lobby screen 

### LINKED ISSUE #1557: Show lobby timer and description in lobby screen
Server ref https://github.com/nextcloud/spreed/issues/4921

Information should be available already via the room api, so just needs frontend work?

timkrueger:
resolve #1557

I didn't get it the way like in [iOS](https://github.com/nextcloud/talk-ios/pull/615#issue-987829178). So stay with the timer like in spreed. Is that OK?

Days | Hours | Minutes | Soon
------- | -------- | ---------- | --------
days | hours | minutes | soon

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -1127,7 +1127,8 @@ class ChatController(args: Bundle) :
                         DateUtils.getLocalDateStringFromTimestampForLobby(
                             currentConversation?.lobbyTimer
                                 ?: 0
-                        )
+                        ),
+                        currentConversation!!.description
                     )
                 } else {
                     binding.lobby.lobbyTextView.setText(R.string.nc_lobby_waiting)
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -1119,20 +1119,26 @@ class ChatController(args: Bundle) :
                 binding.messageInputView.visibility = View.GONE
                 binding.progressBar.visibility = View.GONE
 
+                val sb = StringBuilder()
+                sb.append(resources!!.getText(R.string.nc_lobby_waiting))
+                    .append("\n\n")
+
                 if (currentConversation?.lobbyTimer != null && currentConversation?.lobbyTimer !=
                     0L
                 ) {
-                    binding.lobby.lobbyTextView.text = String.format(
-                        resources!!.getString(R.string.nc_lobby_waiting_with_date),
-                        DateUtils.getLocalDateStringFromTimestampForLobby(
-                            currentConversation?.lobbyTimer
-                                ?: 0
-                        ),
-                        currentConversation!!.description
+                    val timestamp = currentConversation?.lobbyTimer ?: 0
+                    val stringWithStartDate = String.format(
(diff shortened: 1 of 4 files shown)


mahibi:
i'm a bit confused by the german text. something is wrong here with translations or concatenation..?

grafik

timkrueger:
> i'm a bit confused by the german text. something is wrong here with translations or concatenation..?
> 
> grafik

I'll take a look

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -1119,19 +1119,20 @@ class ChatController(args: Bundle) :
...
                         resources!!.getString(R.string.nc_lobby_waiting_with_date),
-                        DateUtils.getLocalDateStringFromTimestampForLobby(
-                            currentConversation?.lobbyTimer
-                                ?: 0
-                        )
+                        DateUtils.relativeStartTimeForLobby(timestamp, resources!!)
                     )
                 } else {
-                    binding.lobby.lobbyTextView.setText(R.string.nc_lobby_waiting)
+                    resources!!.getText(R.string.nc_lobby_waiting)
```
mahibi, line 1131: resources?.getText

app/src/main/java/com/nextcloud/talk/utils/DateUtils.kt
```diff
@@ -41,6 +54,48 @@ object DateUtils {
...
+                }
+                minutes > 0 -> {
+                    fmt.format(
+                        minutes.toDouble(),
+                        Direction.NEXT,
+                        RelativeUnit.MINUTES
+                    )
+                }
+                else -> {
+                    resources.getString(R.string.nc_lobby_start_soon)
```
nickvergessen, line 92: the soon case should not be done the same way the relative date works. As Marcel pointed out it breaks with translations

timkrueger: > the soon case should not be done the same way the relative date works. As Marcel pointed out it breaks with translations

You're right, of course. I'll change it.

app/src/main/res/values/strings.xml
```diff
@@ -350,7 +350,9 @@
     <string name="nc_lobby">Lobby</string>
     <string name="nc_start_time">Start time</string>
     <string name="nc_lobby_waiting">You are currently waiting in the lobby.</string>
-    <string name="nc_lobby_waiting_with_date">You are currently waiting in the lobby.\n This meeting is scheduled for %1$s.</string>
+    <string name="nc_lobby_waiting_with_date">You are currently waiting in the lobby.\n\nThis meeting is scheduled for
```
mahibi, line 353: i think we should avoid linebreaks in translations. better would be to have 2 seperate strings?

app/src/main/res/values/strings.xml
```diff
@@ -350,7 +350,9 @@
...
+        %1$s.</string>
+    <string name="nc_lobby_start_soon">The meeting will start soon</string>
```
mahibi, line 355: not used anymore?

app/src/main/java/com/nextcloud/talk/utils/DateUtils.kt
timkrueger: I left the soon case in the else part of the relative dates handling but append the complete text in the lobby screen in an other way.

mahibi (approved):

PR #1594, opened 2021-09-22
differentiate unread counters

### LINKED ISSUE #1559: Differenciate unread counters for group mentions and direct mentions
- [x] Server ref https://github.com/nextcloud/spreed/issues/4623
- [x] Needs API changes first to have either 2 unreadMentions returned or the level

mahibi:
resolve #1559 

differentiate unread counters for group mentions and direct mentions

before | after (server with new capability) | after (server without new capability)
------- | -------- | ---------- 
grafik | grafik | grafik | 

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
@@ -42,6 +42,7 @@ import com.facebook.drawee.view.SimpleDraweeView;
 import com.google.android.material.chip.Chip;
 import com.nextcloud.talk.R;
 import com.nextcloud.talk.application.NextcloudTalkApplication;
+import com.nextcloud.talk.models.database.CapabilitiesUtil;
 import com.nextcloud.talk.models.database.UserEntity;
 import com.nextcloud.talk.models.json.chat.ChatMessage;
 import com.nextcloud.talk.models.json.conversations.Conversation;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
```diff
@@ -136,20 +137,42 @@ public void bindViewHolder(FlexibleAdapter<IFlexible> adapter, ConversationItemV
...
                 holder.dialogUnreadBubble.setTextColor(Color.WHITE);
+            } else if (conversation.isUnreadMention()) {
+                if (CapabilitiesUtil.hasSpreedFeatureCapability(userEntity, "direct-mention-flag")){
+                    if (conversation.getUnreadMentionDirect()) {
+                        holder.dialogUnreadBubble.setChipBackgroundColorResource(R.color.colorPrimary);
+                        holder.dialogUnreadBubble.setTextColor(Color.WHITE);
+                    } else {
+                        holder.dialogUnreadBubble.setChipBackgroundColor(lightBubbleFillColor);
+                        holder.dialogUnreadBubble.setTextColor(lightBubbleTextColor);
+                        holder.dialogUnreadBubble.setChipStrokeWidth(4.0f);
```
nickvergessen, line 161: From the screenshots I think the border could be thicker.

nickvergessen (approved):
Looks good technical, only the border size is to discuss I think.

app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
nickvergessen: @marcoambrosini can you have a look at the screenshots and what border size is the web UI going to use?

mahibi: stroke width 4.0
grafik

stroke width 5.0
grafik

stroke width 6.0
grafik

mahibi: updated screenshots. now with stroke width 6.0

marcoambrosini (changes requested):
@marcelklehr could you please match the style of the new counter to the one we have in the vue library? 
Screenshot 2021-09-07 at 10 05 04

timkrueger (approved):

mahibi:
> @marcelklehr could you please match the style of the new counter to the one we have in the vue library?
> Screenshot 2021-09-07 at 10 05 04

changed background for group mentions to white and updated screenshots @marcoambrosini

jancborchardt:
> changed background for group mentions to white and updated screenshots @marcoambrosini

2 more details there, as shown in the screenshot @marcoambrosini posted:
- When the border is blue (team mention only), the number should also be blue
- The text needs more padding, that is the bubble size should increase a bit – ideally make it look exactly like on server :)

mahibi:
>     * When the border is blue (team mention only), the number should also be blue
done

>     * The text needs more padding, that is the bubble size should increase a bit – ideally make it look exactly like on server :)

i played around with that... it's possible to change the padding, but depending on which number is inside the bubble this looks quite bad. For example a "31" looks quite okay, but a 12 feels like there is too much padding at the beginning. i saw that @AndyScherzinger already set the current values startPadding=-3dp and endPadding=-1dp and i can confirm these values are the best compromise.

as an example see the following screenshots with padding "0":
grafik

jancborchardt:
> @AndyScherzinger already set the current values startPadding=-3dp and endPadding=-1dp and i can confirm these values are the best compromise.

Reducing the left padding sounds good, but the right padding seems ok at 0. And there should be a bit more padding on top and bottom. :)

nickvergessen:
The top/bottom padding is clashing with the timestamp (which should be shown like in any other messenger) which is not there on the web.

mahibi:
discussed with @jancborchardt in person:
- left & right padding values are best like they were before.
- top & bottom padding won't change (and can't be changed that easily)
- date won't be bold anymore

Ivansss:
> changed background for group mentions to white and updated screenshots

Should the background be white or same color as the cell background?

IMG_E0093

@marcoambrosini @jancborchardt

nickvergessen:
Background color, otherwise the fallback of primary element will fail

PR #1607, opened 2021-09-29
Clear message field after sharing

timkrueger:
After the Android sharing functionality was used, the shared text sticks
in the message composing field in the conversation. Also if an text was
forwared to the same conversation the text which was shared before
sticks in the message composing field.

Now the message composing field is empty after sharing or forwarding
something.


nickvergessen (approved):
🐘

PR #1619, opened 2021-10-07
Feature/1547/setup test automation on ci

timkrueger:


.drone.yml
```diff
@@ -16,66 +17,114 @@ trigger:
...
+kind: pipeline
+type: docker
+name: tests
+
+steps:
+  - name: gplay
+    image: nextcloudci/android:android-49
+    commands:
+      # TODO: Add '&' to start emulator in background
+      - emulator-headless -avd android-27 -no-snapshot -gpu swiftshader_indirect -no-window -no-audio -skin 500x833
```
timkrueger, line 46: Currently this command produces an error with the following output:

```
drone exec --pipeline=tests
[gplay:0] + emulator-headless -avd android-27 -no-snapshot -gpu swiftshader_indirect -no-window -no-audio -skin 500x833
[gplay:1] statvfs('/root/.android/avd/android-27.avd/snapshots/default_boot/ram.img') failed: No such file or directory
[gplay:2] emulator: ERROR: x86 emulation currently requires hardware acceleration!
[gplay:3] Please ensure KVM is properly installed and usable.
[gplay:4] CPU acceleration status: /dev/kvm is not found: VT disabled in BIOS or KVM kernel module not loaded
[gplay:5] More info on configuring VM acceleration on Linux:
[gplay:6] https://developer.android.com/studio/run/emulator-acceleration#vm-linux
[gplay:7] General information on acceleration: https://developer.android.com/studio/run/emulator-acceleration.
[gplay:8] 
2021/12/07 14:09:37 gplay : exit code 1
```

I expected that this should work because I copied the command and the image from the Files app:
https://github.com/nextcloud/android/blob/stable-3.13.0/.drone.yml#L105

tobiasKaminsky: I added priviliged: true
This should make it work.

nickvergessen (approved):
Seems to work

app/src/androidTest/java/com/nextcloud/talk/activities/MainActivityTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            Thread.sleep(2000)
+        } catch (e: InterruptedException) {
+            e.printStackTrace()
+        }
+
+        sut.runOnUiThread { sut.resetConversationsList() }
+
+        try {
+            Thread.sleep(20000)
```
tobiasKaminsky, line 52: This is not needed.

app/src/androidTest/java/com/nextcloud/talk/activities/MainActivityTest.kt
```diff
@@ -0,0 +1,57 @@
...
+                { Log.d("test", "complete") }
+            )
+
+        try {
+            Thread.sleep(2000)
+        } catch (e: InterruptedException) {
+            e.printStackTrace()
+        }
+
+        sut.runOnUiThread { sut.resetConversationsList() }
```
tobiasKaminsky, line 49: Here a visual check (e.g. via Karumi) would be good.
Or check that user was created.

Otherwise this just tests that there is no crash, without guaranteeing that the app is doing what you want

PR #1622, opened 2021-10-11
Use build flavor for implementation declaration

timkrueger:
Instead of loading dependencies with an own snippet of code now the
provided functionality of configure flavor based implementation
declaration is used.

See: https://developer.android.com/studio/build/build-variants#dependencies


timkrueger:
@tobiasKaminsky Can you please mark your [change request](#pullrequestreview-776193561) as resolved?

PR #1634, opened 2021-10-14
timeout call notification

mahibi:
this will hangup:
- the call notification after 1 minute (the popup at the top)
- the call notification screen after 1 minute (the screen with the hangup and accept call buttons)

### DIFF
--- a/app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
+++ b/app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
@@ -64,6 +64,7 @@ import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_FROM_NOTIFICATION_START_CA
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_USER_ENTITY
 import com.nextcloud.talk.utils.preferences.AppPreferences
 import com.nextcloud.talk.utils.singletons.ApplicationWideCurrentRoomHolder
+import io.reactivex.Observable
 import io.reactivex.Observer
 import io.reactivex.disposables.Disposable
 import io.reactivex.schedulers.Schedulers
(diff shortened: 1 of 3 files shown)


AndyScherzinger (approved):
LGTM

app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
```diff
@@ -337,29 +341,23 @@ class MagicFirebaseMessagingService : FirebaseMessagingService() {
                     hasParticipantsInCall = participantList.isNotEmpty()
                     if (hasParticipantsInCall) {
                         for (participant in participantList) {
-                            if (participant.userId == signatureVerification.userEntity.userId) {
+                            if (participant.actorId == signatureVerification.userEntity.userId) {
```
nickvergessen, line 344: also have to compare the actorType?

mahibi: added check for actorType

mahibi:
> you could also get the single notification and check the text there. then you automatically have the updated text available when it changed

@nickvergessen i don't get it

mahibi:
merged this PR to fix at least the invinite ringing. 
follow up issue is https://github.com/nextcloud/talk-android/issues/1724

AndyScherzinger:
@mahibi backport to 13.0 failed btw.

mahibi:
> @mahibi backport to 13.0 failed btw.
@AndyScherzinger 

yep will do.
but will backport it to 13.0.0 ...?
asking because you set it to 13.1.0..

AndyScherzinger:
@mahibi yes, because this PR has targeted master which is already 13.1, so the backport PR will then be 13.0 😊

timkrueger:
> @mahibi yes, because this PR has targeted master which is already 13.1, so the backport PR will then be 13.0 blush

But when we backport this PR to 13.0 then the milestone must also 13.0. Or what I misunderstand?

AndyScherzinger:
> But when we backport this PR to 13.0 then the milestone must also 13.0. Or what I misunderstand?

Yes but only on the newly created one by the backport bot which will set it correctly by itself :)

PR #1639, opened 2021-10-20
Handle incoming intent with server and userId as payload

tobiasKaminsky:
TODO
- [ ] switch account if account is available and not currently used

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -133,6 +133,12 @@
                 <data android:scheme="content" />
                 <data android:scheme="file" />
             </intent-filter>
+
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="text/text" />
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
It is about:
```
THROWS_METHOD_THROWS_RUNTIMEEXCEPTION: Method intentionally throws RuntimeException.

Method intentionally throws RuntimeException.
According to the [SEI CERT ERR07-J rule](https://wiki.sei.cmu.edu/confluence/display/java/ERR07-J.+Do+not+throw+RuntimeException%2C+Exception%2C+or+Throwable), throwing a RuntimeException may cause errors, like the caller not being able to examine the exception and therefore cannot properly recover from it.
Moreover, throwing a RuntimeException would force the caller to catch RuntimeException and therefore violate the [SEI CERT ERR08-J rule](https://wiki.sei.cmu.edu/confluence/display/java/ERR08-J.+Do+not+catch+NullPointerException+or+any+of+its+ancestors).
Please note that you can derive from Exception or RuntimeException and may throw a new instance of that exception.
```

Issue #1677, opened 2021-11-11
Videocalls stop working when turning off PIP Mode


andyxheli:
## Steps to reproduce
1. Turn off PIP
2. Join Call

### Expected behaviour
Once you join the call rings once and hangs up. 

### Actual behaviour
Should join the call

## Device information

**Device:** Fairphone2/…
Oneplus N200 5G
**Android version:** 10/…
Oxygen OS 11.0.3.0.DE17AA
**Talk version:** 11.0.0
QA 
## Server information

**Nextcloud version:** (see admin overview page: `/index.php/settings/admin/overview`)
22.2.1
**Talk version:** (see apps admin page: `/index.php/settings/apps`)
12.1.2
**Custom Signaling server configured:** yes/no and version (see talk admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`)
No
**Custom TURN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#turn_server`)
Yes
**Custom STUN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#stun_server`)
Yes

https://user-images.githubusercontent.com/59488153/141344361-f4f693d7-5147-43d8-a82e-b540fb91f461.mp4

mahibi:
thanks for the report. seems to happen for videocalls only. i have a look

mahibi:
well there is logic that enterPipMode is called when onUserLeaveHint is executed:
https://github.com/nextcloud/talk-android/blob/d602bf3dfe601abed3f98bd1d574e0776a2aecc4/app/src/main/java/com/nextcloud/talk/activities/CallBaseActivity.java#L91

When in enterPipMode it's recognized that pip is not possible, the activity is finished (yes, that not cool but will be improved in the future. most probably with a background service&notification).

The main question is why onUserLeaveHint is executed when a videocall starts?!

![grafik](https://user-images.githubusercontent.com/2932790/141465223-6a163ccd-4aeb-47b3-9860-e848c602e338.png)

So enterPipMode is already called on start and it kills the activity...

Anyone has an idea why onUserLeaveHint could be called when a videocall starts??

mahibi:
log until activity is finished:

```
D/ChatController: getRoomInfo
I/okhttp.OkHttpClient: --> GET http://192.168.8.106/nextcloud8nov/ocs/v2.php/apps/spreed/api/v4/room/wup2odnv
I/okhttp.OkHttpClient: Authorization: ██
[...]

PR #1685, opened 2021-11-15
add workaround to start videocalls when PIP is disabled

mahibi:
this is a dirty workaround for issue #1677

Somehow onUserLeaveHint is executed when the user starts a videocall. If PIP is disabled, the logic inside enterPipMode would finish the activity right after it was started.
This workaround suppresses the execution of enterPipMode right after the activity was started.
However if a user would press the home button in the first three seconds, the call would continue in background without the ability to recover the UI.

To better fix this bug it must be found out why onUserLeaveHint is executed on start (this should not happen!).


timkrueger (approved):
Approval on the condition that work on it continues directly after the release of the 13 beta.

PR #1692, opened 2021-11-15
add hint that currently java8 must be used

mahibi:
any idea what further information is needed to compile the project ?


AndyScherzinger (approved):

tobiasKaminsky:
For me it works with latest Android Studio and
image


Here I must explicitly set an absolute path to JDK8.

timkrueger (changes requested):
@tobiasKaminsky is right. See https://github.com/nextcloud/talk-android/pull/1692#issuecomment-969932871

I tested it with _Android Studio Arctic Fox | 2020.3.1 Patch 3_ and it works fine. Maybe some changes in the past repaired something.

So the only precondition is Java 8.

AndyScherzinger (commented):
Like @tobiasKaminsky stated and @timkrueger tested you actually need `Java8` which then also locks you on gradle `4.1.3` since newer version I think will also enforce Java11 (not sure about the later)

tobiasKaminsky:
> Like @tobiasKaminsky stated and @timkrueger tested you actually need `Java8` which then also locks you on gradle `4.1.3` since newer version I think will also enforce Java11 (not sure about the later)

True, but once JDK11 can be used Gradle is also no problem, but this requires parceler and other libs to be removed/ replaced.

AndyScherzinger:
> True, but once JDK11 can be used Gradle is also no problem, but this requires parceler and other libs to be removed/ replaced.

Yes, true

AndyScherzinger:
👍 LGTM

tobiasKaminsky:
Please also include instruction how to do it:
Install JDK 8 and Screenshot may be enough

timkrueger (approved):

PR #1703, opened 2021-11-17
Skip QA build if keystore not available

timkrueger:
Copied solution from the Nextcloud Files for Android app from commit
6863a8 [1].

From Álvaro's commit message:

> There is no way to skip the whole workflow depending on whether it is
> a PR from a fork or not. The only workaround is to check whether secrets are present or not.
>
> So, I've added a step to check if KS_PASS is empty, and set an output depending on that.
> If it is empty, we just skip the rest of the steps.

[1] https://github.com/nextcloud/android/commit/6863a8a637d4a02dd7a2a908e75d3b7b6f7901e2

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -8,12 +8,18 @@ jobs:
     qa:
         runs-on: ubuntu-latest
         steps:
+            -   name: Check if secrets are available
+                run: echo "::set-output name=ok::${{ secrets.KS_PASS != '' }}"
+                id: check-secrets
             -   uses: actions/checkout@v2
+                if: ${{ steps.check-secrets.outputs.ok == 'true' }}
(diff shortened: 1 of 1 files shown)


AlvaroBrey (approved):

AndyScherzinger (approved):

timkrueger:
> /backport to stable-13.0

In that case the milestone must also changed to 13.0?!

AndyScherzinger:
> In that case the milestone must also changed to 13.0?!

Nope, since this is a PR, not an issue. So

* Issue: assign to the oldest milestone that ships the fix for the issue
* PR: assign the milestone the respective PR is targeting

So in this case, `stable-13.0`=>`13.0` and `master`=>`13.1` This way, even in some years you can just look into a milestone to see all changes made within it.

PR #1723, opened 2021-11-22
fix to keep session

mahibi:
WIP, will be improved.

when you were in a chat and received a call notification, then in onDetach of the chatview the session was removed by leaving the room. This fix will check if a call notification is ongoing and suppress to leave the room.

leaving the room in "onDetach" might cause more problems..(for example on first call when granting permissions?!)

TODO: 
- [ ] check which issues might be related to the onDetach& leaveRoom handling


mahibi:
i think i was on the wrong track here. The fault will lie elsewhere.
It might make sense not to recreate the session, but it should still work.

Issue #892, opened 2020-09-02
Mute notifications in Android, still get a sound


tobiasKaminsky:
- Mute notifications in Android (long press on Talk notification and select silent)
- still receive a sound
- one needs to disable sound additionally in settings of Talk

0ranki:
I'd like to add a little:

Android 10, (OxygenOS 10.5.11 OnePlus Nord N10)
Talk installed from Play Store, v11.1.0

- If you change notification sound only from the app's own settings, the newly selected notification sound and the default notification sound will play for each notification
- If you change notification sound from both the app's settings and Android's notification settings, you'll get two sounds per notification, I think first the Android one and then the one selected from the app
- If you mute the notification from the app's own settings, you get no notifications at all
- If you mute the sound from Android notif.settings, (but not from the app) the sound plays only once per notification, like expected.

mahibi:
fixed with https://github.com/nextcloud/talk-android/pull/1728

PR #1728, opened 2021-11-27
Adapt notification code to Android 8.0 (Oreo) and later

### LINKED ISSUE #756: Double push notification sound
At the first push notification, the sound is played twice. The sound is played correctly the next notifications.

Steps:
1. Receipt of one push notification **(sound played twice)**.
2. Receipt of the second notification (playback sound correctly).

App:
Android 10
Talk Android 8.0.3
Talk Nextcloud 8.0.5
Nextcloud 18.0.1

starypatyk:
Trying to fix #756 and #892

This seems to be caused by the fact that since Android 8 (Oreo) notification sounds are handled by the system - as described in this article: [Create and Manage Notification Channels](https://developer.android.com/training/notify-user/channels).

Changes to implement - most should apply only to devices running Android 8 (Oreo) or later:

- [x] [Do not play notification sounds (system will do that)](https://github.com/nextcloud/talk-android/pull/1728/commits/65f0a777a6acd9caa7afcb6b4d1bf4f1c009b420).
- [x] [Provide UI to change notification settings from application settings (app preferences do not affect system settings)](https://github.com/nextcloud/talk-android/pull/1728/commits/622a266b6fc7ddefb5f4fc91ef186c65efa09b04).
Addtional notes:
* I have refactored sound selection code to cover differences between old (pre Oreo) and new devices.
* I have changed the notification channel handling code to always use the same notification channel for calls. Previous implementation used a hash of the notification parameters as a channel ID. This prevented users from configuring call notification settings in the system UI.
- [x] [Register notification channels on application start-up](https://github.com/nextcloud/talk-android/pull/1728/commits/541ffecccd2006fb14433763f265bf8c9b70f2c1).
Plus some refactoring - move more notification related code into NotificationUtils.
- [x] [Implement changes to the UI related to notifications and simplify channel name - as described below](https://github.com/nextcloud/talk-android/pull/1728/commits/24cf3caf455bc88fa101fe37b8d93e12a2e95b58).
- [ ] _(Optional)_ Check using **NotificationManagerCompat.areNotificationsEnabled** to signal the user that notifications will not be available.

I would like to suggest some changes to the UI related to notifications.

Currently the notification/sounds related section in the app settings looks like this:

On old devices (pre Oreo) I would suggest to change it to:

On new devices (Oreo and later) these UI elements are used to open system notification settings, so I would suggest:

I would also suggest to simplify names of the notification channels. Currently the channels look like this in the system settings:

I suggest changing it to:

### DIFF
--- a/app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
+++ b/app/src/gplay/java/com/nextcloud/talk/services/firebase/MagicFirebaseMessagingService.kt
@@ -27,7 +27,6 @@ import android.media.AudioAttributes
 import android.net.Uri
 import android.os.Bundle
 import android.os.Handler
-import android.text.TextUtils
 import android.util.Base64
 import android.util.Log
 import androidx.core.app.NotificationCompat
@@ -48,7 +47,6 @@ import com.nextcloud.talk.application.NextcloudTalkApplication.Companion.sharedA
 import com.nextcloud.talk.events.CallNotificationClick
 import com.nextcloud.talk.jobs.NotificationWorker
 import com.nextcloud.talk.jobs.PushRegistrationWorker
-import com.nextcloud.talk.models.RingtoneSettings
 import com.nextcloud.talk.models.SignatureVerification
 import com.nextcloud.talk.models.json.participants.Participant
 import com.nextcloud.talk.models.json.participants.ParticipantsOverall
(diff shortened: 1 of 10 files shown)


Valdnet:
It's great that it will get fixed eventually 👍.

@starypatyk I hope you will finish the work successfully 😉.

mahibi:
looks really promising @starypatyk :+1: 
will have a closer look tomorrow!

mahibi:
tested and reviewed. the bugs are fixed.
Well done @starypatyk !! Thank you very much :1st_place_medal:

Issue #1730, opened 2021-11-30
 Guests without a name don't show "Guest" when quotes


timkrueger:
## Steps to reproduce
1. Reply to a guests message

### Expected behaviour
About quoted message the label _Guest_ should be shown.

### Actual behaviour
No label is shown.

PR #1733, opened 2021-12-03
Show quotes of unnamed guests with "Guests"

timkrueger:
Before this change the default "Guests" label only was set if the actor
display name was 'null'.

Resolves: #1730


mahibi (approved):

AndyScherzinger:
you only merged it to master, so it would go into 13.1, no?

timkrueger:
> you only merged it to master, so it would go into 13.1, no?

Sorry @AndyScherzinger and thanks for correction.

AndyScherzinger:
No worries @timkrueger - I asked just to be sure :)

PR #1749, opened 2021-12-22
Image keyboard support in chat

### LINKED ISSUE #411: Add GIF (Giphy) keyboard

starypatyk:
Should resolve #411

Added support for images received from keyboard as described here:
https://developer.android.com/guide/topics/text/image-keyboard

When a keyboard (like Google Keyboard) sends an image - usually a GIF 😜 - to the text entry field, the code uploads the image to the current chat.

Implementation is a bit tricky, since it "injects" a new class _ImageEmojiEditText_ into a part of the layout managed by external **ChatKit** component. However since the _ImageEmojiEditText_ inherits from _EmojiEditText_ used originally, **ChatKit** seems to work just fine. And we can attach the required callback in the _ChatController_ class.


mahibi (approved):

mahibi:
works fine, thank you very much @starypatyk :+1:

PR #1751, opened 2021-12-27
Additional fixes related to notification channels

starypatyk:
Additions to PR #1728

**1. Remove unused notification channels during upgrade**

I have not correctly considered app upgrade. There were the following unintended side-effects when upgrading:

- the old "Calls notification channel" was still visible in the system UI but not used,
- the channel "Message notification channel" was not renamed to "Messages" as expected (registering the channel again with the same ID but a new name does not change the name of the channel).

The new code:

- removes all unused notification channels,
- uses a new ID (NOTIFICATION_CHANNEL_MESSAGES_V4) for the "Messages" channel.

**2. App would crash on pre-Oreo devices with "No sound" setting**

Null ringtone Uri must be allowed.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/RingtoneSelectionController.java
+++ b/app/src/main/java/com/nextcloud/talk/controllers/RingtoneSelectionController.java
@@ -181,9 +181,9 @@ public class RingtoneSelectionController extends BaseController implements Flexi
 
     private String getRingtoneString() {
         if (callNotificationSounds) {
-            return NotificationUtils.INSTANCE.getDEFAULT_CALL_RINGTONE_URI();
+            return NotificationUtils.DEFAULT_CALL_RINGTONE_URI;
         } else {
-            return NotificationUtils.INSTANCE.getDEFAULT_MESSAGE_RINGTONE_URI();
+            return NotificationUtils.DEFAULT_MESSAGE_RINGTONE_URI;
         }
     }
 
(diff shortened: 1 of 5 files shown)


mahibi:
looks good to me, thank you very much @starypatyk :+1: 
only 2 questions:

- while testing i didn't manage to receive an intent for   android.intent.action.MY_PACKAGE_REPLACED 
i manually installed an old version of the app and afterwards installed the new version by starting debug mode in AS. Shouldn't this trigger the intent? or would it only work when installing from playstore?

- if i use a device >= Oreo and change the ringtone, how to reset the ringtone to the default one provided by the app? Right now if i select "back to standard", the ringtone is set to "none".

starypatyk:
Hi @mahibi 

> * while testing i didn't manage to receive an intent for   android.intent.action.MY_PACKAGE_REPLACED
>       i manually installed an old version of the app and afterwards installed the new version by starting debug mode in AS. Shouldn't this trigger the intent? or would it only work when installing from playstore?

To test this I have used **adb** with **-r** command line switch to install the new version:
`adb -e install -r app-gplay-debug.apk`

I have found this suggestion here:
https://newbedev.com/android-broadcast-action-my-package-replaced-never-received

I assume that when the new version is installed from Google Play store this will also trigger the intent.

> * if i use a device >= Oreo and change the ringtone, how to reset the ringtone to the default one provided by the app? Right now if i select "back to standard", the ringtone is set to "none".

Unfortunately I have not found a way to reset the ring tone or the message tone back to the default one. :confused:

Different phone brands use different UI, on the ones I have used (stock Android on an emulator and a Xiaomi device), I have not even seen the "back to standard" item. Apparently other applications have the same issue - I have checked Hangouts and Skype.

mahibi:
thank you @starypatyk :1st_place_medal:

PR #1752, opened 2021-12-27
mark message as unread / conversation as read

### LINKED ISSUE #1464: mark message as unread / conversation as read
This was introduced in Talk 12

AndyScherzinger:
Resolves #1464


app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -2312,6 +2312,38 @@ class ChatController(args: Bundle) :
                         clipboardManager.setPrimaryClip(clipData)
                         true
                     }
+                    R.id.action_mark_as_unread -> {
+                        val chatMessage = message as ChatMessage?
+                        ncApi!!.setChatReadMarker(
+                            credentials,
+                            ApiUtils.getUrlForSetChatReadMarker(
+                                1,
```
AndyScherzinger, line 2320: @nickvergessen I hard-coded v1 API since calling it for v4 gives my a `404` return code.

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java
```diff
@@ -480,6 +482,17 @@ public void onComplete() {
                         });
 
                 break;
+            case 96:
+                // TODO: why does it break with v4?
```
AndyScherzinger, line 486: @nickvergessen same here, same API, so I is then just a marker for me :)

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
nickvergessen: Chat endpoints are at API level 1 at the moment, so use:
https://github.com/nextcloud/talk-android/blob/3bc4c0c9838489277bf6c1fec6abae50a60f21a8/app/src/main/java/com/nextcloud/talk/utils/ApiUtils.java#L173 with supported versions "1":
https://github.com/nextcloud/talk-android/blob/bb422fb11213192caabe290bb8065723f1bc631a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt#L1802

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java
nickvergessen: Same above

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
AndyScherzinger: Thanks :)

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -2312,6 +2312,38 @@ class ChatController(args: Bundle) :
...
+                                conversationUser?.baseUrl,
+                                roomToken
+                            ),
+                            chatMessage!!.jsonMessageId.minus(1)
```
AndyScherzinger, line 2324: @nickvergessen sorry forgot to ask, I don't 'know' the message id of the message above the one where you click 'mark as unread' but simply do an 'id-1'.. Is that okay? Seems to work :D

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java
```diff
@@ -480,6 +482,17 @@ public void onComplete() {
...
+                ncApi.setChatReadMarker(credentials,
+                                        ApiUtils.getUrlForSetChatReadMarker(1,
+                                                                            currentUser.getBaseUrl(),
+                                                                            conversation.getToken()),
+                                        conversation.lastMessage.jsonMessageId)
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .retry(1)
+                    .subscribe(genericOperationsObserver);
```
timkrueger, line 495: Here a `break` is missing. Currently a click on _Mark as read_ also marks the conversation as favorite ⭐

timkrueger (dismissed):
Beside that favorite thing it works very well. Thank you @AndyScherzinger

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/1752/commits/0ba210a0cae90883395eb8aeec190a063831a567

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java
AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/1752/commits/0ba210a0cae90883395eb8aeec190a063831a567

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java
AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/1752/commits/0ba210a0cae90883395eb8aeec190a063831a567

AndyScherzinger:
@timkrueger just fixed the falling through the switch/case which was also detected by spotbugs :)

timkrueger (approved):

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
nickvergessen: Will break when it's in another conversation. What the webUI does it when creating the message list add the previous message id to this message element as "previousId".
But I also start seeing the API to be not very good here. It comes from the fact that we just have one "mark read here" endpoint. Maybe a new one should be added which then takes care of all this.

AndyScherzinger: I'll have a look, but thatz one might be tricky, let's see

AndyScherzinger:
Just started working on the handling of the message ids while there are still issues with newly created and updated messages in the chat. So still trying to find a way to implement the unread marker part.

AndyScherzinger:
@nickvergessen @timkrueger feel free to re-test and re-review. I implemented the previous-id calculation for unread-marking based on the feedback from https://github.com/nextcloud/talk-android/pull/1752#discussion_r775948309 - I also don't think it be much easier to do on the server side, so I think doing it on the client side is just fine. 👍

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
AndyScherzinger: So be fixed now :)

nickvergessen:
> I also don't think it be much easier to do on the server side, so I think doing it on the client side is just fine. 

Problem is we currently can't do it from the conversation list in the web, so will have to go for a server endpoint at some time

AndyScherzinger:
> Problem is we currently can't do it from the conversation list in the web, so will have to go for a server endpoint at some time

You mean mark as unread? Yeah, that use case would need another endpoint, true.

mahibi:
first of all it's very confusing when there is an "mark as unread" option, but no visible indicator at all which incoming messages are marked as unread. So inside the chat there's just no feedback if something happened. In the Web there is this "Unread messages"-line. So we should have this for android talk too if possible? Is it available in iOS talk @Ivansss ?


and after some testing, the "mark as unread" feature seems to be quite buggy:

when marking a message as unread in android...

- sometimes it works
- sometimes it works not at all
- sometimes the following messages are marked as unread only on android or only on web (also when waiting a long time)
- sometimes the unread message has some offset, so the message that was clicked is still marked as read but one message later ("newer") the messages are now unread. I guess this only happens when playing around a lot..

mahibi (changes requested):
see comment above

Ivansss:
> So inside the chat there's just no feedback if something happened. In the Web there is this "Unread messages"-line. So we should have this for android talk too if possible? Is it available in iOS talk @Ivansss ?

There is an "Unread messages" indicator in the Talk iOS app, but the "Mark as unread" feature is not implemented yet.

nickvergessen:
> first of all it's very confusing when there is an "mark as unread" option, but no visible indicator at all which incoming messages are marked as unread. So inside the chat there's just no feedback if something happened. In the Web there is this "Unread messages"-line. So we should have this for android talk too if possible? Is it available in iOS talk @Ivansss ?

There is a line in iOS

> and after some testing, the "mark as unread" feature seems to be quite buggy:

Maybe you are staying too long on the chat and it marks read with the next pulling of the chat again?

AndyScherzinger:
Yeah, the mark as unread is a big hack since we currently need to calculate the previous message id into each element, kinda like a singly linked list which has its issues in case of messages coming in, own, written messages and skipping the meta data items in the list view created by the chat UI library...

So maybe then split this PR so the mark as read could be merged? While I am afk atm and could only start working on this again next week.

mahibi:
> So maybe then split this PR so the mark as read could be merged? While I am afk atm and could only start working on this again next week.

yep makes sense, will merge this PR

mahibi (approved):

AndyScherzinger:
@mahibi erm, now you merged it as is which means is ships the Buggy mark as unread which I the only thing wanted to remove on this PR prior merging ;)

So next week I then need to create a simple PR disabling the mark as unread part at least.

mahibi:
Sorry @AndyScherzinger , I misunderstood your comment in a hurry, this merge was indeed very rushed 🙈
let me know if you want us to take over

AndyScherzinger:
Hi @mahibi I might have the time next week to disable the mark as unread part. If you want the mark as unread part rather being fixed, please take over since I won't have the time to debug and fix that part in the near future :/

mahibi:
@AndyScherzinger i fear i won't find the time in the next days so maybe it's really better to disable for now :/

PR #1776, opened 2022-01-18
Feature/1740/recover from deleted push token

### LINKED ISSUE #1740: Recover from deleted push notification hash on NC server or push token on FCM
On each start of the app and roughly once every 24h the app should try to "heal" the push registration to make sure it recovers when FCM lost the token or the Nextcloud server did.

Similar to https://github.com/nextcloud/talk-ios/issues/691

mahibi:
fix #1740

additionally: check token from FCM periodically


app/src/gplay/java/com/nextcloud/talk/utils/ClosedInterfaceImpl.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override fun setUpPushTokenRegistration() {
+        val pushRegistrationWork = OneTimeWorkRequest.Builder(PushRegistrationWorker::class.java).build()
+        WorkManager.getInstance().enqueue(pushRegistrationWork)
+
+        val periodicPushRegistration = PeriodicWorkRequest.Builder(
+            GetFirebasePushTokenWorker::class.java, 15,  // TODO: discuss intervall. joas 24h, google 1 month
+            TimeUnit.MINUTES
+        )
+            .build()
```
mahibi, line 46: TODO: change intervall, 15 min is for testing..

app/src/main/java/com/nextcloud/talk/utils/PushUtils.java
```diff
@@ -226,162 +221,159 @@ public void pushRegistrationToServer() {
...
-                                Log.d(TAG, "Failed to parse account push data");
-                                accountPushData = null;
-                            }
-                        } else {
-                            accountPushData = null;
-                        }
 
-                        if (((TextUtils.isEmpty(providerValue) || accountPushData == null) && !userEntity.getScheduledForDeletion()) ||
-                                (accountPushData != null && !accountPushData.getPushToken().equals(token) && !userEntity.getScheduledForDeletion())) {
-
```
mahibi, line 258: i didn't like that if check, so first i refactored and came up with

grafik

After that, i reduced it to only check if user is marked for deletion. 
Because it doesn't hurt to update the token even if the device is still of the opinion that it should still be the same..?
what do you think @nickvergessen @tobiasKaminsky  ?

mahibi:
i saw that PushRegistrationWorker is called too often on app startup, because in ClosedInterfaceImpl

```
    override fun setUpPushTokenRegistration() {
        registerLocalToken()
        setUpPeriodicLocalTokenRegistration()
        setUpPeriodicTokenRefreshFromFCM()
    }
```
is executed. And the PeriodicWorkRequest's seem to execute immediately on initialization. So maybe best would be to somehow create the periodic workers but skip the first run on initialization...? Any idea @tobiasKaminsky ?

mahibi:
well with some testing i can say this is not yet THE solution to fix all missing notifications.
i can't reproduce it now, but i'm sure i had energysaving options turned off but when sending messages to phone "onMessageReceived" was not triggered..

app/src/main/java/com/nextcloud/talk/utils/PushUtils.java
mahibi: @nickvergessen ?

nickvergessen: LGTM

timkrueger (approved):

PR #1782, opened 2022-01-21
add openConversations to search

### LINKED ISSUE #1625: Join open conversations on mobile
The moderator can open a conversations  for all registered users:
image

In the browser every user can search for this group and join:
image

But on mobile the open conversations are not shown in the search:
image

mahibi:
resolve #1625 

grafik


app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
```diff
@@ -667,6 +722,10 @@ private void dispose(@Nullable Disposable disposable) {
                 roomsQueryDisposable != null && !roomsQueryDisposable.isDisposed()) {
             roomsQueryDisposable.dispose();
             roomsQueryDisposable = null;
+        } else if (disposable == null &&
+            openConversationsQueryDisposable != null && !openConversationsQueryDisposable.isDisposed()) {
+            openConversationsQueryDisposable.dispose();
+            openConversationsQueryDisposable = null;
         }
```
mahibi, line 729: i copied the roomsQueryDisposable stuff and named it openConversationsQueryDisposable for the open conversations.
but tbh i have no idea yet why there is this differentiation between the parameter "disposable"(line 717) and roomsQueryDisposable.

tobiasKaminsky: Seems to be a lazy function handling two things: disposing disposable and if this is null, it tries to dispose the roomsQueryDisposable property, which is only assigned in L464.
TBH, I currently do not see why we need this roomsQueryDisposable at all…
It seems to be just needed due to framework reasons…

nickvergessen:
Would it be possible to add a heading between the filtered list and the open conversations, similar to web?
It otherwise is a bit confusing and non-obvious (e.g. when you search for "public" on c.nc.c)

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
mahibi: > TBH, I currently do not see why we need this roomsQueryDisposable at all…

that's the point. I could refactor it be similar to most other uses of rxjava in talk-android. But there are some different variants used. Might make sense / might be random / might have to learn more about rxjava. Would like to talk about RxJava in the next android meeting.. So at least until then i'll keep it like it is.

mahibi:
> Would it be possible to add a heading between the filtered list and the open conversations, similar to web? It otherwise is a bit confusing and non-obvious (e.g. when you search for "public" on c.nc.c)

should be possible, will have a look

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
```diff
@@ -527,32 +558,19 @@ private void fetchData(boolean fromBottomSheet) {
                     }
 
                     adapter.updateDataSet(callItems, false);
+                    adapter.hideAllHeaders();  // TODO: the user still sees that hiding is done by animation. It's
+                    // still not hidden by default...
```
mahibi, line 562: Adding sections for the search works so far, but atm i have this problem:
On inital startup the header of the conversation list is shown for a second before it gets hidden (see video).

So https://github.com/davideas/FlexibleAdapter won't hide it by default, see this issue: https://github.com/davideas/FlexibleAdapter/issues/286
I could try the workaround https://github.com/davideas/FlexibleAdapter/issues/286#issuecomment-279675071 but i really would like to avoid that.

Also setting 
`adapter.setHeadersShown(false)`
before executing
`adapter.updateDataSet(callItems, false);`
has no effect.

Any ideas are welcome...

https://user-images.githubusercontent.com/2932790/150950863-8bcdfcbb-06c7-4be4-a0cf-e311d53bc2ab.mp4

mahibi: fyi @nickvergessen i'm stuck here atm. sections work so far, but for the bug in the video i have no idea...

mahibi: solved by handling 2 different lists for "normal" conversations. one with items that include header attribute and one without... not nice but does the job..

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
```diff
@@ -576,6 +593,95 @@ private void fetchData(boolean fromBottomSheet) {
                 });
     }
 
+    private void sortConversations(List<AbstractFlexibleItem> callItems) {
+        if (CapabilitiesUtil.hasSpreedFeatureCapability(currentUser, "last-room-activity")) {
```
nickvergessen, line 597: Nextcloud 14, which is the minimum already, so drop the else?

nickvergessen:
> should be possible, will have a look

Looks awesome now 💙

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
mahibi: @timkrueger as expected the deletion of this "last-room-activity" check led to the deletion of "CallItem" which in turn led to some refactoring which might affect the UserStatus PR, see commit 70b4eed6541e6d013b8104159faf01483b8b3a18
So maybe keep an eye on that while reviewing..

timkrueger: > So maybe keep an eye on that while reviewing..
I'll rebase that on the UserStatus PR when tthis one is on the master. 

I'll review this tomorrow.

timkrueger (approved):
Looks good. :+1: 

Thank you

PR #1819, opened 2022-02-16
Kotlin controller migration

AndyScherzinger:
migrate further controllers to kotlin


app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
```diff
@@ -0,0 +1,373 @@
...
+            String.format(
+                resources!!.getString(R.string.nc_server_helper_text),
+                resources!!.getString(R.string.nc_server_product_name)
+            )
+        )
+        binding.serverEntryTextInputLayout.setEndIconOnClickListener { checkServerAndProceed() }
+        if (resources!!.getBoolean(R.bool.hide_auth_cert)) {
+            binding.certTextView.visibility = View.GONE
+        }
+        if (resources!!.getBoolean(R.bool.hide_provider) ||
```
mahibi, line 107: reading from resources like this seems to fail

Method threw 'android.content.res.Resources$NotFoundException' exception.

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
```diff
@@ -0,0 +1,373 @@
...
+                binding.helperTextView.setOnClickListener {
+                    val browserIntent = Intent(
+                        Intent.ACTION_VIEW, Uri.parse(
+                            resources!!
+                                .getString(R.string.nc_providers_url)
+                        )
+                    )
+                    startActivity(browserIntent)
+                }
+            } else if (findAccounts(userUtils.users as List<UserEntity>).size > 0) {
```
mahibi, line 127: somehow always returns 0

mahibi: so SwitchAccountController is not executed at all atm.

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
AndyScherzinger: How did you produce the exception? because it works for me™️

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
AndyScherzinger: Also works fine for me, on my device I get 5 accounts from `userUtils`, filtered to 2 via `findAccounts`, do you have several Nextcloud accounts that also have Nextcloud talk _and aren't already added to Nextcloud Talk?

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
mahibi: ah just tricked myself while debugging. this seems to fail only when evaluating the expression...

grafik

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
mahibi: ok, didn't really have a look what findAccounts does ;)
so if i understand that right i have to install NC files and add an account there without to add it to NC android talk app. And account manager should find this?? this fails for me..

timkrueger (approved):
Yeah. More Kotlin, less code. :+1:

app/src/main/java/com/nextcloud/talk/controllers/ServerSelectionController.kt
timkrueger: 🤦 Approved to fast.

I will test this scenario also.

AndyScherzinger: The account manager will find all account configured on the device afaik and the filtering then throws out all which are already in use I think. Nevertheless, the important part is that you get a list > 0 from `userUtils` and this you can check via evaluate in the debugger :) At I can on AS on windows.

AndyScherzinger:
> Like @mahibi I can't trigger the import from the Files app. Is that worked in the past?

I didn't really test the actual import, but I get the helper text and clicking on it launches the accounts view with the filtered accounts. I didn't click on one of them afterward, because both controllers launched and UI work so far... Just did, tghen clicking on the user opens the weblogin flow with the user name prefilled. Guess that is "works as designed"

timkrueger:
> Guess that is "works as designed"
:+1: 
In my opinion this can be merged.

AndyScherzinger:
Thanks @timkrueger I also added another commit that helps with beatifying the display of existing accounts for import so we now also show a avatar placeholder and the server. Came up in my testing since I have several accounts with the same user-id on different servers, have fun choosing the right own. So now that works too 🚀

PR #1828, opened 2022-02-25
Reduce cron frequence to not spam actions

nickvergessen:


.github/workflows/stale.yml
```diff
@@ -1,7 +1,7 @@
 name: 'Close stale issues'
 on:
     schedule:
-        -   cron: '* */2 * * *'
+        -   cron: '0 0 * * *'
```
nickvergessen, line 4: [“At 00:00.” ](https://crontab.guru/#0_0_*_*_*)

.github/workflows/stale.yml
```diff
@@ -1,7 +1,7 @@
 name: 'Close stale issues'
 on:
     schedule:
-        -   cron: '* */2 * * *'
```
nickvergessen, line 4: [“At every minute past every 2nd hour.”](https://crontab.guru/#*_*/2_*_*_*)
So it previously executed 12*60 times... totally unnecessary

tobiasKaminsky (approved):

PR #1841, opened 2022-03-02
Bugfix/1839/fail to set empty status message

timkrueger:


mahibi:
it fails to set an icon without message

timkrueger:
> it fails to set an icon without message

Thanks for the hint. I revert my change to set "" in the case no custom status message is set: https://github.com/nextcloud/talk-android/pull/1841/commits/34117285977fdfec4941aa766ea1d6a1acf87530#diff-cc2c6d21c4982a3e76657411541f9258b5aa92404b8d5a1791013b87ed6be635R397

I add a trim before the received status message is shown to the use to avoid white spaces in the message box: https://github.com/nextcloud/talk-android/pull/1841/commits/5cb3a6597fc28c76ded275ffff184bdf5398e4d0

mahibi:
with the whitespace workaround:
if i set no message and then push the set status button, a whitespace is set as message. this leads to the fact this whitespace is shown instead of the online status messages ("Away", "Do not disturb"..) that should be shown if the user set no customized message.

Example:
image

marcel2 set an icon via web --> message is empty -> will be replaced with "Away"
Marcel set an icon via android --> message is whitespace -> whitespace is shown

i think it should be fixed on serverside api that it allows an empty message. So we wouldn't have to set a whitespace if we only want to set an emoji... right?

timkrueger:
> i think it should be fixed on serverside api that it allows an empty message. So we wouldn't have to set a whitespace if we only want to set an emoji... right?

I think we doing it wrong. That is sent by the web front end and it's working:

```
curl 'https://example.com/ocs/v2.php/apps/user_status/api/v1/user_status/message/custom?format=json' -X PUT --data-raw $'{"message":"","statusIcon":"\ud83d\ude07","clearAt":null}'
```

We currently send:

```
curl 'https://example.com/ocs/v2.php/apps/user_status/api/v1/user_status/message/custom?statusIcon=%F0%9F%98%80&message=%20" -X PUT
```

So we use parameters instead of a JSON body. 

I'll look into it tomorrow.

mahibi:
- [ ] if there is no message set, we should also set the view to GONE in the the account dialog, so there is not gap:

image

mahibi:
> I think we doing it wrong. That is sent by the web front end and it's working:
> 
> ```
> curl 'https://example.com/ocs/v2.php/apps/user_status/api/v1/user_status/message/custom?format=json' -X PUT --data-raw $'{"message":"","statusIcon":"\ud83d\ude07","clearAt":null}'
> ```

for me it's just not working at all. i get 200 as response, but in the response "icon", "message" and "clearAt" are null

mahibi:
fyi @timkrueger:
@Ivansss told me he remembers the same bug when he implemented it for iOS. So i created an issue for server: https://github.com/nextcloud/server/issues/31452

timkrueger:
> if it turns out to be a server issue: then we should leave this PR open to remember to undo the whitespace workaround once server is fixed.

I also talked with @Ivansss and asked him how he solved the problem in Nextcloud Talk for iOS. If the status message field is empty the button `Set status message` is disabled. I did the same for Android: https://github.com/nextcloud/talk-android/pull/1841/commits/cb1dd8e5b5278875beb56b382eecf30330294aab

I'll create a new issue and reference the https://github.com/nextcloud/server/issues/31452 issue.

AndyScherzinger (approved):
seems legit 👍

PR #1849, opened 2022-03-03
Bugfix/noid/migrate push to kotlin

AndyScherzinger:
builds on top of  #1847


AndyScherzinger:
❗ rebased

timkrueger (approved):
@AndyScherzinger I fixed a [performance warning](https://www.kaminsky.me/nc-dev/talk-findbugs/3188.html#Warnings_PERFORMANCE) in the `PushUtils` which was introduced by your changes. Then every check should be green and this can be merged.

Thank you :+1:

AndyScherzinger:
Thanks @timkrueger I also did some cargo culting for spotbugs but forgot to push it 🙈 Did push just now and then things should even be greener than green 😁

PR #1865, opened 2022-03-10
Conversation view shows search bar when opened from notification

### LINKED ISSUE #1861: Conversation view shows search bar when opened from notification
I have been able to repeat exactly the same behaviour as described by @mahibi.

> On my Redmi Note 9 Pro i can't reproduce the issue with the placeholders for the messages, but instead i got the following issue (on master, v13.0 as well as on bugfix/1765/notificationsFromLockScreen):
> With your steps to reproduce issue https://github.com/nextcloud/talk-android/issues/1765 i end up with a chat view that loads the messages correctly, but i see the searchbar at the top instead of the chat toolbar.
> grafik
> 
> This seems to be a lifecycle conflict between ChatController and ConversationListController. 
> At first ChatController#onAttach is executed which leads to 
> com/nextcloud/talk/controllers/base/NewBaseController.kt:175 
> to set "showSearchBar" = false.
> 
> AFTER that ConversationListController is executed which leads to 
> com/nextcloud/talk/controllers/base/BaseController.java:147
> to set "showSearchBar" = true.
> 
> So in the end the searchbar is shown..
> But as this seems to be a seperate issue, it doesn't mean your PR can't be merged..
> And as this might be conductor related together with the "ConductorRemapping#remapChatController" stuff it might not be fixed until we get rid of conductor..
> 
> _Originally posted by @mahibi in https://github.com/nextcloud/talk-android/issues/1766#issuecomment-1054587472_

This seems to be directly related to the sequence of events sent to the MainActivity when notification is opened from the lock screen.

When the sequence is: onNewIntent - onStart - onResume - **onPause** - **onStop** - **onStart** - **onResume** (including this strange onStop/onStart) the app behaves fine.

However when the sequence is just: onNewIntent - onStart - onResume (as one would expect :wink:), I observe the same issue.

I do not know yet what are the reasons of having those two sequences. I observe the following:

- Emulator - Android 11 - long sequence - app behaviour OK
- Emulator - Android 7 - short sequence - app shows this issue
- Device - an old Redmi phone - Android 7 - long sequence - app behaviour OK

I plan to dig deeper. :wink:

Probably a good way to resolve this issue would be to make sure that the **ConversationsListController** does not receive the **onAttach**/**onDetach** sequence **at all** when the chat notification is tapped on the lock screen.

This should have additional benefits:

- less flickering on screen if the app would go directly to the **ChatController**,
- not having a **getRooms** API call from the **ConversationsListController** that seems totally redundant in this scenario.

starypatyk:
Fixes #1861 

When **remapChatController** is called from a notification use a **SimpleSwapChangeHandler** rather than **HorizontalChangeHandler** to avoid animation.

This has the following effects:

- **ConversationsListController.onAttach** is not called in most cases (_occasionally when the **MainActivity** is destroyed and recreated the ConversationsListController.onAttach can still be called, but this seems rare_).
- The toolbar rather than the search bar is shown as expected.
- The **getRooms** API is **not** called.
- Screen flickering is significantly reduced. Sometimes the conversation list still becomes briefly visible, but in many cases the app shows the chat view immediately.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/MainActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/MainActivity.kt
@@ -349,7 +349,7 @@ class MainActivity : BaseActivity(), ActionBarProvider {
             } else {
                 remapChatController(
                     router!!, intent.getParcelableExtra<UserEntity>(KEY_USER_ENTITY)!!.id,
-                    intent.getStringExtra(KEY_ROOM_TOKEN)!!, intent.extras!!, false
+                    intent.getStringExtra(KEY_ROOM_TOKEN)!!, intent.extras!!, false, true
                 )
             }
         }
--- a/app/src/main/java/com/nextcloud/talk/utils/ConductorRemapping.kt
+++ b/app/src/main/java/com/nextcloud/talk/utils/ConductorRemapping.kt
@@ -24,15 +24,28 @@ import android.os.Bundle
 import com.bluelinelabs.conductor.Router
 import com.bluelinelabs.conductor.RouterTransaction
 import com.bluelinelabs.conductor.changehandler.HorizontalChangeHandler
+import com.bluelinelabs.conductor.changehandler.SimpleSwapChangeHandler
 import com.nextcloud.talk.controllers.ChatController
 
 object ConductorRemapping {
+
     fun remapChatController(
         router: Router,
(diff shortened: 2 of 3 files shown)


timkrueger (approved):
That is looking very good. Thank you very much @starypatyk

I made only one minor change in the `ConductorRemapping`.

Issue #1869, opened 2022-03-11
Intermittent crash when showing avatar


starypatyk:
## Steps to reproduce
1. Start Talk app
2. Open existing one to one conversation

### Expected behaviour
Conversation opened

### Actual behaviour
Sometimes the app crashes with the following error in the log.
<details>

```
--------- beginning of crash
03-08 21:49:37.800  6094  6094 E AndroidRuntime: FATAL EXCEPTION: main
03-08 21:49:37.800  6094  6094 E AndroidRuntime: Process: com.nextcloud.talk2, PID: 6094
03-08 21:49:37.800  6094  6094 E AndroidRuntime: java.lang.IllegalArgumentException: width and height must be > 0
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at android.graphics.Bitmap.createBitmap(Bitmap.java:877)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at android.graphics.Bitmap.createBitmap(Bitmap.java:856)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at android.graphics.Bitmap.createBitmap(Bitmap.java:787)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at android.graphics.Bitmap.createScaledBitmap(Bitmap.java:663)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at com.nextcloud.talk.controllers.ChatController$loadAvatarForStatusBar$1.onNewResultImpl(ChatController.kt:427)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at com.facebook.imagepipeline.datasource.BaseBitmapDataSubscriber.onNewResultImpl(BaseBitmapDataSubscriber.java:60)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at com.facebook.datasource.BaseDataSubscriber.onNewResult(BaseDataSubscriber.java:46)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at com.facebook.datasource.AbstractDataSource$1.run(AbstractDataSource.java:176)
03-08 21:49:37.800  6094  6094 E AndroidRuntime: 	at com.facebook.common.executors.UiThreadImmediateExecutorService.execute(UiThreadImmediateExecutorService.java:37)
[...]

PR #1870, opened 2022-03-11
Workaround for crash when loading avatar

starypatyk:
Workaround for #1869

I have not figured out why avatarSize becomes <= 0 in this case. With this workaround the app seems to behave OK - I have not even been able to notice that the avatar is incorrect.


timkrueger (approved):
Thank you again @starypatyk :-)

Looks good to me. When the checks are green (I have a look), I'll merge this.

PR #1872, opened 2022-03-16
solve issue of low contrast

iris-mobile-accessibility-repair:
The original text color of the component is '#FFFFFF', and the contrast between the text color ('#B3DAEF') and the background color ('#0082C9') does not meet the standard of WCAG 2.1. In other words, this color cannot be easily seen by everyone. So, to solve this problem, our team designed a tool that can automatically detect and repair such problems. The test report and recommended replacement colors ('#000000') are as follows:
image
The figure on the left is the original page, the figure on the right is the repaired page, and the figure below is the problem detection report.
If you think it is useful, please give me feedback. Your feedback is very important to me. We sincerely hope to receive your suggestions and opinions as an app developer.

### DIFF
--- a/app/src/main/res/layout/controller_server_selection.xml
+++ b/app/src/main/res/layout/controller_server_selection.xml
@@ -100,7 +100,7 @@
                 android:paddingBottom="2dp"
                 android:text="@string/nc_server_helper_text"
                 android:textAlignment="viewStart"
-                android:textColor="@color/nc_login_text_color" />
+                android:textColor="#000000" />
 
             <LinearLayout
                 android:id="@+id/serverEntryProgressBar"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
pinging @jancborchardt and @nimishavijay While I get the a11y scanner result and we are aware of this given Android a11y scanner, setting the text color to black seem wrong though since it solves the a11y issue but ignores any other aspect of a UI.

jancborchardt:
Yes, unfortunately the change to black text would look very bad.

The issue here is also that the Talk app does not yet use the proper log in screen like the Files app does. That one uses buttons and bigger text, so there are no or less issues. It would be good if the Talk app uses the same component @tobiasKaminsky

tobiasKaminsky:
Android Files has nearly the same screen.
I am happy to change it, if needed :)

iris-mobile-accessibility-repair:
> 
> 
> Android Files has nearly the same screen. I am happy to change it, if needed :)

The detection of the Google Accessibility Scanner shows that there is a problem with text contrast on this page. Through the color replacement algorithm of our tool, it is concluded that the color value we provide meets the standard, and it is recommended to modify it. 
Text or image contrast, always known as color-based accessibility issues, occurs when the color contrast between the text/image and the background is less than the minimum ratio specified by the accessibility guidelines.

PR #1873, opened 2022-03-16
Migrate ProfileController to Kotlin

AndyScherzinger:
* ``` subject```
* * drive-by fixing a NPE if you navigate back from the file broweser too quickly (async call to UI)


timkrueger (approved):
:blue_heart:

Issue #1852, opened 2022-03-04
black screen or single videoframe for audicalls (instead of avatar)


mahibi:
## Steps to reproduce
1. start audiocall from web
2. accept call on phone

### Expected behaviour
avatar is shown

### Actual behaviour
audio connection is okay but
either there is a black frame or a single videoframe from calling participant!

![grafik](https://user-images.githubusercontent.com/2932790/156815602-7b1865e4-17d4-4896-a7a0-70642bb6e5d4.png)

this must be caused by the webRTC lib update https://github.com/nextcloud/talk-android/pull/1774 because:
it works with commit 3af4f368
it fails with commit 89f64b39

## Device information

**Device:** Redmi Note 9 Pro

**Android version:** 10

**Talk version:** test with commit 89f64b39

## Server information

**Nextcloud version:** 24.0.0.4 (Sermo)

timkrueger:
I was able to narrow it a bit more.

Works: bad5b5f9
Works not: 86f20dcf

### Step 1 

The `MagicPeerConnectionWrapper` get passed the `remoteMediaStream` and it's containing a video and a audio stream. For both commits `remoteMediaStream#toString` is `[janus:A=1:V=1]`.

https://github.com/nextcloud/talk-android/blob/bad5b5f9bccae7f88eb533cf900094956487849f/app/src/main/java/com/nextcloud/talk/webrtc/MagicPeerConnectionWrapper.java#L407

### Step 2

The code at this point didn't change between the two commits. The `remoteMediaStream` is neither accessed or wrote again. But the `remoteMediaStream#toString` is different for both commits:
- bad5b5f9 `[janus:A=1:V=0]`.
- 86f20dcf `[janus:A=1:V=1]`

https://github.com/nextcloud/talk-android/blob/bad5b5f9bccae7f88eb533cf900094956487849f/app/src/main/java/com/nextcloud/talk/webrtc/MagicPeerConnectionWrapper.java#L359


The private variable `remoteMediaStream` is only accessed at this two points. I don't have any clue at which point the video stream is removed in the working commit bad5b5f9.

timkrueger:
[Migrating your native/mobile application to Unified Plan/WebRTC 1](https://docs.google.com/document/d/1PPHWV6108znP1tk_rkCnyagH9FK205hHeE9k5mhUzOg/edit#)

Issue #1773, opened 2022-01-14
🤖 Android update WebRTC library


nickvergessen:

mvarchdev:
Hello. I have a problem that when WebRTC Signalling server is used, everything is working fine, calls through browser on android phone and on pc are working. But on app Nextcloud Talk not... Is this issue releated to this?

timkrueger:
@mvarchdev: Sorry for the delay.

> app Nextcloud Talk not... Is this issue releated to this?

I think not. This issue is related only to update the underlying WebRTC library in Talk for Android. 

For help please ask in the [community](https://help.nextcloud.com/c/support/talk/52). 

If you're sure you found a new bug, please create an issue in the [spreed (Talk)](https://github.com/nextcloud/spreed/issues) repository or in the [talk-android](https://github.com/nextcloud/talk-android/issues) repository.

PR #1886, opened 2022-03-29
 Use Transceivers instead of constraints to discard received video tracks 

timkrueger:
__Many thanks to @danxuliu and @Ivansss to figure this out!__

After the migration from WebRTC from plan b to unified plan in commit https://github.com/nextcloud/talk-android/commit/86f20dcfd37b9f510474e576ea8bc4d4df05f0d8
the media constraints are ignored for creating an answer on a peer
connection:

> Offer to Receive Options/Constraints
>
> These constraints are passed to PeerConnection’s methods for creating SDP to
> add placeholder media sections in case a sending track would not have already
> created one.
>
> With Unified Plan semantics, these are still supported for creating offers
> using a backwards compatibility shim (basically creating an RtpTransceiver if
> one doesn’t exist). It is recommended to switch to the RtpTransceiver API
> (example below). _They are no longer supported when creating answers._
>
> [1]

Because of that now all transceivers of type VIDEO will be stopped if it
is only an audio call. After stopping the transceivers the status for the
video tracks will be automatically set to ENDED.

In the 'ParticipantsAdapter' it will be checked if the video tracks are
ENDED. In that case the user avatar will be shown like before.

Resolves: https://github.com/nextcloud/talk-android/issues/1852
See: https://github.com/nextcloud/talk-android/issues/1773, commit https://github.com/nextcloud/talk-android/commit/86f20dcfd37b9f510474e576ea8bc4d4df05f0d8, [1]

[1] https://docs.google.com/document/d/1PPHWV6108znP1tk_rkCnyagH9FK205hHeE9k5mhUzOg/edit#heading=h.9dcmkavg608r

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java
+++ b/app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java
@@ -44,9 +44,11 @@ import org.webrtc.DataChannel;
 import org.webrtc.IceCandidate;
 import org.webrtc.MediaConstraints;
 import org.webrtc.MediaStream;
+import org.webrtc.MediaStreamTrack;
 import org.webrtc.PeerConnection;
 import org.webrtc.PeerConnectionFactory;
 import org.webrtc.RtpReceiver;
+import org.webrtc.RtpTransceiver;
 import org.webrtc.SdpObserver;
 import org.webrtc.SessionDescription;
 import org.webrtc.VideoTrack;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java
```diff
@@ -256,6 +258,15 @@ public boolean isMCUPublisher() {
         return isMCUPublisher;
     }
 
+    private boolean isAudioCallOnly() {
```
danxuliu, line 261: Rather than `isAudioCallOnly()` it might be better to name it something like `shouldReceiveVideo()`. If the PeerConnectionWrapper is the sender peer to the HPB [`OfferToReceiveVideo` would be false too](https://github.com/nextcloud/talk-android/blob/1b30a6ef62dde17b9eb95a6867d03089a5b3dea9/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L447), so in that case `isAudioCallOnly()` would return `true` even in a video call.

Right now this should not be a problem, as `isAudioCallOnly()` is used only from `onSetSuccess()` if there is no local description, so in practise it will not be called for the sender peer to the HPB. But if the method is named `isAudioCallOnly()` it could be misused in the future :-)

danxuliu (commented):
Other than the comment, code (and commit message :-) ) looks good :+1: (but I have **not** tested it!)

mahibi (approved):

mahibi (approved):

PR #1895, opened 2022-04-01
Reactions to chat messages

### LINKED ISSUE #1772: Reactions to chat messages
After being able to reply, replying with a reaction emoji is the next step.
An API is currently developed in https://github.com/nextcloud/spreed/issues/1920
Once done the mobile clients and web UI should show the summary on messages and allow reacting as well.

mahibi:
resolve #1772 

|Message|Reactions menu|Reactions listing|
|---|---|---|
|grafik|grafik|grafik|

### remaining TODOs:

- [x] send reaction
    - [x] emoji picker
- [x] delete reaction
- [x] tab layout
  - [ ] add view pager if time permits
- [x] show reactions for all MessageViewHolders (incoming + outgoing messages)
    - [x] TextMessages
    - [x] VoiceMessages
    - [x] Location
    - [x] Previews
- [x] set textcolor for reaction for outgoing messages
- [x] sort emojis in message view descending by amount
- [ ] analyze why some emojis are not shown
   - grafik
- [x] sort reaction list
    - [x] own user to the top - see sorting rule at https://github.com/nextcloud/talk-android/issues/1772#issuecomment-1088656144
         - [ ] regarding the "all" view, first rule should better be the amount of reaction? @AndyScherzinger 

### known issues:
- [ ] emoji picker SOMETIMES (but unfortunately most of the times on first usage) opens in full screen height. see https://github.com/vanniktech/Emoji/issues/474
   - grafik


AndyScherzinger:
@mahibi changinging the sorting for the all tab can be done but shouldn't block this PR - can't say if/when I find the time or you can give it a go.

XueSheng-GIT:
I downloaded the apk to test the reactions. Adding a reaction only works if there's no reaction yet (reaction menu pops up). But as soon as a reaction exists (from myself or another person), the reaction listing is shown (which doesn't seem to allow to add additional reactions...). I couldn't find a way to add a reaction in this case.
How to switch from the reaction listing to the reaction menu?

mahibi:
> I downloaded the apk to test the reactions. Adding a reaction only works if there's no reaction yet (reaction menu pops up). But as soon as a reaction exists (from myself or another person), the reaction listing is shown (which doesn't seem to allow to add additional reactions...). I couldn't find a way to add a reaction in this case. How to switch from the reaction listing to the reaction menu?

a long click on the message will open the Reactions menu where you can set an emoji. to set more emojis just do this another time. This should work regardless if there are already emojis or not!

if reactions exist, you can tap on them to see details in the Reactions listing. In this view you can't add reactions but just view them (or delete your own reactions when you tap on them.)

Does this answer your question? Or is there some bug?

XueSheng-GIT:
> a long click on the message will open the Reactions menu where you can set an emoji. to set more emojis just do this another time. This should work regardless if there are already emojis or not!

For me this only works if there's no reaction yet. Long press on an existing reaction does nothing (while pressing). But when releasing the finger, the reaction listing appears.

mahibi:
> > a long click on the message will open the Reactions menu where you can set an emoji. to set more emojis just do this another time. This should work regardless if there are already emojis or not!
> 
> For me this only works if there's no reaction yet. Long press on an existing reaction does nothing (while pressing). But when releasing the finger, the reaction listing appears.

if there are emojis in the message, there's a difference if you tap on the message text or the emojis themselves. 
so a long press on the **message text** will open the Reactions menu.
a press on the emojis will open the Reactions listing view.

does this work for you?

But maybe it should be changed that a long tap on the emojis maybe should also open the Reactions menu....

AndyScherzinger:
> But maybe it should be changed that a long tap on the emojis should also open the Reactions menu....

for long press, I'd say yes because that would be the expected result I believe.

mahibi:
> UPDATE: Just realized that long press works if you exactly hit the text iself. Long press on existing reactions does not open the reactions menu. A bit confusing and not obvious.

just changed this. that was a good hint, thanks for testing @XueSheng-GIT :+1:

mahibi:
merged without approval to get towards the next RC (tonight or in the next days i guess).

thanks for your help @AndyScherzinger :+1:

AndyScherzinger:
My pleasure 😊

PR #1896, opened 2022-04-02
Additional fix related to opening chat view from notification

starypatyk:
A follow-up to my previous PR #1865 (an initial fix for #1861).

I have found that the following scenario:

> occasionally when the **MainActivity** is destroyed and recreated the **ConversationsListController.onAttach** can still be called, but this seems rare

is much more frequent than I originally anticipated. It seems to happen almost always when Talk is closed before locking the screen. :disappointed: 

On emulator the app works fine, however on my two Xiaomi devices (an old Redmi with Android 7 and a newer Mi Note device with Android 10) I observe a strange behaviour. Talk enters a state with the chat list still visible - despite the **ChatController** being pushed onto the **Conductor** stack.

screenshot

I have not been able to find the reason for such behaviour. It seems that on some devices **Conductor** does not handle **setRoot** and **pushController** calls coming quickly one after the other in conjunction with the **onStart**/**onResume**/**onPause** sequence that happens when opening the notification from the lock screen. _I suspect some race condition between Conductor and Android system code._

This was not obvious when the transition to the **ChatController** used the **HorizontalChangeHandler** that involved animation (and delay). However after switching to the **SimpleSwapChangeHandler** (apparently instantaneous) the issue became much more annoying.

I propose a change that uses a [single **setBackstack** call on the router](https://github.com/nextcloud/talk-android/blob/bugfix/1861/remap-chat-controller/app/src/main/java/com/nextcloud/talk/utils/ConductorRemapping.kt#L81) rather than **setRoot** followed by **pushController** (in this scenario only). All the other changes are required to enable this one.

This seems to fix the issue.


mahibi:
thank you for continuing to contribute to this @starypatyk :+1: 
please note that the PR's might be reviewed late due to the easter vacations..
i might test this next week, but maybe it will be reviewed after easter..

mahibi:
thank you for continuing to contribute to this @starypatyk :+1: 
please note that the PR's might be reviewed late due to the easter vacations..
i might test this next week, but maybe it will be reviewed after easter..

PR #1922, opened 2022-04-17
Migrate OperationsMenuController to kotlin

AndyScherzinger:


app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.kt
```diff
@@ -0,0 +1,778 @@
...
+@AutoInjector(NextcloudTalkApplication::class)
+class OperationsMenuController(args: Bundle) : NewBaseController(
+    R.layout.controller_operations_menu,
+    args
+) {
+    private val binding: ControllerOperationsMenuBinding by viewBinding(ControllerOperationsMenuBinding::bind)
+
+    @JvmField
+    @Inject
+    var ncApi: NcApi? = null
```
timkrueger, line 92: Can this be changed to the following?

```java
    @Inject
    lateinit var ncApi: NcApi
```

detekt.yml
```diff
@@ -1,5 +1,5 @@
 build:
-  maxIssues: 95
+  maxIssues: 100
```
timkrueger, line 2: I will try to get the `maxIssues` back to 95.

timkrueger (changes requested):

app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.kt
AndyScherzinger: Yeah, makes sense 👍

timkrueger: I can do if you not already did.

timkrueger (approved):

PR #1925, opened 2022-04-19
Catch NPE when UI is already teared down

AndyScherzinger:


mahibi:
:+1:

PR #1936, opened 2022-04-21
fix text for reaction notification + hide content title for one2one rooms and reactions

### LINKED ISSUE #1935: fix notification subject for reaction
Display correctly reactions notifications.
Reference: https://github.com/nextcloud/spreed/issues/7187

mahibi:
fix #1935 

| broken notification before (marcel2 reacted to "Hello")| fixed notification (also the "content title" is hidden)|
|---|---|
|grafik|grafik|

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.java
+++ b/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.java
@@ -78,6 +78,7 @@ import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.util.HashMap;
+import java.util.Objects;
 import java.util.zip.CRC32;
 
 import javax.crypto.Cipher;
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
fine by me! Even though for a future discussion I think we should discuss it reactions are worth a push notification or at least if that should be displayed as such because my view is that they are not "important enough" to be push up the users phone. What do you think @jancborchardt @nickvergessen - also as I said this discussion is for post-14.0.0 then

nickvergessen:
We only push them when you are fully subscribed to the conversation (so default only the case in one-to-one conversations)

Btw the expected fixed version from my POV would be:

> **Subject as delivered**
> Message that was reacted on


So in your sample
> **marcel hat mit 🥰 auf ihre private Nachricht reagiert**
> Hello

mahibi:
> Btw the expected fixed version from my POV would be:
> 
> > **Subject as delivered**
> > Message that was reacted on
> 
> So in your sample
> 
> > **marcel hat mit smiling_face_with_three_hearts auf ihre private Nachricht reagiert**
> > Hello

@nickvergessen @AndyScherzinger 

As far as i see this is against android's concept of notifications for chat apps.
We need to set a person object which name is used to set the contentTitle.

androids documentation: https://developer.android.com/reference/androidx/core/app/NotificationCompat.MessagingStyle
where we set it right now: https://github.com/nextcloud/talk-android/blob/30d57fda4f616a4b77b794a977deb651887fd585/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.java#L449

So i guess that's why signal and telegram mix the reaction and the referred message in one text:
grafik

So this could be an option for us too, but still would be different to iOS then.. (+ it would require own translation...)
Should we do this for 14.0.0 or keep it like it is for now?


PS: From my POV the display of notifications (NotificationsWorker.java) should get more attention after 14.0.0, right now it's a bit complicated..

AndyScherzinger:
I'd prefer consistency as a start so the questions would be 

* how is this solved throughout other messengers on iOS
* if there are consistent patterns per platform I might still tend to then go with each platform since that matches the user's expectation.

maybe @jancborchardt also has a take on this?

PR #1942, opened 2022-04-22
add workaround to avoid emoji picker to take the full screen height

### LINKED ISSUE #1914: emoji picker sometimes opens in full screen height
## Steps to reproduce
1. react to message with custom emoji (-> open emoji dialog by "..." button)

### Expected behaviour
emoji popup always open with the same size that the keyboard would have

### Actual behaviour
- emoji picker SOMETIMES (but unfortunately most of the times on first usage) opens in full screen height. see https://github.com/vanniktech/Emoji/issues/474
   - grafik

this behaviour was fixed for other occurences of the emoji picker inside the talk app. It might be more difficult to find a workaround this time because it's now used inside a BottomDialog?!

related:
https://github.com/nextcloud/talk-android/pull/1668#issuecomment-1032746866

talk app version: 14.1.0 Alpha 1

mahibi:
first try to fix #1914 

---

The toggleEmojiPopup method is a hacky workaround to avoid bug #1914
As the bug happens only for the very first time when the popup is opened,
it is closed after some milliseconds and opened again.

200 milliseconds seems to be a good value to initialize the popup correctly with the desired size

downside: there is even some flickering when opening the "more emojis" window

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
@@ -147,7 +147,7 @@ class MessageActionsDialog(
                     popup.dismiss()
                     popup.show()
                 },
-                200
+                DELAY
             )
         }
     }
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
fine with me as a workaround, and yeah, some flickering but still better than the full-screen original state

PR #1950, opened 2022-04-29
Hide shared items toolbar on scroll

AndyScherzinger:


timkrueger:
@AndyScherzinger  Is this for UI behavior alignment with Android Files? 

Currently also if you're scrolling in a conversation the app bar is always visible.

Personally I prefer it to have the app bar available all the time. 

@mahibi What your opinion?

mahibi:
i also prefer to hide the toolbar

> Currently also if you're scrolling in a conversation the app bar is always visible.

that's an exception that is okay for chat apps i would say..

mahibi (approved):

timkrueger:
OK. Then I'm fine with that 😃

timkrueger (approved):

PR #1969, opened 2022-05-03
Add old screenshot from F-Droid

timkrueger:
In the F-Droid store screenshots can only be added or updated. So to get
rid of the old screenshots a 1x1px version is added. After the next
release to the F-Droid store they can be deleted. From that point on
existing screenshots should only be updated!

See: https://gitlab.com/fdroid/fdroidserver/-/issues/490


AndyScherzinger (approved):
let's see if this works 👍 f-droid screenshots isn't very easy/straight-forwards in my view

PR #1973, opened 2022-05-04
Reduce number of Detekt issues

timkrueger:
PR #1969 is failing because of they are too much of them

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/LocationPickerController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/LocationPickerController.kt
@@ -213,7 +213,7 @@ class LocationPickerController(args: Bundle) :
         return true
     }
 
-    @Suppress("Detekt.TooGenericExceptionCaught")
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/controllers/LocationPickerController.kt
```diff
@@ -213,7 +213,7 @@ class LocationPickerController(args: Bundle) :
         return true
     }
 
-    @Suppress("Detekt.TooGenericExceptionCaught")
+    @Suppress("Detekt.TooGenericExceptionCaught", "Detekt.ComplexMethod")
```
AndyScherzinger, line 216: still needed since you refactored the method?

timkrueger: Yes. The method is still too complex Detekt says. But I can live with it.

But if you insist I will look again on it :-)

AndyScherzinger: Nah, all good then 👍

AndyScherzinger (approved):

Issue #1881, opened 2022-03-22
Implement handling for "event.participants.update.all" from HPB


nickvergessen:
## Steps to reproduce
0. Run HPB
1. Create a conversation
2. Restrict permissions
3. Start a call on web as moderator
4. Join as a normal user in the app
5. As moderator "End meeting for everyone"

### Expected behaviour
The new event of the HPB is handled gracefully:
* All connections are terminated
* The mobile client leaves the call

```json
{"type":"event","event":{"target":"participants","type":"update","update":{"roomid":"5439271893","incall":0,"all":true}}}
```

### Actual behaviour
The event is not handled

### Ref
The spreed PR will only be merged once iOS and Android understand that message and correctly leave the meeting, but currently the "End meeting for all" does not work with more than 50 users, so this should be fixed rather soon:

* https://github.com/nextcloud/spreed/pull/6948
* https://github.com/strukturag/nextcloud-spreed-signaling/pull/208
* https://github.com/nextcloud/talk-ios/issues/744

nickvergessen:
iOS code for inspiration: https://github.com/nextcloud/talk-ios/pull/747/files

timkrueger:
I think this is a duplicate of #1660. It will be released with 14.0.0 and is testable with the latest [RC](https://github.com/nextcloud/talk-android/releases/tag/v14.0.0rc4).

nickvergessen:
No, it was a follow up change we realised after #1660 got used on a big instance

nickvergessen:
Basically something like this: https://github.com/nextcloud/talk-android/commit/f70f94e6be18c07e64bdb1b6491b48407c98bee1
Needs to happen if `all: true` is part of the HPB event and there is no participant list:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1751-L1757

nickvergessen:
Ref: 
https://github.com/strukturag/nextcloud-spreed-signaling/pull/208/files#diff-9752dbbe894845f1e7cc6dfe36a6775c2b0c6b5633f9700c31d413bc44914cea

So there is no `changed` array and no `users` array in this event. so I guess you also need to make sure that part is not invoked or breaking on the json parsing

PR #1979, opened 2022-05-04
Add handling for "event.participants.update.all" 

timkrueger:
In case a moderator of a restricted room ends the call for all participants a update participants event with the field 'all=true' will
be thrown by the HPB.

```json
    {
      "type": "event"
      "event": {
        "target": "participants",
        "type": "update",
        "update": [
          "roomid": "the-room-id",
          "incall": new-incall-state,
          "all": true
        ]
      }
    }
```

In that case the call can be ended directly without handling every single participant.

Resolves: https://github.com/nextcloud/talk-android/issues/1881


timkrueger:
> # SpotBugs increased!

I started fixing that and continue this evening.

mahibi:
tested with sermo which currently supports the "update all" request..
but i have to say with this current state "everything" is broken.. can't establish a successful call, web keeps ringing after hangup, phone keeps ringing 2 times in parallel etc..
we should debug this together i guess.

timkrueger:
> tested with sermo which currently supports the "update all" request.. but i have to say with this current state "everything" is broken.. can't establish a successful call, web keeps ringing after hangup, phone keeps ringing 2 times in parallel etc.. we should debug this together i guess.

🤦 

I added the wrong constant in the [`CallActivity`](https://github.com/nextcloud/talk-android/blob/d05566b67d75742a2be25181847bb65f46f8a3fc/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1488):

```
-            case TARGET_PARTICIPANTS:
+            case PARTICIPANTS_UPDATE:
                 Log.d(TAG, "onMessageEvent 'participantsUpdate'");
```

Can you please try again @mahibi ?

mahibi:
LGTM code wise, didn't test because sermo was not set up with the hpb changes. 
so remember to test it once everything is merged..

mahibi (approved):

mahibi:
i guess @nickvergessen wants to have this in for 14.0.2 ?

AndyScherzinger:
@timkrueger can you manually backport please?

PR #1988, opened 2022-05-05
Add empty view visuals part to shared items

### LINKED ISSUE #1984: "no files shared yet" for mediatab
The mediatab feature currently shows a black screen when no files were shared at all.

In this case, there should be at least an info like "no files shared yet"

AndyScherzinger:
Resolves #1984 

@timkrueger maybe you can pick this one up since this would also add the first scenario where the new patetrn has to handle general "state changes" as in 
* loading
* success (display the data)
* error

Before the new pattern in place we i.e. did it like this in here https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/controllers/ProfileController.kt#L247-L253

So now I guess we need to emit such events so the UI can be changed to hide&/show certain parts, maybe @AlvaroBrey or you have an idea how that would need to be done. Given the current implementation I would expect the view model to emit the state and the activity to handle the show/hide? Not sure though. Also I believe that at some point in the near future we would also need to move most of the activity to a fragment but to be discussed I guess.

### DIFF
--- a/app/src/main/res/layout/activity_shared_items.xml
+++ b/app/src/main/res/layout/activity_shared_items.xml
@@ -56,6 +56,11 @@
 
     </com.google.android.material.appbar.AppBarLayout>
 
+    <include
+        android:id="+id/emptyContainer"
+        layout="@layout/empty_list"
+        android:visibility="gone" />
+
     <androidx.recyclerview.widget.RecyclerView
         android:id="@+id/image_recycler"
         android:layout_width="match_parent"


AlvaroBrey:
> So now I guess we need to emit such events so the UI can be changed to hide&/show certain parts, maybe @AlvaroBrey or you have an idea how that would need to be done. Given the current implementation I would expect the view model to emit the state and the activity to handle the show/hide? Not sure though. Also I believe that at some point in the near future we would also need to move most of the activity to a fragment but to be discussed I guess.

I know of two common strategies:
a) Have a separate LiveData for each thing we want to observe

```kotlin
class FooViewModel: ViewModel(){
...
     val data: LiveData<List<MyDataClass>>
     val loading: LiveData<Boolean>
     val error: LiveData<ErrorCode>
```

b) Publish a single LiveData with the entire state. I like this one more, as view logic  becomes much simpler

```kotlin
class FooViewModel: ViewModel(){
    sealed interface ViewState;
    object Loading: ViewState;
    data class Error(val errorCode: ErrorCode): ViewState;
    data class Loaded(val data: List<MyDataClass>): ViewState;
...
   val state: LiveData<ViewState>
```

```kotlin
    viewModel.state.observe(this){ state->
       when(state){
          is Loading -> showLoading()
          is Error -> showError(state.errorCode)
          is Loaded -> showData(state.data)
       }
    }
```

Of course, these approaches can be mixed. For example, if you want to still show the previous data while loading or if there is an error, it would make sense to keep `data` in a separate `LiveData` from the state.

AlvaroBrey:
This is done in #2015

PR #2019, opened 2022-05-10
Feature/1948/respect chat permission

### LINKED ISSUE #1948: Respect "🛃 Chat permission" for 25
If the server has the capability `chat-permission` a new permission `128` is required to be able to:
* Post and delete chat messages
* React and delete reactions
* Sharing files to the chat/conversation
* Sharing objects (location, contact, voice messages) to the chat

Server PR: https://github.com/nextcloud/spreed/pull/7235
Server Issue: https://github.com/nextcloud/spreed/issues/3886
Target version: 25

mahibi:
resolve #1948

new permission 128:
https://nextcloud-talk.readthedocs.io/en/latest/constants/#attendee-permissions

# how to test
post content to chat (post chat message, share items, do reactions). If permission is missing, the views/buttons in the android app to trigger these actions should be hidden.

- older server without chat-permission capability: posting of content should just be allowed
- server with chat-permission capability: in the participants list on the right, select to edit the permissions and enable/disable "can send messages/reactions"
grafik
- enable/disable permission for all participants
grafik

# remaining todos:
- [x] restrict "sendTo"
   - [x] double check just before sending file
- [x] move permission check from initialisation of ChatController back to the places where it's needed? -> on the fly check when permissions were changed while being in chat
   - done by updating this in getRoomInfo 
- [x] forwarding messages to this chat must be forbidden

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -2492,6 +2492,7 @@ class ChatController(args: Bundle) :
                 currentConversation,
                 chatMessage,
                 conversationUser,
+                hasChatPermission,
                 ncApi!!
             ).show()
         }
@@ -2519,6 +2520,7 @@ class ChatController(args: Bundle) :
                     conversationUser,
                     currentConversation,
                     isShowMessageDeletionButton(message),
+                    hasChatPermission,
                     ncApi!!
                 ).show()
             }
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -346,7 +346,9 @@ class ChatController(args: Bundle) :
                         setTitle()
 
                         hasChatPermission =
-                            AttendeePermissionsUtil(currentConversation!!.permissions).hasChatPermission(conversationUser)
+                            AttendeePermissionsUtil(currentConversation!!.permissions).hasChatPermission(
+                                conversationUser
+                            )
 
                         try {
                             setupMentionAutocomplete()
@@ -1458,6 +1460,12 @@ class ChatController(args: Bundle) :
 
     private fun uploadFiles(files: MutableList<String>, isVoiceMessage: Boolean) {
         var metaData = ""
+
+        if (!hasChatPermission) {
+            Log.e(TAG, "uploading file(s) is forbidden because of missing attendee permissions")
+            return
+        }
+
         if (isVoiceMessage) {
             metaData = VOICE_MESSAGE_META_DATA
         }
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -147,6 +147,7 @@ import com.nextcloud.talk.ui.dialog.ShowReactionsDialog
 import com.nextcloud.talk.ui.recyclerview.MessageSwipeActions
 import com.nextcloud.talk.ui.recyclerview.MessageSwipeCallback
 import com.nextcloud.talk.utils.ApiUtils
+import com.nextcloud.talk.utils.AttendeePermissionsUtil
 import com.nextcloud.talk.utils.ConductorRemapping
 import com.nextcloud.talk.utils.ConductorRemapping.remapChatController
 import com.nextcloud.talk.utils.ContactUtils
@@ -2830,6 +2831,8 @@ class ChatController(args: Bundle) :
         if (!isUserAllowedByPrivileges) return false
 
         if (!CapabilitiesUtil.hasSpreedFeatureCapability(conversationUser, "delete-messages")) return false
+
+        if (AttendeePermissionsUtil(currentConversation!!.permissions).canPostChatShareItemsDoReaction(conversationUser)) return true
 
         return true
     }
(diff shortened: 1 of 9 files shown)


app/src/test/java/com/nextcloud/talk/utils/AttendeePermissionsUtilTest.kt
```diff
@@ -0,0 +1,47 @@
...
+
+class AttendeePermissionsUtilTest : TestCase() {
+
+    @Test
+    fun test_areFlagsSet() {
+        val attendeePermissionsUtil =
+            AttendeePermissionsUtil(
+                AttendeePermissionsUtil.PUBLISH_SCREEN or
+                    AttendeePermissionsUtil.JOIN_CALL or
+                    AttendeePermissionsUtil.DEFAULT
```
mahibi, line 34: not to myself.. one thing i noticed while writing this test:

if AttendeePermissionsUtil.DEFAULT is not set here and checked with assertFalse, the test fails. i should have another look at this..

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -306,6 +309,9 @@ class ChatController(args: Bundle) :
         }
 
         this.voiceOnly = args.getBoolean(BundleKeys.KEY_CALL_VOICE_ONLY, false)
+
+        hasChatPermission =
+            AttendeePermissionsUtil(currentConversation!!.permissions).hasChatPermission(conversationUser)
```
AlvaroBrey, line 314: Why is this initialized both here and in `getRoomInfo()`? Just trying to understand the flow of this controller

AlvaroBrey (approved):
Looks good besides detekt

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
mahibi: well at first i only initialized in `init`. 
Then i added it also to getRoomInfo() to upadte the UI on the fly when permissions on server were changed.

i was also thinking about to remove it from `init`. This should be okay and work, but i was not 100% sure if somehow hasChatPermission can be accessed before getRoomInfo()...

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -2512,50 +2543,59 @@ class ChatController(args: Bundle) :
...
-
-        ncApi?.deleteChatMessage(
-            credentials,
-            ApiUtils.getUrlForChatMessage(
-                apiVersion,
-                conversationUser?.baseUrl,
-                roomToken,
-                message?.id
+        if (!hasChatPermission) {
+            Log.e(
```
AndyScherzinger, line 2547: this is info or warning level imho

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -2831,6 +2871,8 @@ class ChatController(args: Bundle) :
 
         if (!CapabilitiesUtil.hasSpreedFeatureCapability(conversationUser, "delete-messages")) return false
 
+        if (!hasChatPermission) return false
```
AndyScherzinger, line 2874: to be combined with the above check or adding a return when for all 3 cases

app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
```diff
@@ -82,7 +83,8 @@ class MessageActionsDialog(
         )
         initMenuDeleteMessage(showMessageDeletionButton)
         initMenuForwardMessage(
-            ChatMessage.MessageType.REGULAR_TEXT_MESSAGE == message.getMessageType() &&
+            hasChatPermission &&
```
AndyScherzinger, line 86: wouldn't forward be something ending up in another chat and thus not be blocked by this chat permission, since the permission is set on the room?

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
mahibi: i don't get the point yet.

you want a when condition like

when {
    condition1 -> ...
    condition2 -> ...
    condition3 -> ...
}

for the last 3 cases?

app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
mahibi: see commit message https://github.com/nextcloud/talk-android/pull/2019/commits/b603031011c5fef7f1bfceb6f275e2fde0017566

but tbh i don't get the point anymore why i had this opinion. will check again...

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
mahibi: used return when for more cases here..

AndyScherzinger: Yeah, not limited for 3 cases of course, but that is just what is visible in GH's review view, when not expanding the diff view ;)

AndyScherzinger (approved):
LGTM

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
mahibi: well, in fact it crashed when opening chat from notification when it's done in init..
https://github.com/nextcloud/talk-android/pull/2058

PR #2047, opened 2022-05-14
Parceler migration

AndyScherzinger:
migrating further parts of the api classes


scripts/hooks/pre-commit
```diff
@@ -5,6 +5,7 @@ echo "Running pre-commit checks..."
 if ! ./gradlew --daemon ktlintCheck &>/dev/null; then
     echo >&2 "ktlint failed! Run ./gradlew ktlintCheck for details"
     echo >&2 "Hint: fix most lint errors with ./gradlew ktlintFormat"
+	echo >&2 "Detekt failed! See report at file://$(pwd)/app/build/reports/ktlint/ktlintMainSourceSetCheck/ktlintMainSourceSetCheck.txt"
```
AlvaroBrey, line 8: `s/Detekt/ktlint/`

The reason that I didn't include this in the original hooks in Files, and instead direct the user to run `ktlintCheck` for results, is that `ktlint` generates a separate report for each sourceset. If you have an error in the `test` sourceset, for example, it won't show up in the linked report.

AndyScherzinger: So remove it again @AlvaroBrey

AlvaroBrey: I would remove it, yeah.

AndyScherzinger: DONE 👍

mahibi (approved):

timkrueger (approved):
More 💙💙💙 for @AndyScherzinger :smile:

PR #1923, opened 2022-04-18
Reply from notification

starypatyk:
I wanted to make an experiment - how difficult it is to add a **Reply** action to the notifications. Turned out to be simpler than expected. 😉

This should partially resolve #230 and #1428. This is also a pre-requisite for #1208.

This seems to work already, however I would like to make some additional changes:

- [x] ~~Notify the user when reply could not be sent~~
- [x] Refactor code in NotificationWorker.java:
  - [x] [Extract common code to load user avatars for notifications](https://github.com/nextcloud/talk-android/pull/1923/commits/59ccfd8044e42e365b749de9ba592dca8ed75c5a)
  - [x] [Extract code handling chat (message) notifications to shorten the showNotification function](https://github.com/nextcloud/talk-android/pull/1923/commits/b2510041790e17d18d0452c6fb215287bdaff6c7)
- [x] Additional refactoring [here](https://github.com/nextcloud/talk-android/pull/1923/commits/296cb0721fc9fb97adcea415f631f31a1773596c), [here](https://github.com/nextcloud/talk-android/pull/1923/commits/2857e40634fa61cc5875813dfbb50d4c40a9be63) and [here](https://github.com/nextcloud/talk-android/pull/1923/commits/7a0d58120aa874db7e81ef94a20767094cbc706b).
- [x] [Remove suspicious setAction call](https://github.com/nextcloud/talk-android/pull/1923/commits/c9a3f75384f6cbc1aceea24aaceb130d8d2ea79e)

PR #1259, opened 2021-05-17
Show a temporary messages when sending and keep it on error

nickvergessen:
Success (bonus weird scrolling) | Error case (with copy/paste option)
---|---
Peek 2021-05-17 12-00 | Peek 2021-05-17 12-01

I have no clue why the scrolling freaks out like this.

### Code changes to help testing
Modify your Talk install to make the response slower:
```diff
diff --git a/lib/Controller/ChatController.php b/lib/Controller/ChatController.php
index f8e0f19a6..e71bd1d81 100644
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -226,6 +226,7 @@ class ChatController extends AEnvironmentAwareController {
         *         found".
         */
        public function sendMessage(string $message, string $actorDisplayName = '', string $referenceId = '', int $replyTo = 0): DataResponse {
+               sleep(1);
                [$actorType, $actorId] = $this->getActorInfo($actorDisplayName);
                if (!$actorId) {
                        return new DataResponse([], Http::STATUS_NOT_FOUND);

```
And use the follow change to create a slow error
```diff
diff --git a/lib/Controller/ChatController.php b/lib/Controller/ChatController.php
index f8e0f19a6..e71bd1d81 100644
--- a/lib/Controller/ChatController.php
+++ b/lib/Controller/ChatController.php
@@ -226,6 +226,8 @@ class ChatController extends AEnvironmentAwareController {
         *         found".
         */
        public function sendMessage(string $message, string $actorDisplayName = '', string $referenceId = '', int $replyTo = 0): DataResponse {
+               sleep(1);
+               return new DataResponse([], Http::STATUS_BAD_REQUEST);
                [$actorType, $actorId] = $this->getActorInfo($actorDisplayName);
                if (!$actorId) {
                        return new DataResponse([], Http::STATUS_NOT_FOUND);
```

### Follow up todos
- [ ] Fix scrolling (I have no idea where it goes wild)
- [ ] Messages can have a reference id which should be set and the temporary messages only be removed by the fetching of messages when there is a reference id match.

Fix #838

PR #2048, opened 2022-05-14
Reply from notification - initial error handling

starypatyk:
Follow-up to #1923.

I have implemented a very basic notification when the reply cannot be sent. This is how it looks like at the moment.
obraz

The way this works and looks should probably be aligned with #1259.

Any suggestions?

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/receivers/DirectReplyReceiver.kt
+++ b/app/src/main/java/com/nextcloud/talk/receivers/DirectReplyReceiver.kt
@@ -128,9 +128,7 @@ class DirectReplyReceiver : BroadcastReceiver() {
     private fun informReplyFailed() {
         val errorColor = ForegroundColorSpan(context.resources.getColor(R.color.medium_emphasis_text, context.theme))
         val errorMessageHeader = context.resources.getString(R.string.nc_message_failed_to_send)
-        val errorMessage = SpannableStringBuilder()
-            .append("$errorMessageHeader\n$replyMessage", errorColor, 0)
-            // .append("\n$replyMessage")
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
Looping in @nimishavijay and @jancborchardt for design feedback

jancborchardt:
Hi @starypatyk, nice! Some design feedback:
- It would be good if the original message could be included. Like: "Failed to send: Bla bla blah"
- A better subline style for the error is non-italic and "medium emphasis" text color. This is more nicely readable and communicates it was not sent.
- In the case of a sending error, is it possible to add a "Resend" action right next to "Reply" in the notification?

Great work! :)

starypatyk:
Hi @jancborchardt,

Thanks for your comments. I am not sure if I understood them correctly. I have experimented with a few variants.

Error message and reply message on the same line - only error message in "medium emphasis" text colour:
obraz

Error message and reply message on separate lines - only error message in "medium emphasis" text colour:
obraz

Error message and reply message on separate lines - everything in "medium emphasis" text colour:
obraz

I have committed the last version - looks best to me.

Although the difference between "normal" and "medium emphasis" colour does not stand out - maybe due to low resolution of the emulator.

Regarding "Resend" - certainly doable, but requires more investigation/learning.

jancborchardt:
@starypatyk great stuff, good on the 3 examples and I agree with your choice! :)

I'd also say this is good to go as is, the "Resend" can be done in a separate pull request as to not block this.

jancborchardt (approved):
:+1: desgn-wise

AndyScherzinger (approved):

AndyScherzinger:
Fine with me to straight merge @mahibi

mahibi:
@thank you @starypatyk 
will merge when CI is done..

PR #2055, opened 2022-05-16
Migrate participants to kotlin

AndyScherzinger:
minor next step, the follow up has some issues I need to clarify first regarding conversations


mahibi:
MagicFirebaseMessagingService.kt: (297, 62): Type mismatch: inferred type is List<Participant>? but List<Participant> was expected

AndyScherzinger:
> MagicFirebaseMessagingService.kt: (297, 62): Type mismatch: inferred type is List? but List was expected

Argh, build flavors... 🙈

mahibi (approved):

PR #2061, opened 2022-05-17
Parcel to Parcelize Migration

AndyScherzinger:
`$subject`


.idea/codeStyles/Project.xml
```diff
@@ -59,6 +59,15 @@
...
       </option>
+      <option name="PACKAGES_IMPORT_LAYOUT">
+        <value>
+          <package name="" alias="false" withSubpackages="true" />
+          <package name="java" alias="false" withSubpackages="true" />
+          <package name="javax" alias="false" withSubpackages="true" />
+          <package name="kotlin" alias="false" withSubpackages="true" />
+          <package name="" alias="true" withSubpackages="true" />
+        </value>
+      </option>
```
AlvaroBrey, line 70: Added by IDE? This can probably be undone; imports should be already handled by ktlint.

installGitHooks.bat
```diff
@@ -0,0 +1 @@
+./gradlew installGitHooks
```
AlvaroBrey, line 1: Shouldn't this be `./gradlew.bat` instead of `./gradlew`?  Besides, do we need this `bat` file for just a single command?

.idea/codeStyles/Project.xml
AlvaroBrey, line 70: Automatically added by IDE? this should already be handled by ktlint.

AlvaroBrey (commented):
Both Detekt and Ktlint report errors

.idea/codeStyles/Project.xml
AndyScherzinger: I'll remove this one 👍

installGitHooks.bat
AndyScherzinger: you don't need the bat extension but I'll remove it. I just added it to have an easy way to install the hooks without looking up the command again

AndyScherzinger: removed! 👍

timkrueger (approved):
Thank you @AndyScherzinger

PR #2063, opened 2022-05-18
build: Fix/optimize gradle repo configuration

AlvaroBrey:
- Put jitpack and jcenter at the end of the repos list
   - This should make it so dependencies are fetched from google and maven central first, which are faster and more stable
- remove jcenter, use gradle plugin portal as mirror where needed
   - At the moment when this PR is opened, jcenter is not resolving properly, but gradle plugin portal is.
   - This is NOT a permanent solution. We should upgrade or remove jcenter dependencies, see #2064 
- Some additional cleanup

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ buildscript {
 
     repositories {
         google()
-        maven { url 'https://plugins.gradle.org/m2/' }
+        gradlePluginPortal()
         mavenCentral()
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
Follow up issue in https://github.com/nextcloud/talk-android/issues/2064 for jCenter

timkrueger (approved):

PR #2070, opened 2022-05-19
Revert AGP 7.2.0 -> 7.1.3

AndyScherzinger:
This reverts commit d3827aac9fa4d0a3b05dfd90ba5d5adc9272ef2b.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.2.0'
+        classpath 'com.android.tools.build:gradle:7.1.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${kotlinVersion}"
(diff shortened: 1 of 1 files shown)


AlvaroBrey (approved):

AlvaroBrey:
@AndyScherzinger should you also bump back gradle distribution from 7.4.2 -> 7.3.3?

AndyScherzinger:
@AlvaroBrey I hope not, but let's see, I also migrated stuff in the build script so I might have to revert even more 😭 But let's see if @timkrueger actually has a fix for the CI, then we could keep it 🤞

AlvaroBrey:
> @AlvaroBrey I hope not, but let's see, I also migrated stuff in the build script so I might have to revert even more sob But let's see if @timkrueger actually has a fix for the CI, then we could keep it crossed_fingers

I think the build script migrations are still ok for 7.1.3, since we already did them on Files, and Files is still in 7.1.3. Still, let's wait and see

AndyScherzinger:
obsolete

PR #2071, opened 2022-05-19
Use 'emulator' instead of 'emulator-headless'

timkrueger:


AlvaroBrey (approved):
Looks good theoretically, let's see if it works properly, as for Files it still won't work on gradle 7.2.0

PR #2073, opened 2022-05-19
Bump compile sdk to 31

AndyScherzinger:


AlvaroBrey:
The `Outside Range` errors can be fixed by using `getColumnIndexOrThrow` instead of `getColumnIndex`, in most cases.

Issue #1550, opened 2021-08-26
new software pattern


mahibi:
In the long term the talk app should switch to a new pattern. We want to get rid of https://github.com/bluelinelabs/Conductor 

In this issue there can be discussions which pattern should be used.
There should be the same pattern for the talk app and the files app (https://github.com/nextcloud/android).

timkrueger:
In #1903 we used a new architectural pattern for the implementation. It's based on the [Android Developer Guidelines}(https://developer.android.com/topic/architecture) and using a View Model and the Repository pattern.

AndyScherzinger:
afaik this one can be closed since a decision has been made on the architecture pattern to be used @timkrueger @mahibi

timkrueger:
> afaik this one can be closed since a decision has been made on the architecture pattern to be used @timkrueger @mahibi

You're right.

PR #2086, opened 2022-05-23
Remote browser migration

AndyScherzinger:
WIP


app/src/main/java/com/nextcloud/talk/remotefilebrowser/repositories/RemoteFileBrowserItemsRepositoryImpl.kt
```diff
@@ -0,0 +1,53 @@
...
+            )
+        val davResponse = operation.readRemotePath()
+
+        if (davResponse.getData() != null) {
+            val objectList = davResponse.getData() as List<RemoteFileBrowserItem>
+            return Observable.just(objectList)
+        }
+
+        return null
+    }
```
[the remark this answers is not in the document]
AndyScherzinger: Seems work fine, thanks @AlvaroBrey 🙏

app/src/main/java/com/nextcloud/talk/remotefilebrowser/activities/RemoteFileBrowserActivity.kt
```diff
@@ -0,0 +1,343 @@
...
+import com.nextcloud.talk.utils.database.user.UserUtils
+import com.nextcloud.talk.utils.preferences.AppPreferences
+import net.orange_box.storebox.listeners.OnPreferenceValueChangedListener
+import java.io.File
+import java.util.Collections
+import java.util.TreeSet
+import javax.inject.Inject
+
+@AutoInjector(NextcloudTalkApplication::class)
+class RemoteFileBrowserActivity : AppCompatActivity(), SelectionInterface, SwipeRefreshLayout.OnRefreshListener {
```
AlvaroBrey, line 58: This activity contains some data and logic that should be in the viewmodel (and some of it exposed through the state livedata):

- `browserItems` and sorting logic
- `selectedPaths`, `currentPath` and selection logic

In general, if the activity/fragment is holding model data, then something's wrong. The activity should strictly only transform data into views and pass events back to the viewmodel.
If you feel that this is too much data for a single State, feel free to split it (one for items/load and another for selection, for example)

app/src/main/java/com/nextcloud/talk/remotefilebrowser/repositories/RemoteFileBrowserItemsRepository.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.nextcloud.talk.remotefilebrowser.model.RemoteFileBrowserItem
+import io.reactivex.Observable
+
+interface RemoteFileBrowserItemsRepository {
+
+    fun listFolder(parameters: Parameters): Observable<List<RemoteFileBrowserItem>>
+
+    data class Parameters(
+        val path: String
+    )
```
AlvaroBrey, line 32: No need to encapsulate this, I'd just add a `path: String` parameter to `listFolder`.

app/src/main/java/com/nextcloud/talk/utils/bundle/BundleKeys.kt
```diff
@@ -73,4 +73,6 @@ object BundleKeys {
     val KEY_FORWARD_MSG_TEXT = "KEY_FORWARD_MSG_TEXT"
     val KEY_FORWARD_HIDE_SOURCE_ROOM = "KEY_FORWARD_HIDE_SOURCE_ROOM"
     val KEY_SYSTEM_NOTIFICATION_ID = "KEY_SYSTEM_NOTIFICATION_ID"
+    val KEY_SINGLE_SELECTION = "KEY_SINGLE_SELECTION"
+    val KEY_MIME_TYPE_FILTER = "KEY_MIME_TYPE_FILTER"
```
AlvaroBrey, line 77: Can be `const val`

AndyScherzinger: fixed

app/src/main/java/com/nextcloud/talk/remotefilebrowser/repositories/RemoteFileBrowserItemsRepository.kt
AndyScherzinger: fixed

app/src/main/java/com/nextcloud/talk/remotefilebrowser/activities/RemoteFileBrowserActivity.kt
AndyScherzinger: I lack the knowledge or idea(s) how to move this to the viewmodel, especially the part about selected and current path due to the fact that:
* selected path is a collection holding various, different path sections from the whole file tree while the `browserItems` collection is always just reflecting the current path, hence my struggle

as for the state

* I just adapted the state logic from the initial example from the shared items implementation not truly understanding the state concept :/

So happy for any hints and pointers 🙏 or help with the changes

app/src/main/java/com/nextcloud/talk/components/filebrowser/webdav/LegacyReadFilesystemOperation.java
```diff
@@ -0,0 +1,107 @@
...
+
+import at.bitfire.dav4jvm.DavResource;
+import at.bitfire.dav4jvm.Response;
+import at.bitfire.dav4jvm.exception.DavException;
+import kotlin.Unit;
+import kotlin.jvm.functions.Function2;
+import okhttp3.HttpUrl;
+import okhttp3.OkHttpClient;
+
+public class LegacyReadFilesystemOperation {
```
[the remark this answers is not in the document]
AndyScherzinger: I now at least added the `@Deprecated`

AndyScherzinger:
AS a sidenote anybody is welcome to join in on this PR/branch just let me know because for the time being I always ammend everything into the original commit and push that one plus very regular rebasing

app/src/main/java/com/nextcloud/talk/remotefilebrowser/activities/RemoteFileBrowserActivity.kt
AndyScherzinger: @AlvaroBrey Any idea how to do this? I simply lack the knowledge. Most examples on the internet do the sorting via the repositoy which seems strange since that mean I need to do a server round trip just to re-sort an existing result which might be okay to do it is is a database but feel very wrong for remote calls.
Also given the current implementation I have no access to the result (file list) since it is directly wrapped into the view state adding another issue: being able to change the sorting while the file listing might be empty or loading. Again any help and pointers are appreciated ❤️

AlvaroBrey: > Most examples on the internet do the sorting via the repositoy which seems strange since that mean I need to do a server round trip just to re-sort an existing result which might be okay to do it is is a database but feel very wrong for remote calls. 

Most examples use server-side sorting because they also use server-side pagination, so if sort is changed the already retrieved pages are invalid anyway. I'll take a look and see how it can be improved here.

AlvaroBrey: also done for current path, selected paths

AndyScherzinger: ❤️ Thanks a lot @AlvaroBrey - will have a look to learn how to do this myself in the future 🔮

app/src/main/java/com/nextcloud/talk/remotefilebrowser/activities/RemoteFileBrowserActivity.kt
```diff
@@ -139,11 +139,7 @@ class RemoteFileBrowserActivity : AppCompatActivity(), SelectionInterface, Swipe
                         onItemClicked(remoteFileBrowserItem)
                     }
                         .apply {
-                            items = if (remoteFileBrowserItems.size > 1) {
-                                remoteFileBrowserItems.subList(1, remoteFileBrowserItems.size)
-                            } else {
-                                ArrayList()
-                            }
+                            items = remoteFileBrowserItems
```
AndyScherzinger: this removes skipping the first item in case the list has more than one item, because the 1st item is the folder you are in and should thus be skipped. Not sure though it we shouldn't already remove it in the remote operation when return the result. What do you think @AlvaroBrey ?

AlvaroBrey: Sorry, I moved it in https://github.com/nextcloud/talk-android/pull/2086/commits/f3b5dc1df5ca8b2e769c9554693baf1cb9b5248a because I thought that  it was always root, not current folder. But yeah, it makes more sense to remove it in operation/repository. I would remove by path rather than the first element

AlvaroBrey: Pushed https://github.com/nextcloud/talk-android/pull/2086/commits/4acf71e14fe875eb3245744db5c673c2875eac92 to remove the extra filtering then.

AndyScherzinger:
However I added 1 UX improvement because that nagged me, for selection you can now click the whole line item, before you had to click on the checkbox, else no selection possible...

timkrueger (approved):
Good work @AndyScherzinger 

It is very helpful that we follow the [Recommended app architecture](https://developer.android.com/topic/architecture#recommended-app-arch) more and more. That makes reviewing much more easier 🙂

app/src/main/java/com/nextcloud/talk/remotefilebrowser/model/RemoteFileBrowserItem.kt
```diff
@@ -0,0 +1,49 @@
+/*
+ * Nextcloud Talk application
+ *
+ * @author Andy Scherzinger
+ * @author Mario Danic
+ * Copyright (C) 202 Andy Scherzinger <info@andy-scherzinger.de>
```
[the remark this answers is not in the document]
AndyScherzinger: thanks, and year is fixed via https://github.com/nextcloud/talk-android/pull/2086/commits/41694b64bb986219b2f4b8c9003d18fc4943d8eb 👍

mahibi:
played around a while and the app crashed one time while saving profile:

E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nextcloud.talk2, PID: 7906
    java.lang.NullPointerException
        at com.nextcloud.talk.controllers.ProfileController$save$1.onError(ProfileController.kt:444)
        at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.checkTerminated(ObservableObserveOn.java:281)
        at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.drainNormal(ObservableObserveOn.java:172)
        at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.run(ObservableObserveOn.java:255)
        at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:124)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

mahibi:
indeed feels good to have the new architecture :+1:
i didn't have as much time as i would have liked to review.
but while testing I noticed that the changed avatar is only displayed within the profile overview & profile settings?
In the Conversation List and in the chat the old avatar remains?

AndyScherzinger:
I'll check the log statement 👍

AndyScherzinger:
@mahibi the debug statement doesn't popup for me

AndyScherzinger:
Contributes towards #1550

app/src/main/java/com/nextcloud/talk/remotefilebrowser/viewmodels/RemoteFileBrowserItemsViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+            _fileSortOrder.value = newSortOrder
+            val currentState = viewState.value
+            if (currentState is LoadedState) {
+                val sortedItems = newSortOrder.sortCloudFiles(currentState.items)
+                _viewState.value = LoadedState(sortedItems)
+            }
+        }
+    }
+
+    fun changePath(path: String) {
```
AlvaroBrey, line 153:
```suggestion
    private fun changePath(path: String) {
```

app/src/main/java/com/nextcloud/talk/utils/FileSortOrderByName.kt
```diff
@@ -0,0 +1,57 @@
...
+        override fun compare(left: RemoteFileBrowserItem, right: RemoteFileBrowserItem): Int {
+            return if (!left.isFile && !right.isFile) {
+                return multiplier * AlphanumComparator<RemoteFileBrowserItem>()
+                    .compareRemoteFileBrowserItem(left, right)
+            } else if (!left.isFile) {
+                -1
+            } else if (!right.isFile) {
+                1
+            } else {
+                multiplier * AlphanumComparator<RemoteFileBrowserItem>().compareRemoteFileBrowserItem(left, right)
```
AlvaroBrey, line 53: This is quite inefficient as it will create a new `AlphaNumComparator` for each comparison. We've run into massive lag problems in Files due to this, when opening folders with tons of files. Same problem in line 46 above.

Instead initialize a `AlphaNumComparator` instance as a field of `RemoteFileBreowserItemNameComparator`, and use that same instance every time. 

Additionally, since you're already creating a custom `Comparator` here, there's no need to add specific methods for `RemoteFileBrowserItem` to `AlphaNumComparator`: just call `compare(String, String)` with the paths directly. Having `AlphaNumComparator` handle model classes differs from its purpose.

app/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
```diff
@@ -24,9 +24,8 @@
 
 package third_parties.daveKoeller;
 
-import com.nextcloud.talk.components.filebrowser.adapters.items.BrowserFileItem;
+import com.nextcloud.talk.remotefilebrowser.model.RemoteFileBrowserItem;
```
AlvaroBrey, line 27: See comments in `FileSortOrderByName`

AlvaroBrey (changes requested):
Looks very good generally with some minor improvements left.

Before merge, please remember to lower detekt  maxIssues (I think right now could be lowered to 88 but might change with further code edits)

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -1352,6 +1353,33 @@ class ChatController(args: Bundle) :
...
+                                .putString(KEY_ROOM_TOKEN, roomToken)
+                                .putStringArray(KEY_FILE_PATHS, paths.toTypedArray())
+                                .build()
+                            shareWorker = OneTimeWorkRequest.Builder(ShareOperationWorker::class.java)
+                                .setInputData(data)
+                                .build()
+                            WorkManager.getInstance().enqueue(shareWorker)
+                            paths = java.util.ArrayList()
+                        }
+                    }
```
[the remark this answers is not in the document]
AndyScherzinger: 😍

app/src/main/java/com/nextcloud/talk/components/filebrowser/webdav/ReadFolderListingOperation.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    private fun getModelFromResponse(response: Response, remotePath: String): RemoteFileBrowserItem {
+        val remoteFileBrowserItem = RemoteFileBrowserItem()
+        remoteFileBrowserItem.path = Uri.decode(remotePath)
+        remoteFileBrowserItem.displayName = Uri.decode(File(remotePath).name)
+        val properties = response.properties
+        for (property in properties) {
+            mapPropertyToBrowserFile(property, remoteFileBrowserItem)
+        }
+        if (remoteFileBrowserItem.permissions != null && remoteFileBrowserItem.permissions!!.contains("R")) {
```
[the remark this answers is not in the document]
AndyScherzinger: 👍

AndyScherzinger:
@AlvaroBrey fixed all your comments, thanks for the thorough review 👍

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/talk/controllers/ProfileController.kt
```diff
@@ -486,22 +480,14 @@ class ProfileController : NewBaseController(R.layout.controller_profile) {
...
-            KEY_USER_ENTITY,
-            Parcels.wrap(UserEntity::class.java, currentUser)
-        )
-        bundle.putString(KEY_ROOM_TOKEN, "123")
-        router.pushController(
-            RouterTransaction.with(BrowserForAvatarController(bundle, this))
-                .pushChangeHandler(VerticalChangeHandler())
-                .popChangeHandler(VerticalChangeHandler())
-        )
+        bundle.putString(KEY_MIME_TYPE_FILTER, "image/")
```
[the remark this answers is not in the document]
AndyScherzinger: There you go: #2140

PR #2125, opened 2022-06-03
i18n: Add a space

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>


mahibi (approved):

AndyScherzinger:
checking back with @jancborchardt if this is how we do the ellipsizing in general

Valdnet:
@AndyScherzinger To my knowledge, this is correct, which does not prevent the text strings from being unified in the same way, whether they are correct or not.
But please check with @jancborchardt.

nickvergessen:
> checking back with @jancborchardt if this is how we do the ellipsizing in general

yes it is.
Space before `…` if it's not that word being trimmed

PR #2131, opened 2022-06-07
ChatController: fix scroll to unread messages

### LINKED ISSUE #2122: Scroll to oldest unread message when openign a conversation
@AlvaroBrey since you implemented "Scroll to message if possible" for message search via #2028 maybe you could also add this one?

AlvaroBrey:
While looking at #2122 I found that the logic for scrolling to unread messages was already present,
but was broken because of the condition changed in this PR.

Unfortunately I can't understand the reason for that condition, as the history has been lost and there's no explanation in the code. During testing, I didn't find anything broken

Perhaps one of the reviewers can give input?

Fixes #2122

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
@@ -2314,7 +2314,7 @@ class ChatController(args: Bundle) :
 
                 var chatMessage: ChatMessage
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
works fine for me

AndyScherzinger:
Can't say since when this has been broken but I can't remember when this actually worked in the (recent) past. So I experienced the lack of this working for quite a while.

timkrueger (approved):

PR #2138, opened 2022-06-09
Create codeql.yml

tobiasKaminsky:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,69 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master", stable-[0-9][0-9]* ]
```
AlvaroBrey, line 16:
```suggestion
    branches: [ "master", "stable-[0-9][0-9]*" ]
```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,69 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java', 'python', 'ruby' ]
```
AlvaroBrey, line 35: We should skip jdk install and gradle build for languages other than java. Can be done with a `if:` on setup and assemble steps.

AlvaroBrey (changes requested):

.github/workflows/codeql.yml
tobiasKaminsky: Do we need python / ruby at all?
I removed it in library…

AndyScherzinger:
@tobiasKaminsky can we put this into the template repo so it can also be updated from a central place? :)

AlvaroBrey:
> @tobiasKaminsky can we put this into the template repo so it can also be updated from a central place? :)

https://github.com/nextcloud/android-common/pull/1 <- need to create a github PAT (from nextcloud-android-bot for example) and add it as a secret to the common repo. I can't as I don't have admin permissions on that repo.

AndyScherzinger:
> I can't as I don't have admin permissions on that repo.

@AlvaroBrey now you do 😃

PR #2154, opened 2022-06-19
Add room database layer

### LINKED ISSUE #1549: replace Requery with Room
We need to get rid of requery for the following reasons:
- blocks Android Studio update (JDK11) https://github.com/nextcloud/talk-android/issues/1125
- end of life. see https://github.com/requery/requery/issues/907
- [x] add Room #2154
- [x] replace existing requery calls with room

requery should be replaced by Room (https://developer.android.com/training/data-storage/room)
- medium article about migration from requery to Room: https://medium.com/@markchristopherng/using-room-our-real-life-experience-b1e4875cb2f1

Left-over is #2244 to completely remove requery.

AndyScherzinger:
Adding `room` as a new db layer framework including
- db migration to move to `room`-supported column data types
- db layer implementation with `dao`, `repository` interface/implementation
- business layer `UserManager` as a business component in contrast to the old `UserUtils`
- as a first set of examples migrate `MainActivity` and `SettingsController` to use the new components
- wire components via `dagger`, same as wiring has been done before
- use `Maybe<>` instead on `Observable<>` for now so `disposable` is handled implicitly by using `Maybe`


app/src/main/java/com/nextcloud/talk/users/UserManager.kt
```diff
@@ -0,0 +1,433 @@
...
+    private fun updateUserData(user: UserNgEntity, userAttributes: UserAttributes) {
+        updateUserIdIfNeeded(userAttributes, user)
+        updateTokenIfNeeded(userAttributes, user)
+        updateDisplayNameIfNeeded(userAttributes, user)
+        updatePushConfigurationStateIfNeeded(userAttributes, user)
+        updateCapabilitiesIfNeeded(userAttributes, user)
+        updateCertificateAliasIfNeeded(userAttributes, user)
+        updateExternalSignalingServerIfNeeded(userAttributes, user)
+        updateCurrentUserStatusIfNeeded(userAttributes, user)
+    }
```
AlvaroBrey, line 319: Besides the fact that this should take `User` instead of  the entity, all the `ifNeeded` methods seem superfluous and inefficient. Just do an `updateUser(user: User)` method and update always. If the data is the same as before then there will be no changes :shrug:

app/src/main/java/com/nextcloud/talk/users/UserManager.kt
```diff
@@ -0,0 +1,433 @@
...
+
+    private fun updateUserIdIfNeeded(userAttributes: UserAttributes, user: UserNgEntity) {
+        if (userAttributes.userId != null && (user.userId == null || user.userId != userAttributes.userId)) {
+            user.userId = userAttributes.userId
+        }
+    }
+
+    private fun createUser(username: String?, userAttributes: UserAttributes): UserNgEntity {
+        val user = UserNgEntity()
```
AlvaroBrey, line 381: This should be hidden in the Repository. Only `User` class should be used here.

app/src/main/java/com/nextcloud/talk/utils/DisplayUtils.java
```diff
@@ -593,6 +599,34 @@ public static void loadAvatarImage(UserEntity user, SimpleDraweeView avatarImage
         avatarImageView.setController(draweeController);
     }
 
+    public static void loadAvatarImage(UserNgEntity user, SimpleDraweeView avatarImageView, boolean deleteCache) {
```
AlvaroBrey, line 602: Should use `User` rather than entity

app/src/main/java/com/nextcloud/talk/utils/database/user/CapabilitiesNgUtil.java
```diff
@@ -0,0 +1,230 @@
...
+
+import com.nextcloud.talk.data.user.model.UserNgEntity;
+import com.nextcloud.talk.models.json.capabilities.Capabilities;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import androidx.annotation.Nullable;
+
+public abstract class CapabilitiesNgUtil {
```
AlvaroBrey, line 32: This class should use `User` model class rather than  entity, if possible. Also should be Kotlin since it's a new class.

app/src/main/java/com/nextcloud/talk/data/user/model/UserNgEntity.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.nextcloud.talk.models.json.push.PushConfigurationState
+import com.nextcloud.talk.utils.ApiUtils
+import kotlinx.android.parcel.Parcelize
+import kotlinx.serialization.Serializable
+import java.lang.Boolean.FALSE
+
+@Parcelize
+@Serializable
+@Entity(tableName = "User")
+data class UserNgEntity(
```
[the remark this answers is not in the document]
AndyScherzinger: TO have an easy way to distinguish from `UserEntity` (the old, generated requery class, but I am fine with renaming, worst case we see full-qualified class names used, while I guess we shouldn't if I fix your other comments.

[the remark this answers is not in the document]
AlvaroBrey: I get that, I just don't understand what `Ng` means, why not `New` or something like that?

app/src/main/java/com/nextcloud/talk/dagger/modules/DatabaseModule.java
```diff
@@ -44,7 +46,13 @@ public class DatabaseModule {
...
     @Provides
     @Singleton
-    public SqlCipherDatabaseSource provideSqlCipherDatabaseSource(@NonNull final Context context) {
+    public SqlCipherDatabaseSource provideSqlCipherDatabaseSource(
+        @NonNull final Context context,
+        final AppPreferences appPreferences) {
+        int version = DB_VERSION;
+        if (appPreferences.getIsDbRoomMigrated()) {
+            version++;
+        }
```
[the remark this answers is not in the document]
AndyScherzinger: this is needed due to the upgrade vs. new install scenario you faced. So initially it needs to be version 7 based on requery, then Room updates it to version 8, then on any future start of the app requery when loading the database needs to also request version 8. But if requery does v8 than the db will be at v8 and Room's upgrade won't run because the Db is at v8 but then Room will crash because the structure hasn't been migrated.

And yes, this is a one-way street with very strict road: 
**no db changes before having removed all uses of requery to access the database**

After that we should keep the framework and db initialization in for a bit, since requery did the upgrades to v7, unfortunately we would probably need to spent some deep digging to find out the differences between the version. If we have them, then it is easier since then Room could do all the migrations.

So my idea was, to wait until we have everything on Room, then wait for older versions to have very little to no usage in the gplay reports, then throw out requery completely.

app/src/main/java/com/nextcloud/talk/data/user/model/UserNgEntity.kt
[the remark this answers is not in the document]
AndyScherzinger: Ng = next/new generation, but I'll change that 👍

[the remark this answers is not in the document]
AndyScherzinger: Renamed to UserEntity

AndyScherzinger:
@AlvaroBrey I need your help again, sorry :/

The adapted integration test fails
```
com.nextcloud.talk.activities.MainActivityTest > login[android(AVD) - 8.1.0] FAILED
	at org.junit.Assert.fail(Assert.java:88)
```

Which translates to the following code not working (scenario for a new, non-existing user)
https://github.com/nextcloud/talk-android/blob/a25767073e5ffb4d0f7354769012e55f479f7b93/app/src/main/java/com/nextcloud/talk/users/UserManager.kt#L132-L152

However, if I don't implement a flow but do it like this, then it works:
```
val existingUser = userObservable.blockingGet()
val userModel = when (existingUser) {
    null -> createUser(
        username,
        userAttributes
    )
    else -> {
        updateUserData(
            existingUser,
            userAttributes
        )
        existingUser
    }
}
val id = userRepository.insertUser(userModel)
val user = userRepository.getUserWithId(id).blockingGet()
```
My guess would be this being due to some reactive specific thing I am not aware of (yet)

app/src/main/java/com/nextcloud/talk/users/UserManager.kt
AndyScherzinger: In between (like yesterday) for once I experience NPEs due to the token being null for existing users that have been used before and lived on the device for some time. I haven't plugged in the storing of users via room yet, so I have no clue what might have happened killing of the tokens in the database...

app/src/main/java/com/nextcloud/talk/utils/database/user/CapabilitiesUtilNew.java
```diff
@@ -0,0 +1,229 @@
+/*
```
AlvaroBrey, line 1: No new files in Java! :coffee: 

Additionally the same note about deprecation duplication applies here. Change methods in the old CapabilityUtils to call the ones in this one after converting the `UserEntity` to a `User`.

app/src/main/java/com/nextcloud/talk/utils/database/user/UserUtils.java
```diff
@@ -35,6 +35,7 @@
 import io.requery.query.Result;
 import io.requery.reactivex.ReactiveEntityStore;
 
+@Deprecated
```
AlvaroBrey, line 38:
```suggestion
/**
 * @deprecated use {@link com.nextcloud.talk.users.UserManager} instead.
 */
@Deprecated
```

AlvaroBrey (changes requested):

app/src/main/java/com/nextcloud/talk/data/user/UsersDao.kt
```diff
@@ -0,0 +1,155 @@
...
+                // unused atm
+            }
+
+            override fun onError(e: Throwable) {
+                Log.e(TAG, "Error setting user active", e)
+                result = FALSE
+            }
+        })
+
+        return result
```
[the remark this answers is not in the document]
AndyScherzinger: Since RX2 doesn't have a blocking subscribe, how would I do this? `blockingGet()` + catching an Exception for the error scenario and take it from there?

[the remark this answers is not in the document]
AndyScherzinger: Like I did on markUserForDeletion ?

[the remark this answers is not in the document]
AlvaroBrey: Either do everything synchronously, or drop the `suspend` (which you should do anyway), return a `Single<Boolean>` and use `map` to map the subscription to the result.

app/src/main/java/com/nextcloud/talk/utils/database/user/CapabilitiesUtilNew.java
AndyScherzinger: That way you migrate all call-path to the methods in the old utils class at once which means you change executed code throughout the whole app, which implies testing everything. If you work with a adapted copy and migrate it piece by piece you can then remove the old methods when all references have been gone. This way you have more, fine-grained control over what you change, so I would suggest not do this.

app/src/main/java/com/nextcloud/talk/models/json/capabilities/Capabilities.kt
```diff
@@ -25,9 +25,11 @@ import android.os.Parcelable
 import com.bluelinelabs.logansquare.annotation.JsonField
 import com.bluelinelabs.logansquare.annotation.JsonObject
 import kotlinx.android.parcel.Parcelize
+import kotlinx.serialization.Serializable
 
 @Parcelize
 @JsonObject
+@Serializable
```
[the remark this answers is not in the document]
AndyScherzinger: tested and removed, also for the other relevant json classes

app/src/main/java/com/nextcloud/talk/utils/database/user/CapabilitiesUtilNew.java
AlvaroBrey: > If you work with a adapted copy and migrate it piece by piece you can then remove the old methods when all references have been gone.

Fair enough, but with the current approach we run into risk of inconsistencies if we forget to modify both classes at the same time. Your call

app/src/main/java/com/nextcloud/talk/data/user/UsersDao.kt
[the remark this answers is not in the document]
AlvaroBrey: Done https://github.com/nextcloud/talk-android/pull/2154/commits/5aa05a29705f827319f37bae0fad48e7a69a778a

AlvaroBrey (approved):
Great work!

app/src/main/java/com/nextcloud/talk/data/user/model/UserEntity.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.nextcloud.talk.models.json.capabilities.Capabilities
+import com.nextcloud.talk.models.json.push.PushConfigurationState
+import kotlinx.android.parcel.Parcelize
+import java.lang.Boolean.FALSE
+
+@Parcelize
+@Entity(tableName = "User")
+data class UserEntity(
+    @PrimaryKey(autoGenerate = true) @ColumnInfo(name = "id") var id: Long = 0,
+    @ColumnInfo(name = "userId") var userId: String? = null,
```
mahibi, line 39: is there a reason why @ColumnInfo is used? It could be omitted when the var name is the same...

mahibi: ok, as i read it's recommended this way to have more flexibility to rename the members without having to rename the DB column..

AndyScherzinger: Yes, that is why, because some of the variables might not be as nicely named, so I might change them in the future, but wouldn't migrate the database column names just for that :)

mahibi (approved):

mahibi:
tested a lot, everything fine :+1: 

Thank you very much @AndyScherzinger & @AlvaroBrey :1st_place_medal:

AndyScherzinger:
Thanks a lot for collaborating, reviews and testing  @mahibi @AlvaroBrey

PR #2165, opened 2022-06-24
Improvements to analysis scripts and Drone configuration

timkrueger:
- Replace 'getBranchName.sh' with Git command
- Add stable branches as trigger for Drone
- Detect stable branches on the fly

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,6 +11,7 @@ steps:
 trigger:
   branch:
     - master
+    - stable-*
   event:
     - pull_request
     - push
@@ -29,6 +30,7 @@ steps:
 trigger:
   branch:
     - master
+    - stable-*
   event:
     - push
     - pull_request
@@ -69,6 +71,7 @@ services:
 trigger:
   branch:
     - master
+    - stable-*
   event:
     - push
     - pull_request
@@ -124,6 +127,7 @@ steps:
 trigger:
   branch:
     - master
+    - stable-*
   event:
     - push
     - pull_request
--- a/.drone.yml
+++ b/.drone.yml
@@ -91,7 +91,7 @@ steps:
       LOG_PASSWORD:
         from_secret: LOG_PASSWORD
     commands:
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -91,7 +94,7 @@ steps:
       LOG_PASSWORD:
         from_secret: LOG_PASSWORD
     commands:
-      - export BRANCH=$(scripts/analysis/getBranchName.sh $GIT_USERNAME $GIT_TOKEN $DRONE_PULL_REQUEST)
```
tobiasKaminsky, line 94: This script retrieves the base branch of this PR, not the current branch name.
Naming of the script is a bit misleading.

tobiasKaminsky (changes requested):
getBranchName is needed

.drone.yml
timkrueger: > This script retrieves the base branch of this PR, not the current branch name. Naming of the script is a bit misleading.

For me it looks like it's not he base/parent branch.

The following command is used in the deleted script `getBranchName.sh`:

```
curl 2>/dev/null -u $1:$2 https://api.github.com/repos/nextcloud/talk-android/pulls/$3 | grep \"ref\": | grep -v master | cut -d"\"" -f4
```
The result for this PR:
```shell
$ curl 2>/dev/null https://api.github.com/repos/nextcloud/talk-android/pulls/2165 | grep \"ref\": | grep -v master | cut -d"\"" -f4
chore/noid/script-work
```

Also in the afterwards the usage of the base/parent branch makes no sense in my view. For example pushing to the parent/base branch in `findbugs-up.rb`:

https://github.com/nextcloud/talk-android/blob/chore/noid/script-work/scripts/analysis/findbugs-up.rb#L132

tobiasKaminsky (approved):

.drone.yml
tobiasKaminsky: I understood it wrong :)

PR #2168, opened 2022-06-27
Migrate requery entity use to room entity

AndyScherzinger:
* Growing from the root: started migration of `ConversationListController` and migrated any uses until app compiled again
* Migrate `SettingsController`


app/src/main/java/com/nextcloud/talk/jobs/CapabilitiesWorker.java
```diff
@@ -67,44 +71,45 @@ public class CapabilitiesWorker extends Worker {
...
+                user -> eventBus.post(new EventStatus(user.getId(),
+                                                      EventStatus.EventType.CAPABILITIES_FETCH,
+                                                      true)),
+                throwable -> {
+                    eventBus.post(new EventStatus(internalUserEntity.getId(),
+                                                  EventStatus.EventType.CAPABILITIES_FETCH,
+                                                  false));
+                    dispose(dbQueryDisposable);
+                },
+                () -> dispose(dbQueryDisposable));
```
AndyScherzinger, line 103: @AlvaroBrey any idea/hint what I need to do to make this a synchronous call, currently everything throughout the app is `rxjava2` which doesn't offer a `blockingSubscribe` :(

AlvaroBrey: Any reason why you want to make it synchronous at all? I don't see any need

AndyScherzinger: While not having looked into manually testing it, I would think that it won't work async

AndyScherzinger:
Obsolete

PR #2173, opened 2022-06-28
NewBaseController: Inject kotlin fields without @JvmField and nullable type

AlvaroBrey:
These do not need to be nullable as they will be injected immediately. This will make child controllers code much cleaner.

I've chosen not to fix the now unneeded `!!` and `??` in all the child controllers as it would be too big a diff and could easily cause conflicts.
Instead, I suggest we clean those up as we work on Controllers in the future.


AndyScherzinger (approved):

AndyScherzinger:
👍 I also saw it but didn't take the time yet to do a PR 👍

PR #2176, opened 2022-06-29
Introduce DisposableSet to handle multiple disposables at once

AlvaroBrey:
This should be used in places where multiple Disposables are handled simultaneously as a simple way of
ensuring that all disposables are cleared on exit. For example on ViewModels and Controllers should likely use one
of these and clear it on their respective lifecycle end.

I've included an example use in ChatController, where it replaces an existing `ArrayList<Disposable>`, but the benefit
should be more visible in places where Disposables are currently handled separately.


AndyScherzinger (approved):
Nice improvement 🚀

PR #2199, opened 2022-07-07
Mark as read via notification

### LINKED ISSUE #2164: "Mark as read" via notification
@starypatyk implemented "Quick reply via notification" what works very well
Mark as read would be awesome, too.

AndyScherzinger:
Resolves #2164


AndyScherzinger:
@nickvergessen the push notification "only" ships the `DecryptedPushMessage` https://github.com/nextcloud/talk-android/blob/feature/noid/markAsRead/app/src/main/java/com/nextcloud/talk/models/json/push/DecryptedPushMessage.kt which has an idea (room token) and a notification id but doesn't provide the message id thus from a notification we can't mark a message as read since we don't know the message id. Any way we could ship that via the notification(s)? Else I fear we can't implement the feature.

starypatyk:
Hi @AndyScherzinger,

I have checked quickly that the notification handling code does indeed [pull the notification from the server](https://github.com/nextcloud/talk-android/blob/6e2ee9853614df9b1e1219c02b6902a6bc43c88b/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.java#L190), as suggested by @nickvergessen. So, with some small adaptations, marking the message as read seems implementable.

I can pick your work from the branch and continue with the implementation over the next few days. It finally seems that I will be able to devote some time to the project... :wink:

Or would you prefer to complete this yourself?

AndyScherzinger:
@starypatyk if you can pick up this PR that be awesome since I likely won't have the time, so that would be awesome 👍

AndyScherzinger:
❗ rebased ❗

AndyScherzinger:
@starypatyk nice polishing and reducing spotbugs is always welcomed 😊

There might be some followup work for future PRs 😁

* migrate to kotlin
* migrate from UserUtils to UserManager

While I am perfectly fine postponing that to another PR.

@AlvaroBrey @mahibi ready for review 😊

Super nice collaboration @starypatyk ✨

PR #2205, opened 2022-07-11
fix crash when changing status during call

### LINKED ISSUE #2204: crash while setting status during call
## Steps to reproduce
1. be in a call on web
2. with the same user, change status in android talk app while being in the call

### Expected behaviour
no crash, status can be edited.

### Actual behaviour
app crashes.

`The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | java.util.NoSuchElementException: Collection contains no element matching the predicate.`

mahibi:
fix #2204

"call" is not contained in predefinedStatusesList so it crashed when searching for it with predefinedStatusesList.first()


AndyScherzinger (approved):
LGTM, just a cosmetic thing

PR #2084, opened 2022-05-20
🎨 Material 3️⃣

AndyScherzinger:
Implementing **Material 3** on top of server theming support

Resolves #2076

Requires / is builds on top of #2217

In a future follow-up PR I am going to restructure the theming code to be better separated between highlevel theming functions and the coloring of UI elements / parts of UI elements.

|Old|New Light|New Dark|
|---|---|---|
|conversationlist_light_old|conversationlist_light|conversationlist_dark|
|chat_light_old|chat_light|chat_dark|
|reactions_light_old|reactions_light|reaction_dark|
|accountChooser_light_old|accountChooser_light|accountChooser_dark|
|setStatus_light_old|setStatus_light|setStatus_dark|
|profileEdit_light_old|edit_profile_light|profile_edit_dark|

Signed-off-by: Andy Scherzinger <info@andy-scherzinger.de>

PR #2217, opened 2022-07-14
🎨 Server theming

### LINKED ISSUE #250: 🎨 Support Server theme
As a company owner I would like to have the theme configured at the server visible in the mobile app.

Maybe related to #207

AlvaroBrey:
Fixes #250 

- [x] fix crash on initial, clean app launch when no user exists
- [x] fallback to xml provided colors in case no theme information from server available
- [x] ~~grey theming for settings dialogs radio buttons (see files app for reference)~~ Do this _after_ moving SettingsController to an Activity implementation

Chat controller WIP:
- [x] Send button
- [x] Outgoing messages
   - [x] Normal outgoing messages
   - [x] File icons (folder and archive)
   - [x] Audio
   - [x] location links in outgoing shared locations
   
- [x] **CREATE `stable-14.2` BEFORE MERGING TO `master`**

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -79,6 +79,7 @@ android {
             disable 'InvalidPackage'
             disable 'MissingTranslation'
             disable 'VectorPath'
+            disable 'UnusedQuantity'
         }
 
         javaCompileOptions {
--- a/app/src/main/java/com/nextcloud/talk/activities/TakePhotoActivity.java
+++ b/app/src/main/java/com/nextcloud/talk/activities/TakePhotoActivity.java
@@ -41,8 +41,10 @@ import android.widget.Toast;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.nextcloud.talk.R;
+import com.nextcloud.talk.application.NextcloudTalkApplication;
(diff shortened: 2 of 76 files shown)


AndyScherzinger (approved):
Nice work @AlvaroBrey 👍

AndyScherzinger:
Follow-up issue for improvements is #2255

AndyScherzinger:
👍 fine by me of course and heavily tested/in use via #2084 work

@nickvergessen @jancborchardt what do you think? If we merge it we would ship it with 14.2.0 in August while with 15.0.0 we would enhance/replace it with #2084 bringing a Material 3 implementation of theming, so the UI changes twice in a short timeframe. I'd be fine with it since theming is already in place for iOS, so you can consider this one feature-parity, while Material 3 will then just the the next evolutionary step on top and comes in a relatively short timeframe after but that shouldn't be an issue.

timkrueger (approved):
I tested on all my devices and in the emulator. 

Very nice work @AlvaroBrey and @AndyScherzinger . Also the nice little improvements here and there :heart_eyes:

timkrueger:
stable-14.2 is created.

I don´t merge this because the drone job is still running. Requested an abort.

Issue #2218, opened 2022-07-14
switching accounts fails / wrong username shown in account switcher


mahibi:
## Steps to reproduce
there exist 2 users: userA and userB

1. login with userA
2. go to account switcher
3. BUG: avatars are shown correctly, but userA has the name of userB
4. keep userA selected. go to some conversation and send a message
5. BUG: message is sent as userB
6. go back to account switcher and select userB
7. go to the conversation and send a message
8. BUG: message is sent as userA

this doesn't always happen but sometimes.

### Expected behaviour
- names in account switcher match the avatars
- actions are not executed as another user

i guess this is caused by the room migration..

## Device information
**Talk version:** master (>= 14.2.0 Alpha 02)

## Server information
sermo

AndyScherzinger:
@mahibi are both users on the same server or different ones? I tested with 3 users on 3 servers and don't have an issue there.

mahibi:
both on the same server (sermo)

mahibi:
just reproduced this with 14.1.0 RC2 !
this time the names in the account switch dialog were correct but the messages were still send as the wrong user!

AndyScherzinger:
@mahibi but it works with 14.0.2 - no?

mahibi:
same error with talk app 14.0.2 on sermo. so either i proved it's a bug present in the app or (which i assume as described) on sermo... now trying with older servers..

AndyScherzinger:
same error with 14.0.0 with personal server on 24.0.2 while it works fine with 13.0.0 so it is not the DB technology switch but something else.

AndyScherzinger:
--> issue must be client side

mahibi:
as expected, and now approved:
this happens only for accounts that are on the same server. So this should not affect too many users, anyway it's a serious bug.

AndyScherzinger:
Some thinking and digging broad me to testdriving cookie clearing, see #2229 which leads to the correct expected results whenever to clear all cookies _before_ the next server call, seems the server's response resets the cookie state to a "wrong" one, so without clearing all follow-up server calls are wrong again.

mahibi:
@jld3103 was able to reproduce this once! 
So we tried again but now fail to reprduce it another time...

Talk app version 15.1.0 Alpha 03
tested with 2 accounts on the same localhost

mahibi:
never experienced this again. closing..

PR #2222, opened 2022-07-16
Migrate further areas to room

AndyScherzinger:
* settings
* profile
* switching accounts

I keep synchronous process synchronous (i.e. always calling blockingGet) because the current focus is to get rid of any requery interactions with the database especially since the controllers aren't here to be kept nor is their individual implementation build for async atm.


AndyScherzinger:
@mahibi I couldn't reproduce the account switching issue from #2218 with my latest changes on this branch maybe because this update to room now ships with a set active user method that wraps all changes into a single transaction.

mahibi:
> @mahibi I couldn't reproduce the account switching issue from #2218 with my latest changes on this branch maybe because this update to room now ships with a set active user method that wraps all changes into a single transaction.

this doesn't fix the issue, reproduced it with this branch :(

AndyScherzinger:
> this doesn't fix the issue, reproduced it with this branch :(

😢 then I was just lucky to not have hit the issue I guess. 
Anyway any review and 👍 is highly appreciated, so I can work on the follow-up PRs more easily :)

mahibi:
> > this doesn't fix the issue, reproduced it with this branch :(
> 
> cry then I was just lucky to not have hit the issue I guess. Anyway any review and +1 is highly appreciated, so I can work on the follow-up PRs more easily :)

i will take another look regarding the issue. i think the issue appeared when writing a message but the conversation list before was correct if i remember correctly. So there might just be more parts that need the `.blockingGet()` in order to fix ALL scenarios..?

AndyScherzinger:
Thanks. I can't say for sure, since the behavior we have seen was rather strange because it seems that some async operation would mix-up the user-context which wasn't on DB level since restarting the app showed the correct data, just any further calls went wrong which makes me believe it is either something in a worker happening in the background or some cache issue in a library like okhttp

mahibi (approved):

app/src/main/java/com/nextcloud/talk/users/UserManager.kt
```diff
@@ -123,6 +123,18 @@ class UserManager internal constructor(private val userRepository: UsersReposito
         }.toSingle()
     }
 
+    fun updateUser(user: User): Single<Int> {
+        return if (user.id != null) {
+            Single.just(userRepository.updateUser(user))
+        } else {
+            Single.just(userRepository.insertUser(user).toInt())
+        }
```
AlvaroBrey, line 131:
```suggestion
        return Single.fromCallable {
            when (user.id) {
                null -> userRepository.insertUser(user).toInt()
                else -> userRepository.updateUser(user)
            }
        }
```
Simplified code a bit.

I'd prefer if we defined a `Single` version of `updateUser/insertUser` and just return a call to that, but this is fine for now.

AlvaroBrey (approved):

PR #2223, opened 2022-07-17
Migrate remaining controllers from requery to room

AndyScherzinger:
This PR migrates the remaining controllers from requery to room

* server selection
* web view login
* account verification
* location picker
* cleanup ChatController & GeocodingController

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.kt
@@ -367,7 +367,7 @@ class WebViewLoginController(args: Bundle? = null) : NewBaseController(
                     if (currentUser != null) {
                         currentUser.clientCertificate = appPreferences!!.temporaryClientCertAlias
                         currentUser.token = loginData.token
-                        val rowsUpdated = userManager.updateUser(currentUser).blockingGet()
+                        val rowsUpdated = userManager.updateOrCreateUser(currentUser).blockingGet()
                         Log.d(TAG, "User rows updated: $rowsUpdated")
 
                         if (finalMessageType != null) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/nextcloud/talk/controllers/WebViewLoginController.kt
```diff
@@ -327,15 +325,16 @@ class WebViewLoginController(args: Bundle? = null) : NewBaseController(
         val loginData = parseLoginData(assembledPrefix, dataString)
         if (loginData != null) {
             dispose()
-            val currentUser = userUtils.currentUser
+            val currentUser = userManager.currentUser.blockingGet()
             var messageType: ApplicationWideMessageHolder.MessageType? = null
-            if (!isPasswordUpdate && userUtils.getIfUserWithUsernameAndServer(loginData.username, baseUrl)) {
+            if (!isPasswordUpdate &&
+                userManager.getIfUserWithUsernameAndServer(loginData.username!!, baseUrl!!).blockingGet()
```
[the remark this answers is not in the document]
AlvaroBrey: If it _always_ needs to emit a value (`true` or `false` in this case), then it should not be a `Maybe`, but a `Single` instead.
Here's how to do that:

```kotlin
    fun getIfUserWithUsernameAndServer(username: String, server: String): Single<Boolean> {
        return userRepository
            .getUserWithUsernameAndServer(username, server)
            .map { true }
            .switchIfEmpty(Single.just(false))
    }
```

AndyScherzinger:
@AlvaroBrey currentUser being null has been fixed with my latest commit, in case you have some time to spare :) I did test adding 1 and 2 users to a fresh installation and that worked as well as user switching

app/src/main/java/com/nextcloud/talk/utils/AccountUtils.kt
```diff
@@ -1,7 +1,9 @@
 /*
  * Nextcloud Talk application
  *
+ * @author Andy Scherzinger
  * @author Mario Danic
+ * Copyright (C) 2021 Andy Scherzinger <info@andy-scherzinger.de>
```
AlvaroBrey, line 6: Wrong year!

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/talk/utils/AccountUtils.kt
AndyScherzinger: Fixed :)

PR #2232, opened 2022-07-21
Push token recovery should not use the standard httpClient object 

### LINKED ISSUE #2218: switching accounts fails / wrong username shown in account switcher

AndyScherzinger:
...and its cookieStore

Resolves #2218

Issue had been introduced by #1776 

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/PushRegistrationWorker.java
+++ b/app/src/main/java/com/nextcloud/talk/jobs/PushRegistrationWorker.java
@@ -1,10 +1,12 @@
 /*
  * Nextcloud Talk application
  *
- * @author Mario Danic
+ * @author Andy Scherzinger
  * @author Marcel Hibbe
- * Copyright (C) 2017 Mario Danic <mario@lovelyhq.com>
+ * @author Mario Danic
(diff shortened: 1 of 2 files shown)


mahibi:
i tested a lot and can't reproduce the bugs :+1:
Well done @AndyScherzinger  :1st_place_medal: :+1:

mahibi (approved):

AndyScherzinger:
Thanks @mahibi - tested the weeklies between 13.0.0 and 14.0.0 💪 until I hit the faulty behavior, then the PRs between, luckily it was the first PR after a weekly 🎉

PR #2267, opened 2022-08-03
Fix status visibility in conversation list

tobiasKaminsky:
Status was not hidden on group/chats without status, thus old status was shown.
(I am unsure if changed hashCode is needed, but FlexibleAdapter checks in some way if item has changed and thus to be sure that a changed status will result in an updated view, I added it.)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
@@ -108,7 +108,9 @@ public class ConversationItem extends AbstractFlexibleItem<ConversationItem.Conv
 
     @Override
     public int hashCode() {
-        return conversation.hashCode() * (status == null ? 1 : status.hashCode());
+        int result = conversation.hashCode();
+        result = 31 * result + (status != null ? status.hashCode() : 0);
+        return result;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
tested successfully

app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
```diff
@@ -108,7 +108,7 @@ public Conversation getModel() {
 
     @Override
     public int hashCode() {
-        return conversation.hashCode();
+        return conversation.hashCode() * (status == null ? 1 : status.hashCode());
```
AlvaroBrey, line 111: Multiplying both hashcodes can result in overflow. 

Here's how the autogenerator does it instead: multiply one of them by a prime number and sum them

```suggestion
        int result = conversation.hashCode();
        result = 31 * result + (status != null ? status.hashCode() : 0);
        return result;
```

AlvaroBrey (changes requested):

app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
AndyScherzinger: 👍

AndyScherzinger:
@AlvaroBrey committed your suggestion 👍

AlvaroBrey (approved):

PR #2274, opened 2022-08-04
Make equals null-safe in case no status is set

AndyScherzinger:
follow-up for #2267 since status can be null hence running into a NPE which kills the "unread notification" popup bubble.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
@@ -51,6 +51,7 @@ import com.nextcloud.talk.utils.DisplayUtils;
 import com.nextcloud.talk.utils.database.user.CapabilitiesUtilNew;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 import androidx.core.content.ContextCompat;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.java
```diff
@@ -97,7 +97,9 @@ public ConversationItem(Conversation conversation, User user,
     public boolean equals(Object o) {
         if (o instanceof ConversationItem) {
             ConversationItem inItem = (ConversationItem) o;
-            return conversation.equals(inItem.getModel()) && status.equals(inItem.status);
+            return conversation.equals(inItem.getModel()) &&
+                (status == null && inItem.status == null) ||
+                (status != null && status.equals(inItem.status));
```
AlvaroBrey, line 102: You can use `Objects.equals(status, inItem.status)` instead, it's null safe. I can't add it as a suggestion because it requires the `Objects` import

AndyScherzinger: @AlvaroBrey DONE and pushed 👍 feel free to re-review/approve :)

AlvaroBrey (approved):

PR #2285, opened 2022-08-10
ChatController: fix crash when opening conversation from notification

### LINKED ISSUE #2262: New message notification don't open the conversation anymore
Happens on master / 14.2.0 Alphas

AlvaroBrey:
This happens because there is no Conversation object available in the Controller when coming from a notification. Unfortunately it's needed for the polls viewholders, as it's used for the `isOwnerOrAdmin` attribute.

This patch means that when opening a notification for a poll, it will always look like the user is not the owner of the poll even if they are.
This is unfortunately difficult to fix as the initialization logic for ChatController is all kinds of messed up: ideally we'd always load the Conversation object before showing any kind of UI (other than the loading UI) but that's too big of a change to do without tests.

Fixes #2262

- [ ]  create `stable-14.2` _after_ merging this PR


timkrueger (approved):

timkrueger:
I increase the allowed lint warnings, because on the master are currently also 112.

AndyScherzinger (approved):

PR #2294, opened 2022-08-12
Use issue form for bug report

AlvaroBrey:
- Replace bug template with an issue form
- Add forum link to the issue choose page

Based on https://github.com/nextcloud/android/pull/10528 and adapted for Talk.

Test run on https://github.com/AlvaroBrey/issue-test/issues/new/choose . Feel free to submit issues to check the results.


timkrueger:
I like that really and that would improve the quality of reports which we get. But I'm a bit worried that there will be less reported because so many details are required. We have to ask ourselves who the target group is:

- Regular users (but do they always have a GitHub account?).
- Experienced users
- Developers

What's the opinion of @AndyScherzinger, @nickvergessen or @mahibi?

nickvergessen:
Yeah same concern.

It's okay to have the fields but from my POV they need to be optional and nullable...
E.g. I myself find the form on the server https://github.com/nextcloud/server/issues/new?assignees=&labels=bug%2C0.+Needs+triage&template=BUG_REPORT.yml&title=%5BBug%5D%3A+ so annoying that I always take the feature form and wipe all content.

Also you e.g. require info the users might not have access to 
> ### Nextcloud server version
> See admin overview page: /index.php/settings/admin/overview

Same issue with turn/stun/signaling. 

Things like:
>  I've written a clear and descriptive title for this issue, not just "Bug" or "Crash". *
> This is a bug, not a question or an enhancement. *

should not be a checkbox. Add some hint text as a first step instead. Also instead of negative samples, tell examples which are useful/positive
Also well if they report a feature request as a bug you can just change the label? No need to annoy with another click on a checkbox and a line of text to read.

When people reached the issue tracker they already made a lot of extra steps for us to help improve the situation they are facing. it should be easy to off-load their stuff and not be blocked by a 20 input form

AlvaroBrey:
> Yeah same concern.
> 
> It's okay to have the fields but from my POV they need to be optional and nullable... 

I've updated the fields to make everything optional except repro steps, expected and actual behaviour, and device information.

> E.g. I myself find the form on the server https://github.com/nextcloud/server/issues/new?assignees=&labels=bug%2C0.+Needs+triage&template=BUG_REPORT.yml&title=%5BBug%5D%3A+ so annoying that I always take the feature form and wipe all content.

As a tip, you can directly go to https://github.com/nextcloud/server/issues/new (without going through the UI) to create an empty one, even if that's disabled in the UI.

> Also you e.g. require info the users might not have access to
> 
> > ### Nextcloud server version
> > See admin overview page: /index.php/settings/admin/overview
> 
> Same issue with turn/stun/signaling.

This was  in the old template so I just copied it over. I have now made it optional. Also for this specific field I've directed users to a way to see the version without admin privileges.

> Things like:
> 
> > I've written a clear and descriptive title for this issue, not just "Bug" or "Crash". *
> > This is a bug, not a question or an enhancement. *
> 
> should not be a checkbox. Add some hint text as a first step instead. Also instead of negative samples, tell examples which are useful/positive Also well if they report a feature request as a bug you can just change the label? No need to annoy with another click on a checkbox and a line of text to read.

Changed all the textboxes to non-actionable bullet points. Removed the "this is actually a bug" one, and gave an example of a good title.

> When people reached the issue tracker they already made a lot of extra steps for us to help improve the situation they are facing. it should be easy to off-load their stuff and not be blocked by a 20 input form

There needs to be a balance between ease to report and ease to maintain; I wasted way too much time in the Files app closing empty or useless reports before creating the form. Obviously that is not the case here due to the smaller userbase, but still. Either way I agree that there was too much mandatory stuff in this one.

Updated version can, again, be tested in github.com/AlvaroBrey/issue-test/issues/new/choose

AlvaroBrey:
> I like that really and that would improve the quality of reports which we get. But I'm a bit worried that there will be less reported because so many details are required. We have to ask ourselves who the target group is:
> 
>     * Regular users (but do they always have a GitHub account?).
>     * Experienced users
>     * Developers

Generally I'd say the target group should be "experienced users". You don't need to be a developer to create an account and report but it is harder than just sending an email; however, less experienced users will produce lesser quality of reports (and we get crash/ANR reports on Play Console anyway).

nickvergessen:
Fine by me

AndyScherzinger:
👍 merging then

PR #2296, opened 2022-08-12
Fix some easy detekt issues and detekt.yml syntax

AlvaroBrey:
Detekt number was wrong on master, so instead of bumping it up, I just fixed some easy issues.

Also detekt was printing out some warnings about the YML syntax which I also fixed.


timkrueger (approved):
I also started fixing them. You win the race 🐢

AndyScherzinger (approved):
Nice one!

Issue #2052, opened 2022-05-16
RFC: Options to maximize field of view for video calls


ASerbinski:
There are three aspects of camera configuration that result in cropped video output and a substantially reduced field of view;

1) Use of 16:9 aspect ratio instead of the sensor native 4:3 https://github.com/nextcloud/talk-android/blob/a43a51b931687c7a3059b2fd5ee1ee30dd828923/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1745
2) Use of Electronic Image Stabilization (EIS)
3) Inability to control zoom ratio results in fixed ratio of 1:1.

The aspect ratio will affect devices universally, since as far as I can tell, all camera sensors seem to have a 4:3 native aspect ratio. The other two will apply on a sensor by sensor basis. Some phones have a mixture of available zoom ratios, and support for EIS or OIS (where OIS is available, EIS is disabled).

Naturally, there will be tradeoffs for each of the three causes of cropping;
1) User may have a preference for a 16:9 ratio.
2) Holding the device in user's hand will result in a jittery video without stabilization.
3) Changing the zoom ratio could cause the HAL to select a different sensor with different characteristics, for example, the main camera on a Pixel 6 [/Pro] has OIS, but reducing the zoom level to lower than 1:1 will cause it to switch to the wide angle lens, which does not have OIS.

Consequently, each of these configuration changes should be user-controlled.

Access to the camera on this software is through webrtc, which unfortunately, does not provide means of user control for EIS or zoom ratio, however, others have coerced it into allowing settings; https://stackoverflow.com/questions/48598752/android-how-to-turn-on-flashlight-using-webrtc-android-io-pristine-libjingle1

ASerbinski:
I've implemented a proof of concept here; https://github.com/ASerbinski/talk-android/commit/049428597ba8ee42811498462917fca03a913fc8

Sets 4:3 aspect ratio (1280x960) at activities/CallActivity.java:1745
Adds custom implementations of webrtc classes Camera2Enumerator, CameraSession, and Camera2Capturer where in ExtCameraSession (CameraSession), zoom ratio is set to 0.5 at line 186, and EIS is switched off at line 210.

Note: added files implement proof of concept solutions for https://github.com/nextcloud/talk-android/issues/2051 and https://github.com/nextcloud/talk-android/issues/2017

With this proof of concept, instead of user's face filling up the entire preview screen, now fills about 1/9th of the preview screen.

[...]

PR #2297, opened 2022-08-12
Maximize FOV on video call

ASerbinski:
The field of view on video calls is very narrow, resulting in a transmitted video that is highly "zoomed in".

This increases the field of view using three approaches;
1) Use 4:3 video aspect ratio to prevent cropping to 16:9
2) Disable EIS (removes the buffer around the video)
3) Zoom out

The aspect ratio will affect devices universally, since as far as I can tell, all camera sensors seem to have a 4:3 native aspect ratio. The other two will apply on a sensor by sensor basis. Some phones have a mixture of available zoom ratios, and support for EIS or OIS (where OIS is available, EIS is disabled).

Naturally, there will be tradeoffs for each of the three causes of cropping;

User may have a preference for a 16:9 ratio.
Holding the device in user's hand will result in a jittery video without stabilization.
Changing the zoom ratio could cause the HAL to select a different sensor with different characteristics, for example, the main camera on a Pixel 6 [/Pro] has OIS, but reducing the zoom level to lower than 1:1 will cause it to switch to the wide angle lens, which does not have OIS.
Consequently, it may be preferable for each of these configuration changes to be user-controlled.

Access to the camera on talk-android is through webrtc, which unfortunately, does not provide means of user control for EIS or zoom ratio, however, others have coerced it into allowing settings; https://stackoverflow.com/questions/48598752/android-how-to-turn-on-flashlight-using-webrtc-android-io-pristine-libjingle1
Unfortunately, this approach is a little bit hacky since it requires duplication of some of webrtc.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -156,6 +156,7 @@ import org.webrtc.CameraVideoCapturer.CameraSwitchHandler
 import org.webrtc.DefaultVideoDecoderFactory
 import org.webrtc.DefaultVideoEncoderFactory
 import org.webrtc.EglBase
+import org.webrtc.ExtCamera2Enumerator
 import org.webrtc.Logging
 import org.webrtc.MediaConstraints
 import org.webrtc.MediaStream
@@ -231,6 +232,9 @@ class CallActivity : CallBaseActivity() {
     private val cameraSwitchHandler = Handler()
 
     private val callTimeHandler = Handler(Looper.getMainLooper())
+
+    private var disableEIS = true
+    private var zoomOut = true
 
     // push to talk
     private var isPushToTalkActive = false
@@ -708,7 +712,9 @@ class CallActivity : CallBaseActivity() {
         } catch (t: Throwable) {
             Log.w(TAG, "Camera2Enumerator threw an error", t)
         }
-        cameraEnumerator = if (camera2EnumeratorIsSupported) {
+        cameraEnumerator = if (camera2EnumeratorIsSupported && (disableEIS || zoomOut)){
+            ExtCamera2Enumerator(this, disableEIS, zoomOut)
+        } else if (camera2EnumeratorIsSupported) {
             Camera2Enumerator(this)
         } else {
             Camera1Enumerator(WebRTCUtils.shouldEnableVideoHardwareAcceleration())
@@ -2071,7 +2077,7 @@ class CallActivity : CallBaseActivity() {
 
     private fun startVideoCapture() {
         if (videoCapturer != null) {
-            videoCapturer!!.startCapture(1280, 720, 30)
+            videoCapturer!!.startCapture(1280, 960, 30)
         }
     }
 
--- a/app/src/main/java/org/webrtc/ExtCamera2Capturer.java
+++ b/app/src/main/java/org/webrtc/ExtCamera2Capturer.java
@@ -1,0 +1,42 @@
+package org.webrtc;
+
+import android.content.Context;
+import android.hardware.camera2.CameraManager;
+
+import org.jetbrains.annotations.Nullable;
+
+public class ExtCamera2Capturer extends Camera2Capturer {
+
+    @Nullable private final CameraManager cameraManager;
+    private final boolean disableEIS, zoomOut;
+
+    public ExtCamera2Capturer(Context context, String cameraName, CameraEventsHandler eventsHandler,
+                              boolean disableEIS, boolean zoomOut) {
+        super(context, cameraName, eventsHandler);
+        cameraManager = (CameraManager) context.getSystemService(Context.CAMERA_SERVICE);
+        this.disableEIS = disableEIS;
+        this.zoomOut = zoomOut;
+    }
+
+    @Override
+    protected void createCameraSession(CameraSession.CreateSessionCallback createSessionCallback,
+                                       CameraSession.Events events, Context applicationContext,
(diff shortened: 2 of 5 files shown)


timkrueger:
I don't  saw this before writing my [comment](https://github.com/nextcloud/talk-android/issues/2052#issuecomment-12132367020) . I'll a look next week.

andyxheli:
Hi @timkrueger did you have time to review this ?

timkrueger:
> Hi @timkrueger did you have time to review this ?

I also liked to have it. But as I wrote in the corresponding issue https://github.com/nextcloud/talk-android/issues/2052#issuecomment-1232930509 the is work to do.

ASerbinski:
There is a bug that causes the preview to be cropped to the screen ratio rather than showing the full preview, introduced in this commit; https://github.com/nextcloud/talk-android/commit/892ffe87dd45a61e8e84769af3f18f4e89ea08e2

andyxheli:
Happy New Year everyone! Just wanted to see if there's any progress on this? It would sure be nice to have this. Not sure if this has anything to do with it but for me the images are always blurry compared to iOS talk

AndyXheli:
Hi @timkrueger just wanted to see if there's any progress on this ? It would be really nice to have this. 😃

AndyXheli:
Hi all, any status on getting this merged ? @mahibi anyway you can you review this when you have time i don't think we've heard from @timkrueger in a while

AndyXheli:
Hey all! Any updates on this?.

AndyXheli:
Hello, any status when will this be available

ASerbinski:
> Hello, any status when will this be available

Hi @AndyXheli ; I wish I could report on a timeline, but I'm stuck at this point without any guidance on how to proceed. I've been waiting for feedback on how to proceed since 2022.

AndyXheli:
Hey thanks @ASerbinski  I think @timkrueger is no longer part of nextcloud. @mahibi would someone be able to assist with this? Any feedback would be great. As we currently stand the video on the talk app is not very good and this looks like this PR would help resolve that

mahibi:
So sorry for the long delay!! Unfortunately this one slipped through when i handled all open community PR's a few weeks ago 🙈
As this is a bigger topic i will bring it up for planning!

mahibi:
I'm pinging @nickvergessen @AndyScherzinger @nimishavijay @jancborchardt 
because we need some decisions regarding where we want to go with this. 
This is about the impacts if we want to have a higher Field of View (FOV). 
Please also have a look at the associated issue https://github.com/nextcloud/talk-android/issues/2052
For more infos on OIS and EIS, this video may be helpful: https://www.youtube.com/watch?v=igNO2F21dA4
As @ASerbinski  pointed out, there are tradeoffs because of technical limitations. 

I try to summarize this:

| **Current state used ever since**                    | **Pro**                                                               | **Con**                                                        |
|-------------------------------|-----------------------------------------------------------------------|----------------------------------------------------------------|
| **16:9 Ratio**                 | Many users expect 16:9, especially on smartphones.                    | Crops sides or top/bottom depending on orientation.           |
| **Electronic Image Stabilization (EIS)** | Stabilizes video even without Optical Image Stabilization (OIS).    | Crops the video to maintain stabilization.                   |
| **Zoom 1:1**                   | No lens switching, maintaining OIS if present.                        | Reduces field of view (FoV) by not zooming out.              |

While this table looks like 16:9, EIS and 1:1 zoom are the better choice, keep in mind that all of them reduce the FoV.
So using 4:3, no EIS and zooming out would create a higher FOV.

So the question is:
1) Do we want to keep the video calls like they are?
vs
2) do we want to change them to have a higher FoV? (approach of this PR)
vs 
3) do we want to have options for the user for it? (but this could overwhelm users due to the complexity as the settings would be very technical)

If we would go for the changes to have a higher FoV, this would mean more complexity as @ASerbinski  pointed out in the PR while users might complain about the downsides. 
So we should have a decision beforehand and try to avoid to put more efforts into this (from @ASerbinski  or anyone involved) and afterwards realize that the majority prefers the old solution.

ASerbinski:
Thank you @mahibi 

One thing I would like to point out is that I did my implementation thus far with the anticipation that different users may have different preferences, therefore setting the variables disableEIS and zoomOut determine which of these are enabled, as well as manually setting the capture resolution. Having these controllable via settings UI would probably be a good way to go.

The main concerns that I had in my implementation is that I had to copypaste a few hundred lines of WebRTC in order to override EIS and set the zoom. The actual changes I made to WebRTC are very small, so copying that much code feels like a dirty hack.

jancborchardt:
@mahibi @nickvergessen @nimishavijay @ASerbinski I would say a primary factor is (as so often): How do other products do it?
In this case especially cause if you are in a Zoom call or Discord call and then change to Nextcloud Talk, and then your FOV changes (and thus you need to move your webcam), that would be highly annoying. So we should be consistent with others here.


And to answer your 3rd question @mahibi:

> do we want to have options for the user for it? (but this could overwhelm users due to the complexity as the settings would be very technical)

As usual we should only introduce settings when absolutely necessary. A setting for FOV is definitely much too technical and should not be necessary to adjust.

mahibi:
Hi @ASerbinski 
there was now the decision from Netxcloud: we don't want any extra settings. 
To improve the FOV situation anyway, we could do an automatic switch depending on the orientation, so it would be 4:3 in portrait mode and 16:9 in landscape mode.
I created the PR https://github.com/nextcloud/talk-android/pull/4838 for this. Feel free to comment on it.
Other than that, we want to keep it simple and don't deal with the other options that affect the FOV.
Because of this i would like to close this PR, hope this is fine for you?
And another time: Sorry that this PR got forgotten too often. This does not meet my own or Nextcloud's requirements and won't happen again.

PR #2307, opened 2022-08-16
handle expiring messages

### LINKED ISSUE #2281: ⏳ Handle expiring messages
Server ref https://github.com/nextcloud/spreed/issues/7359

mahibi:
resolve #2281

Conversation settings | Conversation settings dark | Expiring messages dialog 
------- | ------- |-------- 
grafik | grafik | grafik | 

Tasks:
- [x] Make sure items are removed from the view when their message expiration is reached even when the user doesn't leave the conversation
- [x] Make sure when entering a conversation we don't show messages that expired in the meantime
- [x] Handle system message

Follow-up possible:
- [x] Add an option to set the expiration with:
    - Off
    - 4 weeks
    - 1 week
    - 1 day
    - 8 hours
    - 1 hour

# remaining TODOS:
- [x] fix klint/detect
- [x] remove some hardcodings?

# Notes
- Handling of expiring messages must be implemented in a cleaner way once offline support is done with the new architecture. (right now data is modified on the adapter itself which is not a good practice).

- Setting time for expiring messages must be implemented in a cleaner way once dependency `com.yarolegovich:mp` is replaced with plain android solutions.

### DIFF
--- a/app/src/main/res/layout/controller_conversation_info.xml
+++ b/app/src/main/res/layout/controller_conversation_info.xml
@@ -3,6 +3,8 @@
   ~
   ~ @author Mario Danic
   ~ @author Andy Scherzinger
+  ~ @author Marcel Hibbe
+  ~ Copyright (C) 2022 Marcel Hibbe <dev@mhibbe.de>
   ~ Copyright (C) 2021 Andy Scherzinger <info@andy-scherzinger.de>
   ~ Copyright (C) 2017-2018 Mario Danic <mario@lovelyhq.com>
   ~
@@ -45,9 +47,10 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content">
 
-        <RelativeLayout
+        <LinearLayout
             android:layout_width="match_parent"
-            android:layout_height="wrap_content">
+            android:layout_height="wrap_content"
+            android:orientation="vertical">
 
             <com.yarolegovich.mp.MaterialPreferenceCategory
                 android:id="@+id/conversation_info_name"
@@ -87,7 +90,6 @@
                 android:id="@+id/conversation_description"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_below="@id/conversation_info_name"
                 android:animateLayoutChanges="true"
                 android:visibility="gone"
                 apc:cardBackgroundColor="@color/bg_default"
@@ -109,7 +111,6 @@
                 android:id="@+id/otherRoomOptions"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:layout_below="@id/conversation_description"
                 android:visibility="gone"
                 apc:cardBackgroundColor="@color/bg_default"
                 apc:cardElevation="0dp"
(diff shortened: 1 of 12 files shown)


mahibi:
this PR might be paused because api is not 100% ready yet

app/src/main/res/values/strings.xml
```diff
@@ -564,4 +564,16 @@
     <string name="call_without_notification">Call without notification</string>
     <string name="set_avatar_from_camera">Set avatar from camera</string>
 
+    <string name="nc_conversation_settings">Conversation settings</string>
+
+    <!-- Expiring messages -->
+    <string name="nc_expire_messages">Expiring messages</string>
```
nickvergessen, line 570:
```suggestion
    <string name="nc_expire_messages">Expire chat messages</string>
```

Also if possible can we copy the explanation from the web UI part, to explain what happens to shared files, etc?

app/src/main/res/layout/controller_conversation_info.xml
```diff
@@ -3,9 +3,11 @@
   ~
   ~ @author Mario Danic
   ~ @author Andy Scherzinger
+  ~ @author Marcel Hibbe
+  ~ Copyright (C) 2022 Marcel Hibbe <dev@mhibbe.de>
   ~ Copyright (C) 2021 Andy Scherzinger <info@andy-scherzinger.de>
   ~ Copyright (C) 2017-2018 Mario Danic <mario@lovelyhq.com>
-  ~
+
```
AndyScherzinger, line 10: would revert this one

app/src/main/res/values-v28/arrays.xml
```diff
@@ -39,6 +39,24 @@
...
+    </array>
+
+    <array name="message_expiring_values">
+        <item>expire_0</item>
+        <item>expire_2419200</item>
+        <item>expire_604800</item>
+        <item>expire_86400</item>
+        <item>expire_28800</item>
+        <item>expire_3600</item>
+    </array>
```
AndyScherzinger, line 58: these values won't be available on any pre v28 Android version and I guess crash, so you need to put them into `/values/` I guess if that file is available, else create it. Sidenote not for this PR, arrays.xml seems to have a lot of duplicated for "all" and "v28+", so some housekeeping to do in the future I think.

app/src/main/res/values/arrays.xml
```diff
@@ -39,6 +39,24 @@
...
+    </array>
+
+    <array name="message_expiring_values">
+        <item>expire_0</item>
+        <item>expire_2419200</item>
+        <item>expire_604800</item>
+        <item>expire_86400</item>
+        <item>expire_28800</item>
+        <item>expire_3600</item>
+    </array>
```
AndyScherzinger, line 58: see above I would expect you only need the values in here.

AndyScherzinger:
Just some minor comments from my side.

app/src/main/res/values-v28/arrays.xml
mahibi: that's true. deleted them here and tested with v27 device. everything works.

app/src/main/res/values/arrays.xml
mahibi: yep

app/src/main/res/values/strings.xml
nickvergessen: As discussed, I would like to see it, if it looks out of place lets skip it for now.
Not worth looking into a new lib atm.

mahibi: this would be the dirty solution with a normal text field under the setting:

grafik

:+1:  or :-1:  @nickvergessen ?

AndyScherzinger: @mahibi the new header line theme to not be tinted correctly, seems to be a lighter shade of blue (best to test on sermo if it turns orange)

nickvergessen: I think that's good enough?

timkrueger (approved):

TalhaAbbas55:
why this option is not visible in Android app now in v17

AndyScherzinger:
You can only set it per room being an admin for that room, so if you don't see it on the web you won't see it on Android either.

If you see it on web it should show on Android also.

PR #2310, opened 2022-08-17
Remove lovely dialogs uses

AndyScherzinger:
* cleanup settings controller
* migrate any use to Material dialogs in conversation list controller


app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
```diff
@@ -104,7 +105,6 @@
 import com.nextcloud.talk.utils.rx.SearchViewObservable;
 import com.nextcloud.ui.popupbubble.PopupBubble;
 import com.yarolegovich.lovelydialog.LovelySaveStateHandler;
```
AlvaroBrey, line 107: The `lovelySaveStateHandler` field (and this import) is not needed anymore here.

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java
AndyScherzinger: Eagle eyes! Removed it now, thanks! 🎉

PR #2313, opened 2022-08-18
Convert ConversationsListController to Kotlin, and get rid of ButterKnife

AlvaroBrey:
- ConversationsListController: convert to Kotlin
- ConversationListController: use NewBaseController + viewBinding
- ConversationsListController: clean up a bit
- Get rid of BaseController, ButterknifeController, and ButterKnife
- Rename NewBaseController to just BaseController
- Bump detekt.yml issues after ConversationsListController conversion

Check each commit individually for easier review


AndyScherzinger (approved):
Awesome work @AlvaroBrey 🎉

PR #2282, opened 2022-08-09
Fix branch detection for drone PRs

nickvergessen:
Partly reverting #2165 

* When the code summary is reduced, the merged drone state was pushed to `$BRANCH`
* Since #2165 that variable is always master or stable-
* https://drone.nextcloud.com/nextcloud/talk-android/4879/4/2 see lines 605
```
[master a7a56a64] Drone: update Lint results to reflect reduced error/warning count [skip ci]
1 file changed, 1 insertion(+), 1 deletion(-)
To https://github.com/******/talk-android
   fa91ba46..a7a56a64  HEAD -&gt; master
```
* As per "commit" the branch is `master` (first wird in square bracket)
* and was then pushed to master

PR #2350, opened 2022-09-02
Restore executable permissions for getBranchName.sh

AlvaroBrey:
When `getBranchName.sh` was restored in #2282, the executable permission was missing from the new file. Together with a lack of error handling in the scripts, this meant that `BRANCH` was just an empty string, thus shifting all following arguments one position to the left when calling `analysis-wrapper`.  The most noticeable consequence was that uploading lint/spotbugs files was failing.

This PR adds back the executable permission and fixes that.


tobiasKaminsky (approved):

tobiasKaminsky:
Woho! Spotbugs, etc. is finally up again!

PR #2355, opened 2022-09-06
Fix warnings in ChatController

timkrueger:
- Remove unnecessary safe calls and non-null assertions
- Remove unnecessary elvis operator (?:)
- Remove assingments from 'if' clauses
- Convert string concatination to template
- Simplify boolean expressions
- Fix some spelling issues
- Replace 'Integer#toString' with Kotlin functions
- Use indexing insteat of 'Map#get'
- Make properties private if possible
- Remove unused parameter 'timeout' from 'ChatController#processMessages'


mahibi (approved):

timkrueger:
@mahibi Know you can review again. Massive changes since last time.

mahibi (approved):

PR #2364, opened 2022-09-13
Improve chat bubble sizes

### LINKED ISSUE #2362: Make Reader check marks smaller
It would be nice to have the read check mark smaller. Something similar to  Whatsapp 

image

AndyScherzinger:
Resolves #2362 

|Before|After|
|---|---|
|before|after|

**Changes**
* lowered the border radius a bit
* lowered the check-mark size to 18dp
* lowered the top/bottom padding in the bubbles
* made the "sharp corner" slightly rounded

**Sidenote**
* `18dp` is the min-size for icons to be used in general


AndyScherzinger:
What do you think @nimishavijay ?

Also cc @mahibi

mahibi (approved):

nimishavijay (approved):
Looks much better now! Possible follow up: make the border radius smaller for consecutive messages so that it looks more continuous (similar to Signal)

AndyScherzinger:
> Possible follow up: make the border radius smaller for consecutive messages so that it looks more continuous (similar to Signal) 

Put it into a follow-up issue since that logic isn't trivial to implement unfortunately :(

PR #2367, opened 2022-09-14
Respect participant permissions

timkrueger:
This PR adds support for the following permissions:

- [x] 2 Start call
- [x] 4 Join call
- [x] 8 Can ignore lobby
- [x] 16 Can publish audio stream
- [x] 32 Can publish video stream
- [ ] ~64 Can publish screen sharing stream~ _Screen sharing is not supported in NC Talk for Android._
- [x] 128 Can post chat message, share items and do reactions

__Limitation: The permissions don't get updated during a call. Only a reload of the conversation will get permission updates!__


timkrueger:
The implementation of the publish permissions will wait for changes of @danxuliu to avoid massive conflicts in the `CallActivity`.

nickvergessen:
Can this be merged in the meantime and a follow up be used for the remaining permissions?

timkrueger:
> Can this be merged in the meantime and a follow up be used for the remaining permissions?

The blocking changes already in the master. So I implement the missing three today.

timkrueger:
> Limitation: The permissions don't get updated during a call. Only a reload of the conversation will get permission updates!

Can we live with that limitation @nickvergessen ? I'm also not happy with that. But before make the permissions update-able on the fly I want to make a bigger refactoring of the 'CallActivity'. I plan that to do for 15.1!

nickvergessen:
> Can we live with that limitation @nickvergessen ? I'm also not happy with that. But before make the permissions update-able on the fly I want to make a bigger refactoring of the 'CallActivity'. I plan that to do for 15.1!

Respecting any permission is a step forward, so yeah let's do it.

mahibi (approved):

PR #2406, opened 2022-09-19
Determine base branch with 'jq'

timkrueger:
The GitHub API response contains the base branch in which the changes should
pulled into. This is now extracted with 'jq' from the response.

Also some Shellcheck issues were fixed.


tobiasKaminsky (approved):

tobiasKaminsky:
Nice. We should also move this to android-config somehow…

timkrueger:
> Nice. We should also move this to android-config somehow…

I created an issue in android-config:
- https://github.com/nextcloud/android-config/issues/17

Issue #1981, opened 2022-05-04
Forward is not cleared after forwarding


nickvergessen:
## Steps to reproduce
1. Have a message/text in another app that allows forwarding
2. Forward to talk android
3. Go to a conversation and post it.
4. Go back to the overview
5. Click on another unread chat

### Expected behaviour
Cleared behaviour

### Actual behaviour
Input is filled with the same text again?

Not sure if this is expected, for me it's unexpected every time it happens.

nickvergessen:
![threema-20220511-163752655.jpg](https://user-images.githubusercontent.com/213943/167876464-353fbcc3-9370-491e-937b-7dc7f6c1c66e.jpg)

As you can see i just forwarded it to you, i press back amd it shows the forwarding headline again

mahibi:
aaah you mean the appbar header. so the error you describe is in step 4...

nickvergessen:
Yeah and when from there you go to any other app it forwards the details again

mahibi:
i guess in short term the line

`showShareToScreen = !showShareToScreen && hasActivityActionSendIntent();`

in https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java#L360

might cause the bug and could be changed.

But the best solution for this would be that we create a seperate view instead of reusing the ConversationsListController for sharing/forwarding actions. But this would take a bit longer...

nickvergessen:
Okay, anything but a quickfix os out of scope. So let's just keep the ticket in the backlog if anyone ever wants to do it.

I thought maybe its as simple as clearing a variable on posting

mahibi:
> i guess in short term the line
> 
> `showShareToScreen = !showShareToScreen && hasActivityActionSendIntent();`
> 
> in https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java#L360
> 
> might cause the bug and could be changed.

might be caused by commit https://github.com/nextcloud/talk-android/commit/cf471e336f9f0a9623b224e75f551c190f2cf5a9

Issue #1353, opened 2021-06-23
🤝 Account chooser for "share to"


mahibi:
when selecting "share to Nextcloud Talk" there should be an account chooser as a first screen if user is logged in with multiple accounts

jakobroehrl:
In the file app, the user can tip within the sharing screen on the 3 dot menu and choose there "change account" if needed:
![image](https://user-images.githubusercontent.com/24530680/123062191-933ee200-d40c-11eb-8a4b-f0a95f8b8648.png)

mahibi:
> Could this maybe get a bit prioritized? It makes using the app really hard (when sharing pictures) with multiple accounts, imho. :/

moved to planning evaluation

szaimen:
> moved to planning evaluation

Thanks!

mahibi:
also trying to fix https://github.com/nextcloud/talk-android/issues/1981 if possible

szaimen:
🎉🎉🎉

PR #2413, opened 2022-09-21
add account switcher for "share to"

mahibi:
resolve https://github.com/nextcloud/talk-android/issues/1353
fix https://github.com/nextcloud/talk-android/issues/1981

single account | multiple accounts
-------- | -------- 
grafik | grafik


mahibi:
one thing i recognized:
when a translation will be too long, icons may disappear:

grafik

app:showAsAction="collapseActionView|always" doesn't seem to help here.
need to read how this can be avoided. 
but this may be a problem in general also for other appbar design in the app, so might not block to merge this PR..

timkrueger (approved):

Issue #1724, opened 2021-11-23
show notifications for missed calls (+ improve duration for ringing)


mahibi:
https://github.com/nextcloud/talk-android/pull/1634 fixed that the phone rings only for 1 minute and not forever. Now there must be an info for the user when the phone stops to ring that a call was missed!

# for 1:1 calls 
there should be even no timeout but the phone should ring as long as the other participant tries to call. And after the other participant hung up the notification should switch to a lower importance and set the text "you missed a call in ....".

# for group calls:
if the group call is still ongoing after it was not joined for 1 minute, is suggest to switch to a lower notification importance. Means no sound, and only an icon visible in the status bar. 
But it should still check if it's ongoing from time to time (like a 2-5 minutes interval) to decide if the text says "ongoing call in group ...." or "you missed a call in ...."

the phone should only continue to ring if the user never joined on any device.
What should happen if he joined on an other device? No notification at all..?

mahibi:
@Ivansss is the issue description like it works on iOS or are there differences?

Ivansss:
On iOS the phone rings for 45 secs for any type of call.
During that time it checks if the user has joined the call from another device.
After the 45 secs it shows a "missed call" notification.

PR #2415, opened 2022-09-21
Feature/1724/missed call notification

mahibi:
merge blocked by https://github.com/nextcloud/talk-android/pull/2449

### this PR will:
- resolve https://github.com/nextcloud/talk-android/issues/1724
   - image
- convert some classes from java to kotlin
- introduce refactoring of Notification handling (isolate firebase stuff from other logic)
   - All "UI-notification" logic from `ChatAndCallMessagingService` was moved to `NotificationWorker`. `ChatAndCallMessagingService` was renamed to `NCFirebaseMessagingService` because it is now only responsible for firebase stuff. This separation should make it easier for alternative push services to dock with the app (if they are incorporated in the future).

### Notes
- please note that for the next release, minSDK should be at least 23 in order to support the method `notificationManager.activeNotifications`. This method is already in use (which might introduced problems for older phones in the past) but it is further used with this PR.
   - Bumping the minSDK will happen in https://github.com/nextcloud/talk-android/pull/2449
- this PR should not improve or worsen the reliability of notification delivery. However i expect the following impact of the "you missed a call" notifications on phones that are already known to have problems with on-time delivery:
  - if there were problems with on-time delivery and it starts to work again (e.g. after a phone restart), the user might get a bunch of too late delivered notifications at once (this might already happened for text messages, but now also for missed calls).

### possible follow up PR's:
- `NotificationWorker` still needs to be further simplified!


mahibi:
detekt checks are not fully satisfied. Some of the checks fail with ComplexMethod. While this is true and should be resolved in general, i suggest to review/merge anyway because this PR already changed a lot and seems stable for me. Further refactoring might take place in follow up PR's (after my vacation or later).

feel free to improve minor things in this PR while reviewing @timkrueger

timkrueger (changes requested):
The new notifications works as expected. But for the debug information I got a negative delay:



I'll try to look into it today or tomorrow.

Regarding the detekt issues I'm OK with merging them and resolve them afterwards in a separate PR.

mahibi:
> The new notifications works as expected. But for the debug information I got a negative delay:

i reverted the calculation of the delivery time. This doesn't make sense because time is not synchronized... in the future we may be able to trigger a test push from the device itself and measure the full roudtrip time until it is received on the device again. But calculating a difference to a timestamp from firebease server doesn't make sense because it's not time-synchronized and never will be.

timkrueger (approved):

Issue #2426, opened 2022-09-26
broken UI for polls buttons on small screens


mahibi:
### Steps to reproduce

open poll with moderator rights so 3 buttons (end poll, dismiss, vote) are visible

### Expected behaviour

all button fully visible

### Actual behaviour

![grafik](https://user-images.githubusercontent.com/2932790/192247080-871ca954-f70c-4bbb-a849-3f3b2e465277.png)

### Device brand and model

random

### Android version

random

### Nextcloud Talk app version

15.1.0 Alpha 03

PR #2430, opened 2022-09-27
change poll buttons to vertical orientation

mahibi:
fix https://github.com/nextcloud/talk-android/issues/2426

before (on small screens) | after
-------- | -------- 
grafik | grafik

### DIFF
--- a/app/src/main/res/layout/dialog_poll_results.xml
+++ b/app/src/main/res/layout/dialog_poll_results.xml
@@ -42,25 +42,26 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="@dimen/standard_margin"
-        android:gravity="end">
-
-        <com.google.android.material.button.MaterialButton
-            android:id="@+id/poll_results_end_poll_button"
-            style="@style/Button.Borderless"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:layout_marginEnd="@dimen/standard_margin"
-            android:minHeight="@dimen/min_size_clickable_area"
-            android:text="@string/polls_end_poll"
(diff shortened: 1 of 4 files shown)


timkrueger (approved):

AndyScherzinger:
Already merged, but could we "end-align" them @mahibi ?

mahibi:
> Already merged, but could we "end-align" them @mahibi ?

i agree right alignment would be nicer.
just had a look at https://m3.material.io/components/dialogs/guidelines

there are the advice:
- "Dialogs should contain a maximum of two actions."
- "Stacked buttons" should be used with caution

So i suggest we could hide the "End poll" button while editing the own vote (it would still be visible when you not yet voted at all, as well for the intermediate result screen), so there should be never 3 buttons for the polls dialogs. 
So then the 2 buttons ("Dismiss"&"Vote" or "End poll"&"Vote" or "End poll"&"Edit vote") might even be okay to be aligned horizontally next to each other.

what do you think @AndyScherzinger ?

mahibi:
it also seems that buttons in dialogs don't have borders that often in the m3 examples.

I'm especially unsure about the "End poll" button. Not sure if it looks too harmless without border for such an impactful action (although there is a confirm dialog when clicked)?

there is also a difference between iOS and web:

android:
grafik

iOS:
grafik

web:
grafik

Any decision/advice how this should be common for all platforms? @nickvergessen @jancborchardt @nimishavijay

nickvergessen:
We use the error button as it is quite critical

mahibi:
> We use the error button as it is quite critical

@Ivansss

AndyScherzinger:
m3 doesn't have error buttons

@mahibi just apply the primary, outlined button theming from our new theming implementation on that button, it'll apply the grey border 👍

Issue #2432, opened 2022-09-27
Reactions are not shown


neunzehnachtneun:
### Steps to reproduce

1. Update from 14.2.0 to 15.0.0 RC2
2. In any chat react to a message
3. See, that the reaction is not shown.

### Expected behaviour

Reactions to messages should be visible below the message.

### Actual behaviour

Reactions aren't shown. Via webclient they are visible, while they aren't in talk. 

This issue affects both sent and received messages as well as sent and received reactions.

### Device brand and model

Pixel 6a

### Android version

13

### Nextcloud Talk app version

15.0.0 RC2

mahibi:
thanks for the report @neunzehnachtneun !
indeed this is broken somehow, we will have a look!

PR #2437, opened 2022-09-28
fix to show reactionsEmojiWrapper

mahibi:
fix https://github.com/nextcloud/talk-android/issues/2432

with https://github.com/nextcloud/talk-android/pull/2364 it was implemented to hide the reactionsEmojiWrapper when no emojis are set for a message. whenever a emoji was added, it was actually not shown because the wrapper was still hidden.

with the fix, the wrapper is made visible again


AndyScherzinger (approved):

AndyScherzinger:
Works fine 👍 
Sorry for introducing this regression :/

Issue #2346, opened 2022-09-01
Show info dialog to user if loading of conversations fail on startup


mahibi:
If loading the conversations fails on startup, the user only sees a blank screen (because there is no offline support yet) but has no idea what's going wrong.

There should be a dialog in case something goes wrong.

Furthermore:
- log error (but don't flood logfile in future...)
- show option to re authenticate
- show option to remove user?
- show hint what else could be checked

mahibi:
fixed with https://github.com/nextcloud/talk-android/pull/2453

PR #2453, opened 2022-09-30
Show error when loading chats fails

jld3103:
Partially addresses https://github.com/nextcloud/talk-android/issues/2346, but only shows that an error happened and doesn't offer any options.
Screenshot_1664552840

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.kt
+++ b/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.kt
@@ -616,6 +616,7 @@ class ConversationsListController(bundle: Bundle) :
                 handleHttpExceptions(throwable)
                 withNullableControllerViewBinding {
                     binding.swipeRefreshLayoutView.isRefreshing = false
+                    showErrorDialog()
                 }
                 dispose(roomsQueryDisposable)
             }) {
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
Looks good code-wise 👍

UI-wise I tend to suggest a different approach to display this by either going for a Snackbar that can timeout or show an 'empty view' with an error icon and text hinting to retry view pull-to-refresh.

For the empty screen stuff @tobiasKaminsky might be a good contact or me 🙃

AndyScherzinger (approved):

AndyScherzinger:
So I'd say feel free to merge but we should discuss a different UI approach, looping in @jancborchardt and @nimishavijay 👋

jld3103:
@mahibi and I talked about the way we want to display this and it has to be some kind of dialog or popup because messing with all the layout visibility breaks easily and this page needs to be rewritten properlay anyway. the dialog was just the least intrusive method into the logic, but we could also make it a snackbar, yes.

AndyScherzinger:
@jld3103 as I said all good from a technical perspective, so feel free to merge. 👍

tobiasKaminsky:
You already have emptyList in it (https://github.com/nextcloud/talk-android/blob/9bbeb9b420d4db3a24daea78f203a5bf79db0b7d/app/src/main/res/layout/controller_conversations_rv.xml#L45) and use it already when there is no conversation available:
https://github.com/nextcloud/talk-android/blob/d8d4fde39108b2e20a2eefea3a622a6a891d0b04/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.kt#L560

"just" setting the values there in case of error should be enough.

PR #2462, opened 2022-10-04
remove unused code from NotificationWorker

mahibi:


mahibi:
i changed this to draft.
this PR itself makes sense but in the next days i might change a lot in this class so this PR might be obsolete soon..

PR #2477, opened 2022-10-10
Fix crash when setting layout for call participants

### LINKED ISSUE #2445: App crashes while joining call on 15.0 RC3
### Steps to reproduce

Click to join call 
Rings 
Crashes

### Expected behaviour

Should work

### Actual behaviour

Not working 

https://user-images.githubusercontent.com/59488153/193141907-1b0e13c5-9367-49ee-b682-e0db9d052941.mp4

### Device brand and model

SM-G996U1 

### Android version

12

15.0 RC3

### Nextcloud server version

24.0.5

### Talk version

Lastes version 

No

Yes

Yes

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

danxuliu:
Fixes #2445, which is a regression introduced in #2390 (specifically, in https://github.com/nextcloud/talk-android/pull/2390/commits/c3f1f6c3a8ff74930b91b5f00c6435186f5ec1e2)

When the layout for a participant is setup and the HPB is not used the avatar is based on the ID of the user, [which is got from the list of known participants in the call](https://github.com/nextcloud/talk-android/blob/f880b4f7787d1349e809d3e0847921aaffe346f8/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L2364). However, the list of known participants [is fetched](https://github.com/nextcloud/talk-android/blob/f880b4f7787d1349e809d3e0847921aaffe346f8/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1891) just [before setting up the layout](https://github.com/nextcloud/talk-android/blob/f880b4f7787d1349e809d3e0847921aaffe346f8/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1904). Therefore, when the layout is setup the list of known participants does not contain yet the participant, which ended in a `NullPointerException`.

To prevent that now it is verified if the participant is in the list before getting the user ID, and if not an empty user ID is used. However, that would cause that all users are initially shown as guests. Fortunately the user ID is provided by the signaling message used to create the participant and thus setup the initial layout, so it can be provided to `setupVideoStreamForLayout` to show the proper avatar even if the list of known participants was not fetched yet.

Note that, although highly unlikely, it would be possible that a connection is established with the remote participant before the data for the participants was received. In that case the user ID is provided to `setupVideoStreamForLayout` when the remote participant was initially found, but not when the connection was established, so if the remote participant has her video disabled a guest avatar would be shown due to the missing user ID. This already happened before and it is not introduced (nor fixed) by this pull request.

Although the issues happens when the HPB is not used it would be good to test too with the HPB to ensure that nothing breaks ;-)

## How to test

- Do not setup the HPB
- Ensure that getting the peers will take longer than setting the layout by adding something like `sleep(2);` to the beginning of [`getPeersForCall`](https://github.com/nextcloud/spreed/blob/62f47af2b6143d7c9ff1f24b1ffce2ff6951bc7f/lib/Controller/CallController.php#L69)
- Start a call with the WebUI
- In the Android app, join the call

The app does not crash (and the avatar for the remote participant is shown before establishing a connection)

The app crashes


AndyScherzinger:
/stable-15.0

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -1862,6 +1863,15 @@ private void processUsersInRoom(List<HashMap<String, Object>> users) {
                 } else {
                     oldSessions.add(participant.get("sessionId").toString());
                 }
+
+                // The property is "userId" when not using the external signaling server and "userid" when using it.
```
timkrueger, line 1867: Is that difference for a reason? Or simply history?

timkrueger (approved):
Thank you  very much @danxuliu :blue_heart:

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
danxuliu: As far as I know, simply history :-)

danxuliu: For the record, after some investigation done for #2536 turns out that the comment (which was based on a comment in the WebUI code that has been there since forever) was not fully accurate (it is removed in #2536); although some external signaling messages actually use `userid` none of them were handled by the Android app, so `userId` could be always checked instead (and interestingly enough, after some archaeology it seems that `userId` could have been always checked in the WebUI too, but that is a different story :-P ).

Issue #2418, opened 2022-09-23
Server not supported


mahibi:
### Steps to reproduce

1. login at server with app 

### Expected behaviour

app starts, conversations loaded

### Actual behaviour
happens very unregulary:

error in UI: "Unsupported Server (The server version is too old.....)"

i debugged and saw that the capabilities of the user are `null` when checking for `hasSpreedFeatureCapability`:

```
hasSpreedFeatureCapability:56, CapabilitiesUtilNew (com.nextcloud.talk.utils.database.user)
isServerEOL:42, CapabilitiesUtilNew (com.nextcloud.talk.utils.database.user)
onAttach:266, ConversationsListController (com.nextcloud.talk.controllers)
```

maybe this is somehow related to https://github.com/nextcloud/talk-android/issues/2322 (although i wonder why no "Unsupported Server" error is shown there..)

### Device brand and model

Pixel XL

### Android version

13

### Nextcloud Talk app version

15.1.0 Alpha 02

### Nextcloud server version

25.0.0 beta 7 (sermo)

migulen:
Hi @mahibi , same happened to me on v15.0.0-RC1, 5 or 6 times when i was connected to a special Wifi net (extremelly firewalled and with low bandwith) and I was trying to new logon 2 accounts of my nextcloud servers.

After trying several times with the prior scenario,  I changed to me mobile net connection, 4G, and the problems gone... So I thought the problem was from "bandwith" or "the hard firewall" of that Wifi net. But I'm not sure. I'm going to try another time at the same Wifi net to see if it happens like before.  

I'll tell you something.

------------------------------------------

**Device brand and model**

Xiaomi Mi 9

**Android version**

12

**Nextcloud Talk app version**

15.0.0-RC1

**Nextcloud server version**

23.0.8

**Talk version**

13.0.8

**Custom Signaling server configured**

No response

**Custom TURN server configured**

Yes

**Custom STUN server configured**

Yes

**Android logs**

No response

**Server log**

No response

**Additional information**

No response

mahibi:
thanks for testing and reporting @migulen 

i guess there is something failing in UserManager (maybe depending on the quality of the internet connection).
Maybe @AndyScherzinger / @AlvaroBrey have any ideas here?

AndyScherzinger:
No clues unfortunately. In case of capabilities being null we could still say we default to "supported version" assuming the best case instead of the worst case.

mahibi:
[...]

PR #2485, opened 2022-10-12
Use already fetched capabilities for user

timkrueger:
This should avoid that the capabilities not available for the server EOL
check in 'ConversationListController#onAttach'.

Missing capabilites can also have an impact on multiple actions, but the
server EOL check is the first one.

Resolves: #2418


timkrueger:
@AndyScherzinger Can you have a look if that is a valid workaround in the context of the `UserManager`?

mahibi:
for the record: this was some sort of blind coding because we were not able to reproduce the issue. However we think that this could fix the bug.
There is quite a long chain with workers and thrown events from eventbus which might caused the reported issue.

AndyScherzinger:
I am not into the details of the user manager but the code look sensible, so I'd say let's give it a try, not sure though if it should go into 15.0.0 or rather a 15.0.1 RC shortly after the release since by now the window before the stable is very short.

AndyScherzinger (approved):

PR #2524, opened 2022-10-24
Add decision to use MADR

timkrueger:
Add first decision to use Markdown Any Decision Records. The current
status is 'proposed' and will be discussed in the regarding pull
request.

Also an template for decisions is added. It is a copy of the current
MADR 3.0 template.

It should be discussed:

- The decision 0000 - Use Markdown Any Decision Records.
- The scope of the template.

See: https://adr.github.io/madr/

----
This idea came up in our last Android meeting. We decided to make a pull request and discuss if we want to use decision records to improve our technical documentation. And be more kind to our future self's.
If we will, this will be added also to the android, android-common, android-config and android-library.

### DIFF
--- a/docs/decisions/template.md
+++ b/docs/decisions/template.md
@@ -1,10 +1,8 @@
 ---
-# These are optional elements. Feel free to remove any of them.
-status: {proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)}
+status: { rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)}
 date: {YYYY-MM-DD when the decision was last updated}
+# These are optional elements. Feel free to remove any of them.
 deciders: {list everyone involved in the decision}
-consulted: {list everyone whose opinions are sought (typically subject-matter experts); and with whom there is a two-way communication}
-informed: {list everyone who is kept up-to-date on progress; and with whom there is a one-way communication}
 ---
 # {short title of solved problem and solution}
 
--- a/docs/decisions/template.md
+++ b/docs/decisions/template.md
@@ -1,0 +1,80 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: {proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)}
+date: {YYYY-MM-DD when the decision was last updated}
+deciders: {list everyone involved in the decision}
+consulted: {list everyone whose opinions are sought (typically subject-matter experts); and with whom there is a two-way communication}
+informed: {list everyone who is kept up-to-date on progress; and with whom there is a one-way communication}
+---
+# {short title of solved problem and solution}
+
+## Context and Problem Statement
+
+{Describe the context and problem statement, e.g., in free form using two to three sentences or in the form of an illustrative story.
+ You may want to articulate the problem in form of a question and add links to collaboration boards or issue management systems.}
+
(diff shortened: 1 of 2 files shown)


docs/decisions/template.md
```diff
@@ -0,0 +1,80 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: {proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)}
```
AlvaroBrey, line 3: I would avoid both `proposed` and `rejected`, since we'll typically discuss the decisions in a PR.  Decisions should be PRd with `accepted` state, which will become true when merged, but rejected PRs will just be closed.

docs/decisions/template.md
```diff
@@ -0,0 +1,80 @@
...
+date: {YYYY-MM-DD when the decision was last updated}
+deciders: {list everyone involved in the decision}
+consulted: {list everyone whose opinions are sought (typically subject-matter experts); and with whom there is a two-way communication}
+informed: {list everyone who is kept up-to-date on progress; and with whom there is a one-way communication}
```
AlvaroBrey, line 7: I would skip `consulted` and specially `informed`. I don't think it adds much value. If anything special related to this happens, we can mention it in the relevant section (consulted in decision drivers,  informed in more info)

docs/decisions/template.md
AlvaroBrey: (I still think this field has value and should not be optional, because of the `deprecated` and `superseded` statuses)

timkrueger: > I would remove both `proposed` and `rejected`, since we'll typically discuss the decisions in a PR. Decisions should be PRd with `accepted` state, which will become true when merged, but rejected PRs will just be closed.

I agree with you about`'proposal`. But I think it is interesting to keep track of `rejected` records. It is also interesting to remember why you do not something.

AlvaroBrey (approved):
Let's see what @tobiasKaminsky and @mahibi think!

tobiasKaminsky:
I am fine with it. 
It feels a bit verbose and unusual, but I think this is because we never did it with a real decision.
Once we note something down, we can still adjust it a bit, if needed.

AndyScherzinger:
I have mixed feeling a bit based on what Tobias already said and besides being verbose I am unsure if it is worth the effort and also if 

> If we will, this will be added also to the android, android-common, android-config and android-library.

would be the way to go because that creates fragmentation, if you want or need to understand why things are build the way they are build, would you like to scan through several (Android related) repositories?

On the server-side of live this repo has been created to publicly discuss general aspects and come to a decision (not implying this is how Android should do it):
https://github.com/nextcloud/standards

What I like about it is that is is a central place to look for formerly made decisions as well as current discussions. And in a way since that issue tracker splits discussion and decision you have both in one place and can navigate each part depending on your need.

For me the question is about the value-add, so what would we like to achieve with this?
It is meant as a serious and honest question because I don't have a good answer but since I contribute to the app since the beginning I also never had the personal need for decision records and from a first-time contributor perspective I would probably rather just want the result like: "This is how we build up the app, i.e. using MVVM in general or coil for image loading", so I see a higher benefit in having an aggregated, dense overview of "the things you might want to know about our apps from a technical perspective before diving into the code".

So not denying the use just raising other points and questioning which measure to improve the technical documentation brings the biggest benefit for the given efforts needed. If the dev community around the Android apps grows then decision records might also be helpful, so I am not sure about the timing aspect then.

PR #2536, opened 2022-11-02
Add listeners for signaling messages

danxuliu:
~~Requires #2586 (to avoid a conflict when rebasing, it does not require it from a functional point of view)~~

This is a pure refactoring, so no behaviour changes are expected. The only exception is that now an error is no longer logged if the room type in a received signaling message is not `video` nor `screen`. However that was only useful during debugging and, in any case, it should never happen, so the lose is not high :-)

It is highly recommended to review this commit by commit rather than all at once ;-) (although a lot of content is just unit tests). Commit messages may also provide additional context.

Processing the signaling messages was moved from `CallActivity` and `MagicWebSocketInstance` to `SignalingMessageReceiver`. In the future all signaling related code should be extracted from `CallActivity` and `MagicWebSocketInstance` to its own class, but this is just a first step. Unfortunately this means that, for now, some boilerplate code had to be added to `CallActivity`.

`SignalingMessageReceiver` provides several listener interfaces for the different message kinds; listeners can be registered to be notified when a certain message is received. Right now this is useful just to handle WebRTC messages directly in the `PeerConnectionWrapper`, but it will be mostly used later when call participants and peer connections are split (right now every call participant is assumed to have a peer connection, which is not right).

Due to the use of explicit listeners some messages could be removed from EventBus methods. Note that the thread used to handle and notify messages from the external signaling server does not change; the EventBus subscriber mode was `BACKGROUND`, but as the messages were posted from a WebSocket handler, which runs in a worker thread rather than in the main thread, the subscriber was executed in the same thread as the poster.


danxuliu:
One is just a humble good ol' Java developer not proficient enough in Kotlin, flows and all those things that the cool kids are using today, so sorry for the added Java files and the basic Observer design pattern straight from the _Gang of Four_ ;-) (I know that Android Studio can convert Java files to Kotlin, but I preferred to keep the Java files as I may need to touch them later).

mahibi (approved):

PR #2545, opened 2022-11-07
Add listener for data channel messages

danxuliu:
~~Requires #2543 and #2540 (so the first commit of this pull request is _Replace concrete implementation with interface_)~~

This is (mostly) a pure refactoring, so (almost) no behaviour changes are expected. There might be some differences in very specific sequences of events regarding nicks, but in any case all this is just a temporary step and it will be adjusted later (when call participants and peer connections are split).

It is highly recommended to review this commit by commit rather than all at once ;-) Commit messages may also provide additional context.

An explicit listener was added for data channel messages, so EventBus is no longer used to handle them. The changes are still handled in (a private class of) `CallActivity`, but this will be adjusted at a later point. Unfortunately this means that, for now, some boilerplate code had to be added to `CallActivity`. Besides that, the nick property was moved out of the _PeerConnectionWrapper_, as it did not belong there.

Note that the thread used to handle the data channel messages has changed; the EventBus subscriber mode was "MAIN", but as the messages were posted from a DataChannel observer, which run in a worker thread rather than in the main thread, the subscriber was executed in the main thread rather than in the same thread as the poster. Due to this the actions performed by the handler now must be explicitly run in the main thread.


timkrueger (approved):
:clap:

PR #2572, opened 2022-11-15
Fix ParticipantDisplayItems not updated on events

danxuliu:
Follow up to #2503

What was I thinking on...? :facepalm: Sorry for the mistake :-(

The `ParticipantDisplayItems` were associated to both the session ID and the video stream type ("video" or "screen"), but the code that gets them was not updated to include the video stream type in the key.

I have pushed the commits in this pull request also to the backport for #2503, that is, #2564

Besides the steps below there are several other ways to test it (changing the nick, enabling or disabling the audio, or having a temporary disconnection by the remote participant)

## How to test

- Start a call with video in the WebUI
- Join the call from the Android app
- In the WebUI, disable the video

In the Android app the avatar is shown for the remote participant

In the Android app the video turns black (because the WebUI sends a black video, not because something was updated in the Android app), but the avatar of the remote participant is not shown


timkrueger (approved):

AndyScherzinger (approved):

danxuliu:
Please note that an independent backport is not needed, as the backported commits were already included in #2564

PR #2581, opened 2022-11-18
Bugfix/2446/fix disappearing accounts

### LINKED ISSUE #2446: Accounts disappear (+notifications are still delivered)
### Steps to reproduce

see below

### Expected behaviour

see both servers

### Actual behaviour

After updating from RC2 one of my two servers is gone, the account is away but I get notifications for this account. Reinstalling and adding my 2 servers does not solve it

### Device brand and model

Pixel 4a

### Android version

12

15.0.0-rc.1

### Nextcloud server version

_No response_

### Talk version

_No response_

_No response_

_No response_

_No response_

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

### LINKED ISSUE #2531: App keeps crashed on startup
### Steps to reproduce

Opened the app app keeps crashing

### Expected behaviour

Should not crash. This happen a few days ago and i cleared the cache and data off the app and it started to work again. 

I collacted some adb phone logs ping me via IM so i can send them to you. 

### Actual behaviour

https://user-images.githubusercontent.com/59488153/198362491-04d5a673-b29d-4eb3-b1e8-57959ea59261.mp4

### Device brand and model

SM-G996U1 

### Android version

12

15.0.1

### Nextcloud server version

24.0.6

### Talk version

14.0.5

No

Yes

Yes

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

mahibi:
fix #2446
fix #2531

# Problem
There was a race condition for "updateUser" between 

`userManager.saveUser(user).subscribe(new SingleObserver<Integer>() {`
in PushUtils#updatePushStateForUser

and

`if (userManager.setUserAsActive(user).blockingGet()) {`
in ChooseAccountDialogFragment > onSwitchItemClickListener

So while switching accounts via the Account switcher, also the PushUtil method is executed and sometimes asynchronously overwrote other user attributes with old values:

## Scenario1: no users are marked as "current" -> could lead to **app crash on startup**
As a result of this, there could be no users marked as "current". 
Because of this, 

```
@Query("SELECT * FROM User where current = 1")
abstract fun getActiveUser(): Maybe<UserEntity>
```

would return null, which in turn led to a NullPointerException in ConversationsListController for currentUser and the app crashed.
Because it's not a valid state to have no users marked as current, the app always crashed on startup until the app data was deleted, see issue https://github.com/nextcloud/talk-android/issues/2531

remarks: it cannot be 100% ruled out that there are other async scenarios that lead to this state. Anyway the NPE is now avoided so that the app does not crash but shows an error in UI.

## Scenario2: multiple users are marked as "current" -> **accounts are suddenly hidden**
There could be more than one user marked with "current". This caused to only show one of the current accounts in the account switcher while the other one was hidden. So it seemed that an account was lost (although it was still there, that's why push notifications still could have been received), see issue https://github.com/nextcloud/talk-android/issues/2446

# Solution
update only pushState when updating users (instead updating of whole user)
see commit https://github.com/nextcloud/talk-android/pull/2581/commits/b8e3ce457e882fb0fe4d4de02178c316ea025bf5

# How to test
set up multiple accounts and switch very fast and often between them via the account switch dialog (the avatar in the upper right corner)

## without this PR
sometimes scenario 1 or 2 could happen.

## with this PR
neither scenario 1 or 2 occur

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
@@ -230,6 +230,10 @@ class UserManager internal constructor(private val userRepository: UsersReposito
         return user
     }
 
+    fun updatePushState(id: Long, state: PushConfigurationState): Single<Int> {
+        return userRepository.updatePushState(id, state)
+    }
+
     companion object {
         const val TAG = "UserManager"
     }
--- a/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
@@ -58,22 +58,6 @@ class UserManager internal constructor(private val userRepository: UsersReposito
         return userRepository.getUserWithId(id)
     }
 
-    fun disableAllUsersWithoutId(id: Long): Single<Int> {
-        val results = userRepository.getUsersWithoutUserId(id)
-
-        return results.map { users ->
-            var count = 0
-            if (users.isNotEmpty()) {
-                for (entity in users) {
-                    entity.current = false
-                    userRepository.updateUser(entity)
-                    count++
-                }
-            }
-            count
-        }
-    }
-
     fun checkIfUserIsScheduledForDeletion(username: String, server: String): Single<Boolean> {
         return userRepository
             .getUserWithUsernameAndServer(username, server)
--- a/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
@@ -43,6 +43,7 @@ class UserManager internal constructor(private val userRepository: UsersReposito
     val currentUser: Maybe<User>
         get() {
             return userRepository.getActiveUser()
+                .switchIfEmpty(getAnyUserAndSetAsActive())
         }
 
     val currentUserObservable: Observable<User>
--- a/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/users/UserManager.kt
@@ -149,7 +149,7 @@ class UserManager internal constructor(private val userRepository: UsersReposito
                     else -> {
                         user.token = userAttributes.token
                         user.baseUrl = userAttributes.serverUrl
-                        user.current = true
+                        user.current = userAttributes.currentUser
                         user.userId = userAttributes.userId
                         user.token = userAttributes.token
                         user.displayName = userAttributes.displayName
@@ -237,7 +237,7 @@ class UserManager internal constructor(private val userRepository: UsersReposito
     data class UserAttributes(
         val id: Long?,
         val serverUrl: String?,
-        val currentUser: Boolean?,
+        val currentUser: Boolean,
         val userId: String?,
         val token: String?,
         val displayName: String?,
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
@@ -1661,7 +1661,7 @@ public class CallActivity extends CallBaseActivity {
     }
 
     private void processMessage(NCSignalingMessage ncSignalingMessage) {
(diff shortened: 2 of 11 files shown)


mahibi:
As a discussion:

i came to the solution because i saw that 
`cookieManager.cookieStore.removeAll()`
is used in combination with `setUserAsActive()` also in other classes, for example here:
https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/controllers/SwitchAccountController.kt#L91

But i would not be able to tell what exactly `cookieManager.cookieStore.removeAll()` does. Can you, @AndyScherzinger ?

AndyScherzinger:
@mahibi the call throws away all Cookie data. This is needed since else when switching to a user of the same URL then the http client would send the cookie data of the former user which is wrong and leads to issues with loading the wrong data on server side, hence the deletion of cookie data upon user switch.

mahibi:
> @mahibi the call throws away all Cookie data. This is needed since else when switching to a user of the same URL then the http client would send the cookie data of the former user which is wrong and leads to issues with loading the wrong data on server side, hence the deletion of cookie data upon user switch.

are you sure this only applies to users with the same URL?
in my scenario it was sermo, cnc and some complete different url.
other users for this issue reported they used 2 servers with different urls...

mahibi:
there must 2 different bugs!
- sometimes there is no new user created when trying to add him
- sometimes the user is added but other accounts remain "current"

mahibi:
added some debug logging.
i can't reproduce it now but didn't change anything. it's really hard to reproduce & debug! might be some async problem..

migulen:
Hi @mahibi, 

I'm testing this nextcoud-talk QA fix with 2 accounts in 2 different servers. It's the same scenario: "same servers, same mobile terminal and same mobile company and wifi networks" that I related in the original post/bug "2446", but these are testing accounts with different username).  By now, only 12 hours of use, I haven't any "disappearing account" with this bugfix... but it used only a bit.

On the same mobile terminal, the original scenario (real accounts) is configured with v15.0.1 nextcloud-talk at the same time. Today in the 12 hours of use, 2 o 3 times the reals account have disappeared, alternatively. 

I'll continue reporting.

migulen:
By the way, I've found that in the testing scenario (QA) I have this bug (https://github.com/nextcloud/talk-android/issues/2587 too), **but only happening on one of the new testing accounts ¿?¿?¿?** 

How it's possible ? It's the same app and I think that i've created the testing accounts on the 2 servers exactly with the same protocol... and configured at the app the same way.

:-(

migulen:
Hi @mahibi , I don't use so much the "test accounts" and for now, these haven't disappeared any time. May I reinstall 2581-talk.apk and continue testing ? Or 2581-talk.apk it's the same that I installed some days ago ?

"i was able to see that the loop was somehow interrupted during debugging which caused two users to have **current =true**" <- That's just what I thought.

mahibi:
if anyone is interested to test, please fell free to do so 
@migulen 
@jakobroehrl

app/src/main/java/com/nextcloud/talk/users/UserManager.kt
```diff
@@ -43,6 +43,7 @@ class UserManager internal constructor(private val userRepository: UsersReposito
     val currentUser: Maybe<User>
         get() {
             return userRepository.getActiveUser()
+                .switchIfEmpty(getAnyUserAndSetAsActive())
```
mahibi, line 46: this should fix the app on phones that have an invalid state because of no users marked as "current".
:+1: or :-1: @AlvaroBrey ?

AlvaroBrey: Yeah, I don't see any problem with this.

However I will say that most classes should aim to use `CurrentUserProvider` instead,  which always holds a reference to the updated current user, without having to call the DB every time.

mahibi: :+1: but for now PR approved?

AlvaroBrey: I don't have the bandwidth to review the full PR right now, but as I said, no problem with this.

timkrueger (approved):

AndyScherzinger (approved):

PR #2600, opened 2022-11-30
Split call participants and peer connections

### LINKED ISSUE #2544: Split call participants and peer connections
Currently all call participants are assumed to have a peer connection, which is not right; a participant who is not sending media (either because the participant does not have a microphone nor a camera selected, or because the participant does not have rights to publish) does not have a peer connection.

Besides solving some current issues (like trying to connect to participants without a peer connection), splitting call participants and peer connections is a pre-requisite to implement pro-active establishment of connections when call flags change, which is needed to get rid of forced reconnections.

Requires:
- [X] #2536
- [X] #2543 
- [X] #2545
- [X] #2590

danxuliu:
Fixes #2544
~~Requires #2590 (so the first commit of this pull request is _Handle the raw ICE connection state in the views_)~~

It is highly recommended to review this commit by commit rather than all at once ;-) Commit messages may also provide additional context.

Model classes for call participants were added. `CallParticipant` is the main model class, which observes the changes in its peer connections and data channels and updates a data object that keeps track of the state*. This data object is the one exposed to the view classes through a read-only object (`CallParticipantModel`) that can be also observed for changes on specific threads. Thanks to that the `ParticipantDisplayItems` are now automatically updated in the main thread when the model changes, and they also propagate the changes to the adapter (so `notifiyDataSetChanged` is automatically called when any item changed).

*In the future it should also listen to signaling messages, for example for raised hands, but also for audio/video enabled/disabled events, as they are also sent through signaling messages besides the data channels just in case the data channel connection could not be established.

Besides that, a helper class is now used to keep track of the participants in the call based on the signaling messages, instead of based on the peer connections. For now it just uses the same signaling messages as before, but in the future this could be used to also handle the message sent by the external signaling server when a participant leaves the room, which implicitly means that the participant left the call (if a participant abruptly leaves the call and the room at the same time no message for leaving the call is sent, only about leaving the room).

With all this now call participants are added and removed when they join and leave the call, and peer connections are only started when the joined participant is publishing media (according to the call flags of the participant for audio/video, and based on the offers received for screen shares). Updating the peer connections of an existing call participant when the flags change during the call is still needed for proper renegotiation support, and it will be done in a follow up.

The steps below should be tested with HPB; without HPB, as the Android app is publishing media, a peer connection is anyway needed to send it to the other participant even if that participant is not publishing.

## How to test

- Start a call without a microphone or camera selected in the browser (that is, not just disabled, but not selected), or without audio and video permissions

The Android app does not try to establish a peer connection with the other participant, and no progess bar is shown on that other participant

The Andoid app tries to establish a peer connection with the other participant to receive media (and the external signaling server logs eventually shows that an offer could not be requested), and a progress bar is always shown on that other participant

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
@@ -60,6 +60,8 @@ import com.nextcloud.talk.adapters.ParticipantDisplayItem;
 import com.nextcloud.talk.adapters.ParticipantsAdapter;
 import com.nextcloud.talk.api.NcApi;
 import com.nextcloud.talk.application.NextcloudTalkApplication;
+import com.nextcloud.talk.call.CallParticipantModel;
+import com.nextcloud.talk.call.MutableCallParticipantModel;
 import com.nextcloud.talk.data.user.model.User;
 import com.nextcloud.talk.databinding.CallActivityBinding;
 import com.nextcloud.talk.events.ConfigurationChangeEvent;
@@ -231,7 +233,6 @@ public class CallActivity extends CallBaseActivity {
     private MediaStream localStream;
     private String credentials;
     private List<PeerConnectionWrapper> peerConnectionWrapperList = new ArrayList<>();
-    private Map<String, String> userIdsBySessionId = new HashMap<>();
 
     private boolean videoOn = false;
     private boolean microphoneOn = false;
@@ -266,6 +267,8 @@ public class CallActivity extends CallBaseActivity {
     private Map<String, PeerConnectionWrapper.DataChannelMessageListener> dataChannelMessageListeners = new HashMap<>();
 
     private Map<String, PeerConnectionWrapper.PeerConnectionObserver> peerConnectionObservers = new HashMap<>();
+
+    private Map<String, MutableCallParticipantModel> callParticipantModels = new HashMap<>();
 
     private SignalingMessageReceiver.ParticipantListMessageListener participantListMessageListener = new SignalingMessageReceiver.ParticipantListMessageListener() {
 
@@ -382,6 +385,7 @@ public class CallActivity extends CallBaseActivity {
             requestBluetoothPermission();
         }
         basicInitialization();
+        callParticipantModels = new HashMap<>();
         participantDisplayItems = new HashMap<>();
         initViews();
         if (!isConnectionEstablished()) {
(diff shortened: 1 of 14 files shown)


danxuliu:
The two new issues reported by SpotBugs are both related to this code:
https://github.com/nextcloud/talk-android/pull/2600/files#diff-50dc4e0167d557f36af93be1eeb64df0cbd347b3a31a4da9d8a2f6a12514b9a5R94-R99

The _Performance - PSC: Presize Collection - Method does not presize the allocation of a collection_ seems to assume that `left` can be initialized to the same size as `knownCallParticipantsNotFound`, because the loop adds the contents from `knownCallParticipantsNotFound` to `left`. However, `left` will not have the same size as `knownCallParticipantsNotFound`, and in most cases it will be much lower, as [the known call participants that are still in the call are removed from `knownCallParticipantsNotFound`](https://github.com/nextcloud/talk-android/pull/2600/files#diff-50dc4e0167d557f36af93be1eeb64df0cbd347b3a31a4da9d8a2f6a12514b9a5R90). Of course `left` could be initialized anyway to the same size as `knownCallParticipantsNotFound`, but in most cases it will just reserve unneeded space, so to fix the performance issue reported by SpotBugs a performance "issue" might be introduced :-) (although in most cases it will be just a few items, so it should not matter much in practice).

The _Dodgy code - UAA: Use Add All - Method uses simple loop to copy contents of one collection to another_ is technically right, but it seems to ignore that the loop does other things with the added elements. Therefore, even if `left.addAll(knownCallParticipantsNotFound)` was used a loop would still need to be used anyway, and I think that the current loop is clearer as it shows in a more cohesive way the changes done to `callParticipant` before adding it to `left` (but, of course, that is just a matter of taste, so also fine by me to change it ;-) ).

mahibi:
tested with sermo, removed mic and video permissions of caller and also set mic and video to "none" when starting the call.
there is still a progress shown in android..
we can have a look together whenever you want @danxuliu

timkrueger (changes requested):
Spinning wheel still exists.

mahibi:
calling from my local machine without hpb with the "none" devices scenario, then the avatar of the android participant is spinning in the webbrowser

mahibi (approved):

danxuliu:
> calling from my local machine without hpb with the "none" devices scenario, then the avatar of the android participant is spinning in the webbrowser

Partially fixed (when the session ID of the Android app participant is higher than the session ID of the browser participant, which was a regression introduced in this pull request). When the session ID of the browser participant is the higher one the connection will still fail if the browser participant is not sending media; this was an existing issue and requires further refactorings so, for now, it is not fixed and tracked in https://github.com/nextcloud/talk-android/issues/2679

PR #2601, opened 2022-11-30
Fix reconnection when the web socket was abruptly closed

danxuliu:
When the web socket is abruptly closed it is connected again and the call is rejoined. However, the call was rejoined in a background thread, so an exception was thrown when trying to modify the views, which prevented the call from being joined again.

Besides that the call state needs to be explicitly changed, as if the web socket was connected again while in a call the state would be already "JOINED" or "IN_CONVERSATION", which prevents the signaling settings from being fetched again after the permissions check, and therefore also prevented the call from being joined again.

## How to test

- Setup the HPB
- Start a call in the WebUI
- Kill the external signaling server and start it again (note that only the external signaling server should be killed; Janus should not, as that would trigger a "publisher failed" reconnection, which is a different one)

The Android app joins the call again

The Android app tries to join the call, but the log shows an _Only the original thread that created a view hierarchy can touch its views._ exception

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
@@ -1578,7 +1578,8 @@ public class CallActivity extends CallBaseActivity {
                     if (currentCallStatus == CallStatus.RECONNECTING) {
                         hangup(false);
                     } else {
-                        initiateCall();
+                        setCallState(CallStatus.RECONNECTING);
+                        runOnUiThread(this::initiateCall);
                     }
                 }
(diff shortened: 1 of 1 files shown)


mahibi (approved):

danxuliu:
> The backport to stable-15.1 failed. Please do this backport manually.

Not needed, as it seems that _stable-15.1_ was branched after this pull request was merged, so [it is already included in _stable-15.1_](https://github.com/nextcloud/talk-android/blame/stable-15.1/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1609) :-)

PR #2605, opened 2022-12-01
add initial pull request template for github

mahibi:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,18 @@
+### 🖼️ Screenshots
+
+🏚️ Before | 🏡 After
+---|---
+B | A
+
+
+### 🚧 TODO
+
+- [ ] ...
+
+### 🏁 Checklist
+
+- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
+- [ ] 🔖 Capability is checked or not needed 
+- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
+- [ ] 📅 Milestone is set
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,18 @@
...
+### 🚧 TODO
+
+- [ ] ...
+
+### 🏁 Checklist
+
+- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
+- [ ] 🔖 Capability is checked or not needed 
+- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
+- [ ] 📅 Milestone is set
```
nickvergessen, line 17: IIRC people outside of the Org can not set milestones, so not sure this makes a lot of sense.

nickvergessen (approved):

.github/pull_request_template.md
mahibi: i would say not all of the points from the checklist must be checked by community members.
so for backport requests and milestone i suggest to add "(should be done by nextcloud employees)" ?

AndyScherzinger: Best to add an not rendered comment about the optional items on the list, same is true for screenshots

nickvergessen: But then people think they can skip it although they should just do 2 screenshots.
Let's keep it and the milestone for now and see if there is community feedback afterwards

AndyScherzinger: Makes sense! Release early, release often :)

AndyScherzinger (approved):

PR #2614, opened 2022-12-05
Feature/2554/call recording support

### LINKED ISSUE #2554: ⏺️ Call recording support
See meta issue in https://github.com/nextcloud/spreed/issues/8324

## Eta 30.11.
- [x] Showing the system messages
- [x] Show "Recording icon" in the call view when the call is being recorded

mahibi:
resolve #2554

- only with High Performance backend

### 🖼️ Screenshots

advanced call options menu | recording in action | stop recording confirm dialog 
------- | -------- | -------- 
grafik | grafik | grafik

### 🚧 TODO

- [x] set recording state when entering a call 
- [x] set recording state when recording is set from other device during call (signaling)
- [x] check moderator rights? (or is capability enough?)
- [x] comment out hardcoded parts and replace with actual checks/api calls once API is ready

#### followup: 
- handle notification to share recording to chat #2722

### 🏁 Checklist

- [x] 📅 Milestone is set


app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -387,6 +431,18 @@ public void onStart() {
         }
     }
 
+    private void initFeaturesVisibility() {
+        // TODO: check for isAllowedToRecordCall once api is ready
```
mahibi, line 435: - [ ] change this when API is ready

app/src/main/java/com/nextcloud/talk/repositories/callrecording/CallRecordingRepositoryImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    var apiVersion = ApiUtils.getCallApiVersion(currentUser, intArrayOf(ApiUtils.APIv4, 1))
+
+    override fun startRecording(
+        roomToken: String
+    ): Observable<StartCallRecordingModel> {
+        return Observable.just<StartCallRecordingModel>(StartCallRecordingModel(true))
+    }
+
+    // override fun startRecording(
```
mahibi, line 46: change this when API is finished

mahibi:
for now only added @AlvaroBrey as reviewer because we might have a look how to write tests for MVVM and adjust some code for this..

abdusalamApps:
@mahibi 
Amazing! Great job!
I just got my Nextcloud Talk working with a TURN server. I'll try out the apk and try to give my feedback.
How do you want to get the feedback. Do I just write here?

mahibi:
> @mahibi Amazing! Great job! I just got my Nextcloud Talk working with a TURN server. I'll try out the apk and try to give my feedback. How do you want to get the feedback. Do I just write here?

Sure just write here, thanks for testing :+1: 
Be aware that this is still work in progress on server side, so the recording itself is not working yet and there might still be some changes or bugs to fix.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -2701,7 +2797,7 @@ public void onComplete() {
 
         /**
          * Adds the local participant nick to offers and answers.
-         *
+         * <p>
```
nickvergessen, line 2800: is never closed

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -2505,8 +2600,9 @@ private void stopCallingSound() {
     }
 
     /**
-     * Temporary implementation of SignalingMessageReceiver until signaling related code is extracted from CallActivity.
-     *
+     * Temporary implementation of SignalingMessageReceiver until signaling related code is extracted from
+     * CallActivity.
+     * <p>
```
nickvergessen, line 2605: is never closed

app/src/main/java/com/nextcloud/talk/models/json/converters/EnumSystemMessageTypeConverter.kt
```diff
@@ -175,6 +177,8 @@ class EnumSystemMessageTypeConverter : StringBasedTypeConverter<ChatMessage.Syst
             "poll_closed" -> POLL_CLOSED
             "message_expiration_enabled" -> MESSAGE_EXPIRATION_ENABLED
             "message_expiration_disabled" -> MESSAGE_EXPIRATION_DISABLED
+            "recording_started" -> RECORDING_STARTED
+            "recording_stopped" -> RECORDING_STOPPED
```
nickvergessen, line 181: audio_recording_started and audio_recording_stopped also exist

app/src/main/java/com/nextcloud/talk/utils/database/user/CapabilitiesUtilNew.kt
```diff
@@ -98,6 +98,17 @@ object CapabilitiesUtilNew {
         return false
     }
 
+    @JvmStatic
+    fun isCallRecordingAvailable(user: User): Boolean {
+        if (hasSpreedFeatureCapability(user, "recording-v1") &&
+            user.capabilities?.spreedCapability?.config?.containsKey("call") == true
+        ) {
+            val map: Map<String, String>? = user.capabilities!!.spreedCapability!!.config!!["call"]
+            return map != null && map.containsKey("recording")
```
nickvergessen, line 107: containsKey says trua also when it's set to false and actually not available?

app/src/main/res/values/strings.xml
```diff
@@ -557,6 +558,18 @@
     <string name="audio_output_dialog_headline">Audio output</string>
     <string name="audio_output_wired_headset">Wired headset</string>
 
+    <!-- Advanced call options -->
+    <string name="call_more_actions_dialog_headline">Advanced call options</string>
+
+    <!-- Call recording -->
+    <string name="record_start_description">Start recording</string>
+    <string name="record_start_loading">starting…</string>
```
nickvergessen, line 566:
```suggestion
    <string name="record_start_loading">Starting …</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -557,6 +558,18 @@
...
+    <string name="record_stop_description">Stop recording</string>
+    <string name="record_stop_loading">stopping…</string>
```
nickvergessen, line 568:
```suggestion
    <string name="record_stop_loading">Stopping …</string>
```

Should have space before `…` as we are not trimming the word?

app/src/main/res/values/strings.xml
```diff
@@ -557,6 +558,18 @@
...
+    <string name="record_stop_confirm_title">Stop Call recording</string>
+    <string name="record_stop_confirm_message">"Do you really want to stop the recording?"</string>
+    <string name="record_active_info">"The call is being recorded"</string>
```
nickvergessen, line 571: Quotes intentional?

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -371,6 +390,41 @@ public void onCreate(Bundle savedInstanceState) {
...
         }
 
+        callRecordingViewModel = new ViewModelProvider(this, viewModelFactory).get((CallRecordingViewModel.class));
+        callRecordingViewModel.setData(roomToken);
+        callRecordingViewModel.setRecordingState(extras.getInt(KEY_RECORDING_STATE));
+
+        callRecordingViewModel.getViewState().observe(this, viewState -> {
+            if (viewState instanceof CallRecordingViewModel.RecordingStartedState) {
+                binding.callRecordingIndicator.setVisibility(View.VISIBLE);
+                if (((CallRecordingViewModel.RecordingStartedState) viewState).getShowStartedInfo()) {
```
AlvaroBrey, line 400: You may want to call back to the ViewModel so that `showStartedInfo` is set back to `false`. Otherwise the state will stay and if, for example, the phone is rotated, or the user minimizes and then resumes the app, the toast and vibration will happen again.

AlvaroBrey (approved):
From an architecture perspective this looks great!

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
mahibi: actually <p> seems to be added by autoformat from now on! But it's intended to not close it, see https://metaterasology.github.io/docs/workflow/documentationJavaDoc.html#html-tags

app/src/main/res/values/strings.xml
mahibi: nope

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
mahibi: well ViewModel will survive configuration changes right?!
and the Observer is only triggered when the state is set which only happens for defined actions..

> for example, the phone is rotated, or the user minimizes and then resumes the app, the toast and vibration will happen again.

so could not reproduce this and i would say it's okay to keep it like this.

AlvaroBrey: Hm, may depend on how the Controllers work/are setup then. In a Fragment, the ViewModel would survive configuration changes but the Fragment would not, which means the Fragment would be recreated, and then process the state again, showing the toast again.

Issue #1551, opened 2021-08-26
get rid of Conductor


mahibi:
- [x] blocked by https://github.com/nextcloud/talk-android/issues/1550

We want to switch to native Android solutions instead to use Conductor.

TODOS:
- [x] research how to avoid a big bang switch

nickvergessen:
Moved to "To do" to perform the "Research how to avoid a big bang switch"
Please let us know if it can be avoided somehow

mahibi:
well the picture in picture feature actually made a start to replace a controller from conductor with a native activity. 
This was possible because there was a controller that was on the top of the stack, which could be replaced easily.
However other controllers might be more difficult to replace because they are not always on top of the stack or there is more interaction with other controllers.
So some research here still totally makes sense..

AndyScherzinger:
Given @mahibi's remarks I think it is also totally fine if there is no way to easily migrate all conductors but if there is a way to split the work like with the PIP then this would be a valid and helpful approach, so there can be valid, creative approaches. The goal would simply be to find a way to not have to do big bang even when it is a exotic approach :)

tobiasKaminsky:
If you need any help, ping me :)

mahibi:
> The goal would simply be to find a way to not have to do big bang even when it is a exotic approach :)

an example for one "exotic" approach:
In https://github.com/nextcloud/talk-android/pull/1285 i created a BottomSheetDialog that defined another conductor router:
https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/ui/dialog/ConversationsListBottomDialog.kt#L274

This was necessary because the BottomSheetDialog just worked differently than the formerly used BottomSheetMenu ( https://github.com/Kennyc1012/BottomSheetMenu )

I introduced the additional router to reuse the existing Controllers for now. So i didn't actually replace Controllers yet, but the router stack was "cut" into pieces with that. Whenever some button in the Dialog should trigger an action in the app, i throw an event like here:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/controllers/bottomsheet/OperationsMenuController.java#L570

which is catched inside the ConversationsListController...

https://github.com/nextcloud/talk-android/blob/f4cb9578136367f8b85d99834cb3872508a75aa0/app/src/main/java/com/nextcloud/talk/controllers/ConversationsListController.java#L1072

... which then closes the Dialog.
[...]

Issue #218, opened 2018-06-27
Design : enable a more compact display of the text messages


GLLM:
Hi !
when using NC talk (text messages) on Android (8.0, latest released version), I find that the display of chats is not taking advantage of the screen estate : 
when the keyboard is visible NC talk displays only 7 lines of text (among 2 people's msg) while whatsapp displays 11, and signal 11 as well.

I findh that the font is too big and the spacing betwenn the bubbles is to large + the empty space within a bubble (between text and border) is too large as well.

Am I the only one to wish for a way more compact design ?
Thanks

jancborchardt:
Good point. Our text is not even bigger than in Signal.

We can put the time into the bubble, and next to it display a sent confirmation as mentioned in #219. And the padding of the bubble can be reduced.

mario:
Time was initially in the bubble. Then we moved it out of the bubble. Are we really moving it OUT of the bubble again?

mario:
cc @skjnldsv

jancborchardt:
@mario if that's a problem, focus on only the bubble padding first.

mario:
I'm not sure this is a trivial change at this point.

skjnldsv:
I would stack all the messages from the same user during a specific period of time. :)

mario:
@skjnldsv that's planned and in a separate issue already - we agreed to do this already ;)

skjnldsv:
So lower the bubble padding and stack up should be enough?
![screenshot_20180627-193204](https://user-images.githubusercontent.com/14975046/41990085-b732db94-7a41-11e8-82be-5b1e0482d9d6.png)

As a reference: my sms app which I really like
![screenshot_20180508-141548](https://user-images.githubusercontent.com/14975046/41990168-f38f8952-7a41-11e8-9144-17a820f23d42.png)

mario:
This is exactly what I'll do. Eventually.

skjnldsv:
You rocks!
@jancborchardt opinions? :)

jancborchardt:
Cool! I mean, to be clear, only reducing the bubble padding is fine. The rest of the style, like the big bubble border-radius is great.

GLLM:
The large radius is bit too large IMHO .... sooooo iphone3ish :-)

mario:
4dp
[...]

PR #2620, opened 2022-12-08
Bugfix/2181/fix conversation title

### LINKED ISSUE #2181: Conversation title often shows different recent conversation
### Steps to reproduce
No specific steps to reproduce, but this happens often here in regular use of the app, closing again, using multiple channels.

### Expected behaviour
The title in the header is always the one of the conversation shown.

### Actual behaviour
Sometimes it glitches out and shows a title which does not match the conversation being shown. (The conversation is always the correct one as tapped.)
As shown here, the conversation is with Nimisha, but the title is from the Groupware team. This creates uncertainty and possible privacy issues as it’s unclear where the message will be sent:
Talk header

## Device information

**Device:** Fairphone 3+

**Android version:** 10

**Talk version:** 14.0.2

## Server information

All current cloud.nextcloud.com

mahibi:
fix #2181

# Steps to reproduce
## how to reproduce:
- https://github.com/nextcloud/talk-android/issues/2181#issuecomment-1172224102

## Actual behaviour
#2181 (sometimes avatar/title are wrong + most likely other weird bugs)

## Expected behaviour
avatar and title always match to the opened conversation

# Problem

issue #218 is caused by the sessionId being "0" when switching between chats. This happens only sometimes.

sessionId being "0" has the following effect which explains the issue:

when going back to another chat from the chat that had the sessionId "0":
-> validSessionId() will be false in onDetach
	-> leaveRoom is not executed
		-> disposable is not disposed
			-> getRoomInfo() continues to execute for controller which had the sessionId "0"
				-> e.g appbar infos can be periodically overwritten (wrong avatar/title)

The api calls to leave roomA and enter roomB will be mixed and maybe this is a problem...?
so after leaveRoom is executed for roomA, then sessionId "0" is retrieved for roomB in getRoomInfo#onNext.
(although roomB already retrieved a valid sessionId just before leaveRoom was execcuted for roomA)

this might be caused by https://github.com/nextcloud/spreed/issues/4670#issuecomment-1300461242 or https://github.com/nextcloud/spreed/issues/8416

# Solution

this PR makes sure that a room is left BEFORE a new conductor controller [is added](https://github.com/nextcloud/talk-android/pull/2620/files#diff-2a0db874d7779de90a1a29521d56db79fbc0d78ae76f0497885e86a277bf31ebR81) or [replaced](https://github.com/nextcloud/talk-android/pull/2620/files#diff-2a0db874d7779de90a1a29521d56db79fbc0d78ae76f0497885e86a277bf31ebR79) by using a callback. This will avoid sessionId to become "0".
This might not be the BEST solution, but it should do the job until conductor is replaced (#1551).
However in theory there might still be rare scenarios to trigger this bug by using the 'back' buttons (Although i was not able to reproduce such a scenario). But these scenarios should at latest be fixed by https://github.com/nextcloud/spreed/issues/4670#issuecomment-1300461242

also done in this PR:
- add logging in ChatController (should be worth it to keep it for other debugging)
- refactor ConductorRemapping.kt

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/MainActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/MainActivity.kt
@@ -50,13 +50,13 @@ import com.nextcloud.talk.databinding.ActivityMainBinding
 import com.nextcloud.talk.models.json.conversations.RoomOverall
 import com.nextcloud.talk.users.UserManager
 import com.nextcloud.talk.utils.ApiUtils
-import com.nextcloud.talk.utils.ConductorRemapping.remapChatController
 import com.nextcloud.talk.utils.SecurityUtils
 import com.nextcloud.talk.utils.bundle.BundleKeys
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ACTIVE_CONVERSATION
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_ID
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_TOKEN
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_USER_ENTITY
+import com.nextcloud.talk.utils.remapchat.ConductorRemapping.remapChatController
 import io.reactivex.Observer
 import io.reactivex.SingleObserver
 import io.reactivex.android.schedulers.AndroidSchedulers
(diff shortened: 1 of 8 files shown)


timkrueger (approved):

mahibi:
as this might be released as hotfix i also added @AlvaroBrey & @tobiasKaminsky for review if you find some time

PR #2671, opened 2023-01-09
 Split different exception scopes / Use own OkHttp client for coil

### LINKED ISSUE #2856: Image thumbnails not displaying
### Steps to reproduce

Running latest beta version of next cloud talk client (16.0.0 RC2).
Server running 25.0.4
Upload image to conversation. If you click on image in android client you can see it, but thumbnail is a default thumbnail.
Shows up just fine in web.
No user thumbnails or initials show either.
Fo me this happens on every image and all user thumbnails/initials on the talk app 

### Expected behaviour

See proper thumbnails and user images/initials

### Actual behaviour

No thumbnails and no user thumbnails/initials 

### Device brand and model

Pixel 7

### Android version

13

16.0.0 RC2

### Nextcloud server version

25.0.4

### Talk version

15.0.4

None

None

None

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

timkrueger:
fix #2856

Before this change the initinalization of the trusted key store was
combined in one try-catch-block. If anything went wrong an new in memory
trusted key store was created. Programming against an exception is an
bad pattern.

So the initialization is now splitted into multiple try-catch-blocks
with it's own scopes and the decision if the trusted key store is newly
created in memory or loaded from a exisitng file is done by an if
condition check.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
@@ -36,7 +36,7 @@ import com.nextcloud.talk.events.CertificateEvent
 import com.nextcloud.talk.ui.theme.ViewThemeUtils
 import com.nextcloud.talk.utils.SecurityUtils
 import com.nextcloud.talk.utils.preferences.AppPreferences
-import com.nextcloud.talk.utils.ssl.MagicTrustManager
+import com.nextcloud.talk.utils.ssl.TrustManager
 import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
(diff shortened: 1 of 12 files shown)


mahibi:
with this PR, it's now possible to log in with https on my local machine (which fails without this PR), but after login
avatars of the participants are not loaded. in the log there is:

```
E/RealImageLoader: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
        at com.google.android.gms.org.conscrypt.ConscryptFileDescriptorSocket.startHandshake(:com.google.android.gms@224915028@22.49.15 (100400-499306216):25)
        at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.kt:379)
        at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.kt:337)
```

This is most probably related to coil.
Probably coil also needs to trust the certificate if the main okhttp client trusts it..

mahibi (changes requested):
TODO: coil needs to accept the certificate too. Otherwise avatars might not load.

mahibi:
While testing/debugging this: always remember to delete the app before reinstall it, otherwise the certificate remains!

tobiasKaminsky (approved):

PR #2697, opened 2023-01-17
fix missing file icons in chat

### LINKED ISSUE #2691: File icons in chat are missing
### Steps to reproduce

1.) Open Talk
2.) Send file attachment
3.) Icon for file is missing, so it's not clickable to open it

Workaround: Shared-Items  via menu.

### Expected behaviour

Open the attachment

### Actual behaviour

Attachment only displayed as text

### Device brand and model

diverse

### Android version

na

15.1.0

### Nextcloud server version

25.0.2

### Talk version

15.0.2

None

None

None

### Android logs

_No response_

### Server log

_No response_

### Additional information

In android app version 15.0.3 the attachments worked

mahibi:
fix #2691

bug was introduced with fresco->coil migration

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
@@ -108,28 +108,7 @@ abstract class PreviewMessageViewHolder(itemView: View?, payload: Any?) :
         image.minimumHeight = DisplayUtils.convertDpToPixel(MIN_IMAGE_HEIGHT, context).toInt()
 
         time.text = dateUtils.getLocalTimeStringFromTimestamp(message.timestamp)
-        if (userAvatar != null) {
-            if (message.isGrouped || message.isOneToOneConversation) {
-                if (message.isOneToOneConversation) {
-                    userAvatar.visibility = View.GONE
-                } else {
-                    userAvatar.visibility = View.INVISIBLE
-                }
-            } else {
-                userAvatar.visibility = View.VISIBLE
-                userAvatar.setOnClickListener { v: View ->
-                    if (payload is MessagePayload) {
-                        (payload as MessagePayload).profileBottomSheet.showFor(
-                            message.actorId!!,
-                            v.context
-                        )
-                    }
-                }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
```diff
@@ -228,6 +214,35 @@ abstract class PreviewMessageViewHolder(itemView: View?, payload: Any?) :
             true,
             viewThemeUtils!!
         )
+
+        // super.onBind is at this position, because:
+        // - it needs to be called after "placeholder" is set (otherwise placeholders are initially not loaded)
+        // - it needs to be before the show/hide logic is set (because super methods also have logic for this, which
+        // needs to be overwritten again)
+        super.onBind(message)
```
AlvaroBrey, line 222: This feels like a bit too finicky of a solution, but I guess it's fine in the interest of fixing the bug ASAP. At a minimum I would separate the logic to get the placeholder into its own method and call only that before `super.onBind` instead of all the rest.

I don't understand how this worked before with Fresco :\ It seems like the same problem would happen

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
mahibi: yes it's not a clean solution but it's gonna be a hotfix asap.
will extract the placeholder logic.

the "placeholder" solution was introduced when coil was introduced. but now i think of an easier solution. will have another look in the evening..

mahibi: for now i moved the logic for the placeholder to getPayloadForImageLoader(...)

this may still be not the best solution because the logic inside is also responsible for other things. 
but i think it's better than to call the super method in between other code which is now avoided.
see commit 4ccf8ac5a28e13a3a9f0095b7e0711251e876a4c
cc @AlvaroBrey

nickvergessen (approved):
🐘  Fixes the issue

PR #2708, opened 2023-01-19
avoid NPE on hangup

mahibi:

### 🏁 Checklist

- [x] 📅 Milestone is set


danxuliu (approved):
Makes sense :+1: 

I did not guard it because [`callParticipantList` is initialized in `performCall`](https://github.com/nextcloud/talk-android/blob/67e259f79257fe5cd7679f872f7546737d2d5fbf/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L1466), so I assumed that it will be always initialized when `hangup` was called.

But a few things are done before `performCall` is called, so the [hangup button can be clicked](https://github.com/nextcloud/talk-android/blob/67e259f79257fe5cd7679f872f7546737d2d5fbf/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java#L464-L466) before that, specially if the server takes a bit to answer the requests :see_no_evil:

PR #2741, opened 2023-02-07
Optimize scroll-to-bottom

AndyScherzinger:
* Replace imageButton with MaterialButton (to also have a ripple effect)
* Apply primary tonal dynamic coloring
* Replace icon with a version with higher emphasis (translates to thickness in M3)
* resize 44dp ->48dp as minSize for accessibility
* raise end-padding from 6dp tp 8dp to stay in the typical Material Design spacings

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_20230207_103440 | Screenshot_20230208_093815

### 🚧 TODO

- [x] all good to go

### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
@rapterjet2004 just for your interest, this is my follow-up PR to your scroll-down work adding the theming optimizations as mentioned in the review. In case your are interested :)

AndyScherzinger:
Also CC @nimishavijay for the visual optimizations check 💙

mahibi (approved):

nimishavijay:
Looks super nice! :) Looking at the [m3 FAB specs](https://m3.material.io/components/floating-action-button/specs) do we also want to:
- Make it a rectangle with `border-radius: 12dp` instead of a circular button? 
- Add a shadow (I would this is not necessary since it is not a primary action like creating a conversation as the examples in the specs, but what do you think?)

Additionally we could increase the margin on the right of the button such that it is right aligned with the messages.
image

Other than that it is great! :)

AndyScherzinger:
Hi @nimishavijay, 

see https://github.com/nextcloud/talk-android/pull/2741#issue-1573995363 I changed 

* the right padding and 
* added a subtle drop-shadow for the elevation.

I would keep it circular-rounded - not like a fab - because it is technically not a fab, also checked signal, telegram, whatsapp and Google messages and they all use a circular button like we do for the scroll-to-bottom.

app/src/main/java/com/nextcloud/talk/controllers/ChatController.kt
```diff
@@ -674,6 +674,7 @@ class ChatController(args: Bundle) :
         }
 
         binding?.scrollDownButton?.setOnClickListener { binding?.messagesListView?.smoothScrollToPosition(0) }
```
mahibi, line 676: @AndyScherzinger 
smoothScrollToPosition is a bit jerky for me.
i suggest we change smoothScrollToPosition to scrollToPosition which doesn't have the automated scrolling but feels smoother UX wise..?

AndyScherzinger: Yes, just did via https://github.com/nextcloud/talk-android/pull/2741/commits/45f6e3b47426d6dd1fa6bd869adace8c3f839525 - also check whatsapp and they also do it like you suggested @mahibi 👍

PR #1767, opened 2022-01-10
only show mark as unread in debug mode

AndyScherzinger:
@mahibi follow-up to hide the "mark as unread" for now on releases but show on debug builds to be able to easily work on it in the future when time permits

Signed-off-by: Andy Scherzinger <info@andy-scherzinger.de>

PR #2742, opened 2023-02-07
Activate mark-as-unread on message level for everyone

AndyScherzinger:

@mahibi discussed with @nickvergessen - we shall activate the feature for everybody even though it won't always work (hidden in case the code couldn't determin the `id` of the previous message (needed for the unread marker)

### 🏁 Checklist

- [x] 📅 Milestone is set


nickvergessen (approved):

mahibi:
> @mahibi discussed with @nickvergessen - we shall activate the feature for everybody even though it won't always work (hidden in case the code couldn't determin the `id` of the previous message (needed for the unread marker)

okay. 

so for context, see:
https://github.com/nextcloud/talk-android/pull/1752
https://github.com/nextcloud/talk-android/pull/1767

nickvergessen:
PS Should check capability, as the endpoint is only available in Nextcloud 17+ which is currently higher then the minimum

AndyScherzinger:
PS: aLrEaDy PrEsEnT :)

PR #2745, opened 2023-02-08
Bugfix/noid/fix exceptions

mahibi:
### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
@@ -2593,12 +2593,18 @@ public class CallActivity extends CallBaseActivity {
 
     private void stopCallingSound() {
         if (mediaPlayer != null) {
-            if (mediaPlayer.isPlaying()) {
-                mediaPlayer.stop();
+            try {
+                if (mediaPlayer.isPlaying()) {
+                    mediaPlayer.stop();
+                }
+            } catch (IllegalStateException e) {
+                Log.e(TAG, "mediaPlayer was not initialized", e);
+            } finally {
+                if (mediaPlayer != null) {
+                    mediaPlayer.release();
+                }
+                mediaPlayer = null;
             }
-
-            mediaPlayer.release();
-            mediaPlayer = null;
         }
     }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -2591,12 +2593,16 @@ private void playCallingSound() {
...
-                mediaPlayer.stop();
-            }
+            try {
+                if (mediaPlayer.isPlaying()) {
+                    mediaPlayer.stop();
+                }
 
-            mediaPlayer.release();
-            mediaPlayer = null;
+                mediaPlayer.release();
```
tobiasKaminsky, line 2601: if mediaPlayer.stop() fails, then it will never be released.
Put it in finally, so that it will always be reached.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
```diff
@@ -2591,12 +2593,16 @@ private void playCallingSound() {
...
-            if (mediaPlayer.isPlaying()) {
-                mediaPlayer.stop();
+            try {
+                if (mediaPlayer.isPlaying()) {
+                    mediaPlayer.stop();
+                }
+            } catch (IllegalStateException e) {
+                Log.e(TAG, "mediaPlayer was not initialized", e);
+            } finally {
+                mediaPlayer.release();
```
AlvaroBrey, line 2603: Should check for `null` here as well, as the field may have become `null` at this point.

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/talk/activities/CallActivity.java
mahibi: Redundant nullcheck of com.nextcloud.talk.activities.CallActivity.mediaPlayer, which is known to be non-null in com.nextcloud.talk.activities.CallActivity.stopCallingSound()

https://www.kaminsky.me/nc-dev/talk-android-findbugs/920.html

AlvaroBrey: But it's not known? it's a field, which means something else could change it in between... Either way, I don't think it's that important, and will be "fixed" when this is ported to Kotlin anyway :)

PR #2787, opened 2023-02-21
Add explanation for translations

mahibi:


mahibi:
@nickvergessen everything correct? / anything missing?

PR #2802, opened 2023-02-27
Implemented automatic scrolling to quoted messages on tap

### LINKED ISSUE #1090: automatically scroll to quoted message
add the ability to tap on a quoted message to automatically scroll to it

rapterjet2004:
fixes #1090 

Implemented automatic scrolling to a quoted message on tap. 

### 🖼️ Video
[issue-1090-test-webm.webm](https://user-images.githubusercontent.com/69230048/221612508-f54713d0-262c-40cf-b092-cc4281248b44.webm)

### 🚧 TODO

- [x] basic functionality

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingTextMessageViewHolder.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingTextMessageViewHolder.kt
@@ -40,6 +40,7 @@ import coil.load
 import com.nextcloud.talk.R
 import com.nextcloud.talk.application.NextcloudTalkApplication
 import com.nextcloud.talk.application.NextcloudTalkApplication.Companion.sharedApplication
+import com.nextcloud.talk.controllers.ChatController
 import com.nextcloud.talk.databinding.ItemCustomIncomingTextMessageBinding
 import com.nextcloud.talk.extensions.loadBotsAvatar
 import com.nextcloud.talk.extensions.loadChangelogBotAvatar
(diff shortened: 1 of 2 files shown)


AndyScherzinger (approved):
works like a charm! Awesome work as always @rapterjet2004 🥇

mahibi:
this works great, thank you @rapterjet2004 :+1: :+1: 

possible follow ups for later:

- when loading of images etc is involved it might happen that the scrolled position is a bit off and the message is not visible. But that would require a lot more work i guess.

- highlight the original message for a second

rapterjet2004:
@mahibi 

Thanks! About that follow up on highlighting the original message, would that just be changing the background color of the message and using a timer to reset it? I know that on Talk web there's a kind of "fade out" effect to the original message's color, I admit I'm not sure how I can implement that as I'm not really familiar with how this app uses theming.

mahibi:
> @mahibi
> 
> Thanks! About that follow up on highlighting the original message, would that just be changing the background color of the message and using a timer to reset it? I know that on Talk web there's a kind of "fade out" effect to the original message's color, I admit I'm not sure how I can implement that as I'm not really familiar with how this app uses theming.

Maybe a ripple effect that is triggered programmatically would be good enough (something like [this](https://stackoverflow.com/questions/27225014/how-to-trigger-ripple-effect-on-android-lollipop-in-specific-location-within-th) maybe). What do you think @AndyScherzinger ?

PR #2848, opened 2023-03-09
Fix horizontal (=landscape) call design

mahibi:
Without this fix, the call buttons became huge in landscape mode.

The LinearLayout of the callControls now has a fixed width which comes in handy in landscape mode.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/layout/call_activity.xml
+++ b/app/src/main/res/layout/call_activity.xml
@@ -181,8 +181,9 @@
 
     <LinearLayout
         android:id="@+id/callControls"
-        android:layout_width="match_parent"
         android:layout_height="@dimen/call_controls_height"
+        android:layout_width="450dp"
+        android:layout_centerHorizontal="true"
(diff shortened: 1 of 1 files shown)


AndyScherzinger (changes requested):
@mahibi I suggest #2850 since it also resolves buttons not getting resized below 48dp in very small devices yet fluently wrapping the call buttons on any display ratio.

mahibi:
> @mahibi I suggest #2850 since it also resolves buttons not getting resized below 48dp in very small devices yet fluently wrapping the call buttons on any display ratio.

sure, i'm open to other solutions. answered in the PR..

AndyScherzinger (approved):

AndyScherzinger:
The other approach will need some polishing, so I think we should go with yours until the other one works in all scenarios 👍

PR #2850, opened 2023-03-09
Fix horizontal (=landscape) call design button layouting/sizing

AndyScherzinger:
Alternative to #2848

* ensures buttons are always 48dp (for accessibility reasons
* wraps the buttons into more lines if they don't fit in one line
* places the raise-hand button according to the "button bar" of the call buttons

### DIFF
--- a/app/src/main/res/layout/call_activity.xml
+++ b/app/src/main/res/layout/call_activity.xml
@@ -54,7 +54,8 @@
                 android:gravity="center"
                 android:numColumns="2"
                 android:scrollbars="vertical"
-                android:stretchMode="columnWidth" />
+                android:stretchMode="columnWidth"
+                tools:listitem="@layout/call_item"/>
 
             <FrameLayout
                 android:id="@+id/selfVideoViewWrapper"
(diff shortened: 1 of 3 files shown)


mahibi:
for me this currently looks like:

voice | video  
------- | -------- 
grafik | grafik

so for voice call the controls are too high?
and the raise hand button.. well i was not yet satisfied with it's placement, but it didn't change with this PR?

AndyScherzinger:
I didn't change the container, size-wise while it seems the buttons aren't vertically centered while they should be, I'll check that next week.

mahibi:
comparison of master and current state of this PR:

master, videocall, all buttons shown | master, videocall, other buttons hidden  
------- | -------- 
grafik | grafik

this PR, videocall, all buttons shown | this PR, videocall, other buttons hidden  
------- | -------- 
grafik | grafik

AndyScherzinger:
@mahibi changed the size from 48dp to 40dp which should be around the size is has been on master (which has been dynamic), so should be fine now.

mahibi (approved):

PR #2891, opened 2023-03-23
WIP. replace Controllers with Activities for location+geocoding

mahibi:
WIP

### 🏁 Checklist

- [ ] 📅 Milestone is set


AndyScherzinger:
@mahibi I think you would need to pop them off the back-stack after navigating away from them which you can do programmatically. I can't say from a navigation perspective on how many routes that would affect where you would need to implement it.

tobiasKaminsky:
Is this working when you remove "noHistory"?
I assume that this will also not work, as on restore only controller/activity is restored.
So to have also such a popup, you would need to store this as in info in "onSaveInstance" for this activity and then upon restore check and recreate/show dialog manually.

mahibi:
> @mahibi I think you would need to pop them off the back-stack after navigating away from them which you can do programmatically. I can't say from a navigation perspective on how many routes that would affect where you would need to implement it.

I now made this with FLAG_ACTIVITY_CLEAR_TOP when going from geocoding view to location view.

> Is this working when you remove "noHistory"?
I assume that this will also not work, as on restore only controller/activity is restored.
So to have also such a popup, you would need to store this as in info in "onSaveInstance" for this activity and then upon restore check and recreate/show dialog manually.

It worked for when the app was in background and came back. But it failed for screen rotation so now i used savedInstanceState to store relevant attributes.

PR #2911, opened 2023-03-29
Replace Controller with Activity for profile

mahibi:
this PR replaces controller with Activity, see https://github.com/nextcloud/talk-android/issues/1551

### 🏁 Checklist

- [ ] 📅 Milestone is set


tobiasKaminsky:
Tested it a bit, and not sure if this is because of this change
- adding phone number -> save -> does not work
- rotate leads to very large header, barely visible info



Other than that it works fine and great to see that slowly conductor can be removed :tada:

PR #2925, opened 2023-04-05
fix(i18n): Adjust links to transifex

stefan-niedermann:
https://help.nextcloud.com/t/transifex-announcement-web-applications-domain-change/159445

**Open Question:** @tobiasKaminsky @AndyScherzinger do we need also to adjust the `host` property in `.tx/config` for automated translations?


AndyScherzinger (approved):

tobiasKaminsky:
> The API URL does not change.

So nothing to do there :)

PR #2960, opened 2023-04-21
Fix to open conversation list when navigating back 

### LINKED ISSUE #2959: after chat was opened by notification: fail to go back to conversation list
### Steps to reproduce

open chat by notification
navigate back

### Expected behaviour

conversation list opens

### Actual behaviour

white screen is shown

16.1.0 Alpha 08

mahibi:
Fix to open conversation list when navigating back after chat was opened by notification

fix #2959 which was when working on https://github.com/nextcloud/talk-android/issues/1551

created https://github.com/nextcloud/talk-android/issues/2961 to migrate away from onBackPressed

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -222,6 +222,7 @@
             android:name=".chat.ChatActivity"
             android:theme="@style/AppTheme"
             android:launchMode="singleInstance"
+            android:noHistory="true"
             android:screenOrientation="portrait" />
 
         <activity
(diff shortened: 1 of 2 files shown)


AndyScherzinger (approved):

mahibi:
If somehow this solution turns out to be problematic, another approach can be to use the TaskStackBuilder:

https://developer.android.com/reference/androidx/core/app/TaskStackBuilder

https://stackoverflow.com/questions/36912325/why-do-we-use-the-taskstackbuilder

PR #2976, opened 2023-04-25
Fix call feature parsing

AndyScherzinger:
~~Parse value as String which should work for old values (Strings) and new Values (Boolean)~~

This PR fixes bugs that were introduced with https://github.com/nextcloud/talk-android/pull/2963/files
Basically replacing `as? String` (which would result in null) with `toString()` 

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/data/user/model/User.kt
+++ b/app/src/main/java/com/nextcloud/talk/data/user/model/User.kt
@@ -43,30 +43,9 @@ data class User(
     var scheduledForDeletion: Boolean = FALSE
 ) : Parcelable {
 
-    fun getMaxMessageLength(): Int {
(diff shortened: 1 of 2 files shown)


mahibi:
@AndyScherzinger so this was caused by https://github.com/nextcloud/talk-android/pull/2963

I'm not sure if using `Any` is the best solution but couldn't find anything better for LoganSquare. LoganSquare doesn't support polymorphic types, see https://github.com/bluelinelabs/LoganSquare/issues/11

In the long term, i wonder if switching to https://github.com/square/moshi would be worth it?

AndyScherzinger:
Maybe, or maybe the generic collection would need to be replaced with a more specific set of objects? Can't say if that is possible with LoganSquare ?

mahibi (approved):

Issue #1553, opened 2021-08-26
release process automation


mahibi:
The release process of release candidates and final versions must be automated.
(Currently only alpha versions are released automated. see https://github.com/nextcloud/talk-android/issues/962 )

- [ ] check with @tobiasKaminsky if solutions could be the same like for the files app

mahibi:
because it's still not automated recently a [typo happened](https://github.com/nextcloud/talk-android/pull/3001) which causes some unsightly workarounds.

so +1 for this issue

tobiasKaminsky:
For incrementing version you can check https://github.com/nextcloud/notes-android/blob/main/fastlane/common.Fastfile

And in general use also fastlane, like on Notes/Files.
We can do this together for first steps.

PR #3001, opened 2023-05-05
bump version to 17.1.0 Alpha 01

mahibi:
### 🏁 Checklist

- [ ] 📅 Milestone is set


AndyScherzinger (approved):

mahibi:
This version bump contains a mistake with the minorVersion.
So versionCode for 17.1 mistakenly became 17.10.
This cannot be changed because versions always have to increase in order to be deployed via gplay.
Because of this, the versionName must be modified for all 17.1.x versions while using the wrong versionCode.
With version 18.x everything is normal again.

So +1 to increase the importance of https://github.com/nextcloud/talk-android/issues/1553

PR #3008, opened 2023-05-09
Issue 2957 , "Fix transitions between views"

### LINKED ISSUE #2957: Fix transitions between views.
Now that [conductor is mostly being replaced](https://github.com/nextcloud/talk-android/issues/1551), new transitions need to be added to the switch between activities.

rapterjet2004:
fixes #2957 

Short video of the fade effect (sorry for the blur)
https://github.com/nextcloud/talk-android/assets/69230048/b8e98db6-d89e-4344-b706-1d7ed6513599

In each of the below files
- SettingsActivity
- ChatActivity
- ProfileActivity
- ConversationsListActivity
- ConversationInfoActivity
- LocationPickerActivity
- GeocodingActivity

Basically, I imported Fade, Window, and ActivityOptions and I added in each activities `onCreate()`,
```
with(window) {
            requestFeature(Window.FEATURE_CONTENT_TRANSITIONS)
            // set set the transition to be shown when the user enters this activity
            enterTransition = Fade()
            // set the transition to be shown when the user leaves this activity
            exitTransition = Fade()

        }
```

 and I replaced every `startActivity(intent)` with
`startActivity(intent,ActivityOptions.makeSceneTransitionAnimation(this).toBundle() )`

### 🚧 TODO

- [x] get it working on my end
- [ ] get it reviewed

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -152,17 +152,17 @@
         <activity
             android:name=".activities.FullScreenImageActivity"
             android:configChanges="orientation|keyboardHidden|screenSize"
-            android:theme="@style/FullScreenImageTheme"></activity>
+            android:theme="@style/FullScreenImageTheme"/>
 
         <activity
             android:name=".activities.FullScreenMediaActivity"
             android:configChanges="orientation|keyboardHidden|screenSize"
-            android:theme="@style/FullScreenMediaTheme"></activity>
+            android:theme="@style/FullScreenMediaTheme"/>
 
         <activity
             android:name=".activities.FullScreenTextViewerActivity"
             android:configChanges="orientation|keyboardHidden|screenSize"
-            android:theme="@style/FullScreenTextTheme"></activity>
+            android:theme="@style/FullScreenTextTheme"/>
 
         <activity
             android:name=".activities.TakePhotoActivity"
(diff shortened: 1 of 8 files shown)


AndyScherzinger:
see https://github.com/nextcloud/talk-android/issues/2957#issuecomment-1541508027 the transition between activities would match the navigational pattern of forward/backwards and thus slide in or out.

mahibi:
I agree it should be a horizontal slide in/out depending on the direction of navigation.

I gave this branch a short try and made these findings:

- Also the transition are affected by

  android:launchMode="singleInstance"
  android:noHistory="true"
  
  in AndroidManifest.xml
  
          <activity
              android:name=".chat.ChatActivity"
              android:launchMode="singleInstance"
              android:noHistory="true"
              android:screenOrientation="portrait"
              android:theme="@style/AppTheme" />
  
  I want to remove these two lines as they cause some other problems. So for this branch they cause weird UI glitches for me. You can just remove them for now for testing.

- i changed transition to Slide just to have a look -> Strange that the default Slide transition from android transition is vertical?? But didn't have a look how to change...

- the "Fade" transition was really horrible on my redmi phone. Slide was okay..

rapterjet2004:
Got it, deleted both lines from the manifest and cleared some trivial but annoying warnings about empty XML bodies.

mahibi:
As deleting the 2 lines will cause other bugs for now, i suggest to not change it yet.
I will work on this in the next days and then create a PR where i delete the lines and provide a new solution that these lines were supposed to solve.

So until then the transitions are also not fixed.
Sorry again for causing work that was not necessary in the end..

PR #3050, opened 2023-05-24
Fixing activity transitions

### LINKED ISSUE #2957: Fix transitions between views.

rapterjet2004:
Fixes #2957 

Signed-off-by: Julius Linus julius.linus@nextcloud.com

Transitions are <-- --> Slide behavior, nothing special just using the default `Slide(Gravity.START)` or `Slide(Gravity.END)`

I think I hit everything, I checked with `ctrl + shift + r`. I didn't do anything with `AccountVertificationController` because that still used conductor or anything that used `browserIntent` or anything with `context.startActivity`

I only tested it out on my emulator, so I'm not sure if it's consistent on different devices, but I think it should be. 

https://github.com/nextcloud/talk-android/assets/69230048/959d9251-16d0-4c7e-9a11-1a109654368a

### 🚧 TODO
- [x] get it working
- [ ] get it reviewed

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
@@ -80,8 +80,8 @@ open class BaseActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
         window.requestFeature(Window.FEATURE_ACTIVITY_TRANSITIONS)
-        window.enterTransition = Slide(Gravity.RIGHT)
-        window.exitTransition = Slide(Gravity.LEFT)
+        window.enterTransition = Slide(Gravity.END)
+        window.exitTransition = Slide(Gravity.START)
         super.onCreate(savedInstanceState)
     }
 
--- a/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/BaseActivity.kt
@@ -25,8 +25,11 @@ package com.nextcloud.talk.activities
 import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Bundle
+import android.transition.Slide
 import android.util.Log
+import android.view.Gravity
 import android.view.View
+import android.view.Window
 import android.view.WindowManager
 import android.webkit.SslErrorHandler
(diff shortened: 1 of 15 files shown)


mahibi:
Now transitions are the same on all my devices :+1: 
However for me the transitions doesn't feel quite smooth. 
Seems like some views are handled seperately (for example second 7 in the video: the notification settings are already in place and all other elements are animated.)
Other messengers slide the whole screen most of the times which is much nicer in my opinion. Any chance this you could give this a try @rapterjet2004 ?

rapterjet2004:
Ok so I got the transitions to work across I believe every screen. It was a small bug with how Slide() requires all views to be set visible at compile time I think. On a side note the views in Conversation Info and Settings use [this library](https://github.com/yarolegovich/MaterialPreferences) which hasn't been updated in 3 years. It seems harmless right now but maybe we should switch to native eventually? The UI looks kinda dated IMO.

AndyScherzinger:
Yeah, getting rid of material preferences library and just doing it plain native would be desirable 👍

AndyScherzinger:
Yet that should be handled via a separate PR of course 😊👍

AndyScherzinger (approved):

rapterjet2004:
I did some reading, and experimented a bit with delays and setting start and end animation speeds separately and decided that just setting both to have a transition speed of 50ms instead of the default makes the transition appear a bit less flashy, but perhaps more responsive? I only tried this on my emulator. But I took a video of it here. I'm not sure if it also appears this way on other devices. 

before

https://github.com/nextcloud/talk-android/assets/69230048/c2f37d9a-7bb0-4f10-a1d5-fce6b45d5ba9


after

https://github.com/nextcloud/talk-android/assets/69230048/8a58ac06-2cba-4945-be25-4abbd08ce3e4

Not sure what's causing the conversations to flicker upon the end transition, that's just them refreshing right?

mahibi:
Thanks for having a look into it @rapterjet2004 
On real devices the behaviour of master still feels better to me.
I suggest we will do the next release without this PR and gain some feedback. So maybe we can have another look at it in the future.
Maybe @AndyScherzinger has an opinion on this?

mahibi (changes requested):
i suggest to not merge it as it doesn't feel smooth on real devices.
Any other options welcome...

PR #3064, opened 2023-06-01
File util fixes

mahibi:
### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
@@ -126,7 +126,9 @@ class UploadAndShareFilesWorker(val context: Context, workerParameters: WorkerPa
 
             initNotificationSetup()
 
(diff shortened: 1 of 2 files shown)


tobiasKaminsky (approved):

AndyScherzinger (approved):

firmianay:
@mahibi Hello, I saw this path traversal vulnerability and its fix. It seems to be a little problematic, because getCanonicalPath returns a path string that is not terminated with a slash, which means that external input may still be copied to siblings 
 of cacheDir, rather than just its children, such as "data/data/package/cachexxxx"
https://codeql.github.com/codeql-query-help/java/java-partial-path-traversal/

nickvergessen:
hi @firmianay please use our security/bugbounty program at https://hackerone.com/nextcloud/ to report related issues.
We are tracking the matter now already, but would be cool for future issues.

firmianay:
Thanks for the reply. I am registering a hackerone account. If you think it is necessary, I can create this issue on it.

mahibi:
as nickvergessen mentioned it's already tracked now but for future security issues just create an issue there. Thanks for your help :+1:

PR #3127, opened 2023-06-28
Update Chinese translation

Doraemon-byte:
Updated one line translation:
nc_remove_participant should be translated as "移除参与者" instead of "远程参与者".


AndyScherzinger:
Translations are managed though transifex. I fixed it here: https://app.transifex.com/nextcloud/nextcloud/translate/#zh_CN/android-talk/182105411?q=text%3ARemove

So tonight the translation in the repo will be updated.

Thanks for raising this and providing the PR, I am going to close it since the bot overrides the translations every night, via transifex.

PR #3190, opened 2023-07-14
Websocket improvements2

mahibi:
### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketConnectionHelper.java
+++ b/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketConnectionHelper.java
@@ -59,11 +59,15 @@ public class WebSocketConnectionHelper {
     }
 
     @SuppressLint("LongLogTag")
-    public static synchronized WebSocketInstance getWebSocketInstanceForUserId(long userId) {
-        WebSocketInstance webSocketInstance = webSocketInstanceMap.get(userId);
+    public static synchronized WebSocketInstance getWebSocketInstanceForUser(User user) {
+        WebSocketInstance webSocketInstance = webSocketInstanceMap.get(user.getId());
 
         if (webSocketInstance == null) {
-            Log.e(TAG, "No webSocketInstance found for user " + userId);
+            Log.d(TAG, "No webSocketInstance found for user " + user.getDisplayName() +
+                " (userId:" + user.getId() + ")");
+        } else {
+            Log.d(TAG, "Existing webSocketInstance found for user " + user.getDisplayName() +
+                " (userId:" + user.getId() + ")");
         }
 
         return webSocketInstance;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/talk/webrtc/WebSocketConnectionHelper.java
```diff
@@ -59,19 +59,24 @@ public WebSocketConnectionHelper() {
     }
 
     @SuppressLint("LongLogTag")
-    public static synchronized WebSocketInstance getWebSocketInstanceForUserId(long userId) {
-        WebSocketInstance webSocketInstance = webSocketInstanceMap.get(userId);
+    public static synchronized WebSocketInstance getWebSocketInstanceForUserId(User user) {
```
SystemKeeper, line 62: Should this be renamed to `getWebSocketInstanceForUser`?

app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt
```diff
@@ -138,7 +139,7 @@ class WebSocketInstance internal constructor(
             isConnected = false
             messagesQueue = ArrayList()
         }
-        restartWebSocket()
```
SystemKeeper, line 141: I am not sure about this one. When we hit the `onFailure` event we call `closeWebSocket`. Doesn't it make sense to then restart the connection? In my tests `onClosed` was not called when we close the webSocket ourselves?!

app/src/main/res/layout/item_custom_incoming_voice_message.xml
```diff
@@ -93,7 +93,7 @@
 
             <SeekBar
                 android:id="@+id/seekbar"
-                android:layout_width="match_parent"
+                android:layout_width="0dp"
```
SystemKeeper, line 96: Not sure this belongs in this PR?

app/src/main/res/layout/item_custom_outcoming_voice_message.xml
```diff
@@ -79,7 +79,7 @@
             <SeekBar
                 android:id="@+id/seekbar"
                 style="@style/Nextcloud.Material.Outgoing.SeekBar"
-                android:layout_width="match_parent"
+                android:layout_width="0dp"
```
SystemKeeper, line 82: Same here

SystemKeeper (approved):
I left some comments above. Only the one about the removed close might be important, I fail to see the point of the removal at this moment...

For reference, it might make sense to delay the reconnection to the websocket a bit to not DOS the server and to also not run in a infinite loop on the device (not for this PR, something for later). On iOS it's done here: 
https://github.com/nextcloud/talk-ios/blob/e3cfc029aafa0f874abb48f8a4f80d223005c672/NextcloudTalk/NCExternalSignalingController.m#L208-L222

Otherwise working nicely in the test-cases we discussed!

app/src/main/res/layout/item_custom_incoming_voice_message.xml
mahibi: it's from the commit "fix lint warnings". it's not related to this PR, but otherwise the checks in this PR will fail. But yeah i could fix it in another PR..

app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt
mahibi: insterted the restartWebSocket() again and will do the delay reconnection part later :+1:

PR #3202, opened 2023-07-24
Waveform SeekBar for Voice Messages

rapterjet2004:
Signed-off-by: Julius Linus <julius.linus@nextcloud.com>

### 🖼️ Screenshots
Working demo

https://github.com/nextcloud/talk-android/assets/69230048/de3f4cbc-2efe-4d11-9475-1c65a4278945

### 🚧 TODO
- [x] Finish AudioUtils 
- [x] Finish WaveformSeekbar Component
- [ ] Implement the WaveformSeekbar into the voice message ViewHolders
- [ ] Test and get it reviewed

### 🏁 Checklist

- [ ] 📅 Milestone is set


rapterjet2004:
@AndyScherzinger 
@mahibi 

The PR is functional but there are some problems with the loading time of AudioUtils that I could use your input on. 

Currently, to process a voice message takes a non-trivial time of about 1-5 seconds, at least in my emulator. To work around this, I made it so that the audio file can still be played as the waveform is loaded asynchronously. The resulting array is stored in each ChatMessage object so It won't have to be loaded again, unless the conversation is exited. Here's a video showing what I mean: 


https://github.com/nextcloud/talk-android/assets/69230048/0fe0bd1d-c2cb-4cf7-94fb-94843288b248



Alternatively if this is too jarring,  I could implement a loading state until the waveform is ready, and have a time limit of about 2 seconds if It takes too long, else canceling the process and just showing a regular SeekBar. 



--- 
In my opinion, an ideal long term solution would be to have each ChatMessage object received from the server with an array of data, if the message is a voice message. It would simply be a matter of feeding that array into the current ViewHolder's WaveForm SeekBar. That way, every message would have a WaveForm look to it as soon as the messages are loaded from the server.

But I believe that would require an additional JSON field, which makes things slightly complicated.  Although, I believe that's how Signal implemented it.

In my previous MicInputCloud component, I had a thread using AudioRecord to record the speakers voice's input, which is how I managed to get it to respond depending on the mic data. Hypothetically, I could also create a array representing that audio data on the fly, and use AudioUtils to decode voice messages recorded before this feature.

app/src/main/java/com/nextcloud/talk/ui/WaveformSeekBar.kt
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            restore()
+
+        }
+    }
+
+    companion object {
+        private val TAG: String? = WaveformSeekBar::class.simpleName
+        private const val DEFAULT_BAR_WIDTH: Int = 5
```
nickvergessen, line 113: If we change this to 2, we get many more "bars"? I would prefer to have much more, on your video it looks like we have 20 bars while the telegram screenshot is around 50 or something?

rapterjet2004: No problem, how's this? 
image


I set the bar width to 1dp and made the bars 50. While the bar width is const, the number of bars is set by the `shrinkFloatArray` function in `AudioUtils`. The function `audioFileToFloatArray` usually returns a large array with hundreds of values, so I created this helper function to shrink it. This code is located on line 874 in `setWaveform` in `ChatActivity`
```
        val r = AudioUtils.audioFileToFloatArray(file)
        val f = AudioUtils.shrinkFloatArray(r, 50)
```

AndyScherzinger: I would suggest slightly wider and also higher -comparing it to the seekbar's "nob" if that is doable

rapterjet2004: image

I made ```MAX_HEIGHT_DIVISOR: Float = 10f``` to ```MAX_HEIGHT_DIVISOR: Float = 6.5f```

AndyScherzinger:
@rapterjet2004 can you 
* make the bars slightly wider (1->2)
* make the gap between the bars slightly larger
* put the time below the wave, else it resizes back and forth if you move between voice messages

rapterjet2004:
@AndyScherzinger 
How's this?

image

```val barGap: Float = (usableWidth - waveData.size * DEFAULT_BAR_WIDTH) / (waveData.size - 1).toFloat()```
The barGap is calculated with the containers width and the default bar width, so I changed the code to have outcoming messages have 38 bars, and incoming messages to have 50

AndyScherzinger:
@rapterjet2004 looks good 👍 Nice polishing. I tend to say let's put the voice message time below the bar at the end so it finds its place between the bar and the message timestamp. Having it next to the message timestamp kind of links it to it while it is related to the voice message.

Then I think we are good to go. Might need to optimize the colors in the future to ensure the contrast is high enough between the 2 bar colors.

AndyScherzinger:
Merged to not keep it open for longer, since it works fine 👍 Really nice work! 🎉 

@rapterjet2004 could you maybe do some additional polishing:
* larger gap between the bars (double or triple of the current one I would say)
* center the time below the play/pause button
* apply same color to time as the play/pause button
* apply same text size to play time as the message timestamp

PR #3216, opened 2023-07-27
show call time and hint for one-hour calls

### LINKED ISSUE #3175: Show hint when call is running since 1 hour
Overview at https://github.com/nextcloud/spreed/issues/6212

* Design: https://github.com/nextcloud/spreed/issues/6212#issuecomment-1623340533
  Maybe just a small popover on the call time display:
  > Note the call is running since 1 hour already.

  Nothing else suggesting a break or wrap up, just pointing out the time.

mahibi:
resolve #3175

### 🖼️ Screenshots

Call screen with call time | Info when call is running since one hour (faked for 20s)
-------- | -------- 
grafik | grafik

### 🚧 TODO

- [x] check capability
- [x] fix multiple calls of fetchSignalingSettings -> fix via #3219 as it might need some attention if this causes other bugs
- [x] no hardcoding for strings
- [x] hide time when no participants in call
- [x] set value for 1 hour (instead 10s)

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -36,6 +36,7 @@ import android.content.Intent
 import android.content.IntentFilter
 import android.content.res.Configuration
 import android.graphics.Color
+import android.graphics.Typeface
 import android.graphics.drawable.Icon
 import android.media.AudioAttributes
 import android.media.MediaPlayer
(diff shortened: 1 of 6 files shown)


AndyScherzinger (approved):

AndyScherzinger:
Looks good to me, however besides showing a `Toast` I would also suggest to have it persisted on the call screen showing it in some kind of a banner-styled way.

What do you think @jancborchardt @nimishavijay ?

jancborchardt (commented):
Discussed improvements would be:
- [x] Bold the timer when it is more than 1 hour (from Andy)
- [ ] Try to only make the Toast timeout/vanish start from when you actually saw it. Not when the app is in the background for example.

PR #3234, opened 2023-08-02
Follow up fixes for the Waveform SeekBar

rapterjet2004:
Fixes [this](https://github.com/nextcloud/talk-android/pull/3202#issuecomment-1662220530)

- Larger Bar gap
- Centered time below play/pause button
- Time is the same color as the play/pause button

Unfortunately, I wasn't able to fix the play time size. It's puzzling as on paper they should be equal 
image
I believe that there is some underlying logic morphing the timestamp in `fun showReactions` located in the `com.nextcloud.talk.adapters.messages.Reaction`, but I didn't want to mess with it. 

---
I did some refactoring, so now the number of bars is calculated based off the width of the SeekBar itself. And by extension, the bar gap should also be calculated based off the width of the SeekBar, therefore no matter the size, it should always maintain a good ratio.

### Before
[follow-up-waveform-seekbar-fixes-before.webm](https://github.com/nextcloud/talk-android/assets/69230048/fd942f91-e72b-4d86-892b-ab2a56708d93)

### After
[follow-up-waveform-seekbar-fixes-after.webm](https://github.com/nextcloud/talk-android/assets/69230048/f547ee99-5dba-41f7-95c7-fc7a999512d9)

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -894,12 +894,7 @@ class ChatActivity :
             message.isDownloadingVoiceMessage = true
             adapter?.update(message)
             CoroutineScope(Dispatchers.Default).launch {
-                val bars = if (message.actorDisplayName == conversationUser?.displayName) {
-                    NUM_BARS_OUTCOMING
-                } else {
-                    NUM_BARS_INCOMING
-                }
-                val r = AudioUtils.audioFileToFloatArray(file, bars)
+                val r = AudioUtils.audioFileToFloatArray(file)
                 message.voiceMessageFloatArray = r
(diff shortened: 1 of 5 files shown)


AndyScherzinger (approved):
Look better on my high-res device 👍

AndyScherzinger:
Just give you an idea what I was aiming for (more or less)

|Talk with this PR|instagramm|
|---|---|
|Screenshot_20230802_211525|Screenshot_20230802_211606|

In a way something between the original code and this PR might be the sweet spot and close to insta. Not saying you should spend that time, just in case you want to :)

AndyScherzinger:
So feel free to merge @rapterjet2004 (or I can since the analysis check is failing (in general, everywhere))

PR #3235, opened 2023-08-03
Migration away from getScheme() calls

### LINKED ISSUE #3042: Migrate away from viewThemeUtils.getScheme calls
Follow-up for #3041

to migrate all occurrences of `viewThemeUtils.getScheme` to element-specifc theming functions either in `android-common` if generic solutions or in `TalkSpecificViewThemeUtils` if specific to Talk.

rapterjet2004:
fixes #3042 

image

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -1080,13 +1080,11 @@ class CallActivity : CallBaseActivity() {
     }
 
     private fun getSpotlightView(): SpotlightView? {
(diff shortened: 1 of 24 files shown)


AndyScherzinger:
@rapterjet2004 for which theming call would you actually need it?

AndyScherzinger:
Well, you could hard-code it because it is Talk specific. I guess one is incoming messages and the other is outgoing messages? So from a theming perspective one could argue that you actually want two distinct methods to reflect the theming by having "business functions", less generic, not specific and by name clear when to use which one.

PR #3249, opened 2023-08-14
Migrating away from StoreBox to DataStore

### LINKED ISSUE #3043: get rid of StoreBox
[StoreBox](https://github.com/martino2k6/StoreBox) is used to store [SharedPreferences](https://developer.android.com/training/data-storage/shared-preferences). StoreBox is outdated as well as usage of SharedPreferences is not recommended anymore.

StoreBox also causes Lint warnings when using @KeyByResource annotation: "Resource IDs will be non-final by default in Android Gradle Plugin version 8.0, avoid using them as annotation attributes" 

We should replace StoreBox with modern solutions like datasore:
https://developer.android.com/topic/libraries/architecture/datastore

rapterjet2004:
fixes #3043 

image

### 🚧 TODO

- [x] Implement AppPreferenceImpl with DataStore to migrate away from StoreBox
- [x] Refactor Settings activity to use Kotlin Flows rather than OnPreferenceValueChangedListeners
- [x] Test Out
   - [x] Theme
   - [x] Call Notification
   - [x] Messages Notification
   - [x] Screen Lock
   - [x] Screen Security
   - [x] Incognito keyboard
   - [x] Phone number integration
   - [x] Read Status
   - [x] Typing Status
   - [x] Proxy Settings
   - [x] Privacy
   - [x] Source Code
   - [x] License
   - [x] Fix duplicate calls to a single flow observer

- [ ] Get it reviewed

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -164,6 +164,8 @@ configurations.all {
 
 dependencies {
     implementation 'androidx.preference:preference-ktx:1.2.1'
+    implementation 'androidx.datastore:datastore-core:1.0.0'
+    implementation 'androidx.datastore:datastore-preferences:1.0.0'
     detektPlugins("io.gitlab.arturbosch.detekt:detekt-formatting:1.23.1")
 	
     implementation fileTree(include: ['*'], dir: 'libs')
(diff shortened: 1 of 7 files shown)


mahibi:
when quickly testing i noticed 2 things:

- why is typingStatusFlow observer triggered e.g. when i for example activate incognito keyboard?

- sometimes the checkboxes are jumping back to their old state right after using it.

mahibi:
:+1:

PR #3261, opened 2023-08-22
avoid falsely shown "Account scheduled for deletion" info

mahibi:
...by setting the correct context.

In removeCurrentAccount
the

WorkManager.getInstance(this) (introduced with commit 943286dd)

must be

WorkManager.getInstance(applicationContext)

because otherwise the worker is dropped when the activity is finished and might not do it's work beforehand (race condition?). This results in users that keep the

scheduledForDeletion = true

so the next time someone tries to login with this user,

userManager.checkIfUserIsScheduledForDeletion(loginData.username!!, baseUrl!!).blockingGet()

in WebViewLoginController#parseAndLoginFromWebView

is true and

ApplicationWideMessageHolder.getInstance().messageType is set to

ApplicationWideMessageHolder.MessageType.ACCOUNT_SCHEDULED_FOR_DELETION

which results in the "The account is scheduled for deletion, and cannot be changed" message in ServerSelectionController.

This is just a guess and i was not able to reproduce it.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
merged for now but not sure if it fully fixes the issue

Issue #3270, opened 2023-08-24
Media playback does not retain state.


parneet-guraya:
**Is your feature request related to a problem? Please describe.**

Player does not retain Playback state and resets.
In `FullScreenMediaActivity.kt` the Playback state is not perserved (since player is initialized and released between onStart() and onStop() respectively)

**Describe the solution you'd like**

Defining class level variables to hold player state(not in viewmodel or onSaveInstanceState() api since activity is not being recreated on configuration changes) ->
- initialize with state while initializing the player
- store the state while releasing the player

Additional context

https://github.com/nextcloud/talk-android/assets/111801812/a1189087-05c0-4935-a61b-831301c1a943

There are two things needed to be handled -->
- Player's `playWhenReady`(chooses whether to play or not when is in Ready state. Tapping play/pause change it's value true/false respectively and thus player choose to play or pause accordingly) state.
- Player's elapsed time (how much playback is elapsed)

PR #3272, opened 2023-08-24
Save media playback state

parneet-guraya:
Solves: #3270 

### Before

https://github.com/nextcloud/talk-android/assets/111801812/64f95423-ce92-4707-ba97-9b1d47efdff0

### After

https://github.com/nextcloud/talk-android/assets/111801812/4c3c856d-59ae-4065-b31c-7b78ca61b18a

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/FullScreenMediaActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/FullScreenMediaActivity.kt
@@ -4,9 +4,11 @@
  * @author Marcel Hibbe
  * @author Andy Scherzinger
  * @author Ezhil Shanmugham
+ * @author Parneet Singh
  * Copyright (C) 2021 Andy Scherzinger <info@andy-scherzinger.de>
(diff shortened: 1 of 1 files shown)


parneet-guraya:
Also, here we're setting listener to the player but the implementation we're setting is blank since we're not overriding any behaviour (`Player.Listener` interface have declared functions with empty body so we can have with implementation without actually implementing any function.)

Is there any particular reason for that? Instead we can get rid of the redundant setting of listener

mahibi (approved):

mahibi:
thanks @parneet-guraya :+1:

mahibi:
@Smarshal21 you were invited to the nextcloud android team so now you should e.g. be able to create pull requests right inside the nextcloud repo if you would like to

mahibi:
@parneet-guraya  you were invited to the nextcloud android team so now you should e.g. be able to create pull requests right inside the nextcloud repo if you would like to

PR #3307, opened 2023-09-08
Show accept call button in chat

### LINKED ISSUE #3260: show accept call button in chat
Currently when being in a chat and a call is started, only a message "xy started a call" is shown. To join it, the icons in the upper right have to be clicked.

There should also a button right inside the chat to join a call. 
Maybe 2 buttons for audio/video, or maybe one button with a popup to choose between audio/video.

rapterjet2004:
fixes #3260 

## Current UI
| Dark | Light | Guest |
|--------|--------|--------|
| image| image| image| 

## Functionality
[show-accept-button-1.webm](https://github.com/nextcloud/talk-android/assets/69230048/f798634a-b69a-4d46-bffd-6004731632fe)

- Both join audio and video buttons
- Gets deleted if call ends, persists despite state or adapter changes
- Themed

## TODO
- [x] Complete Functionality
   - [x] Get it working for guests
   - [x] Get it working for users
- [x] Finish Implementing the mention chip
- [x] detekt 😠 
- [x] review
- [ ] deal with git commit signatures

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -121,6 +121,8 @@ import com.nextcloud.talk.R
 import com.nextcloud.talk.activities.BaseActivity
 import com.nextcloud.talk.activities.CallActivity
 import com.nextcloud.talk.activities.TakePhotoActivity
+import com.nextcloud.talk.adapters.messages.CallStartedMessageInterface
+import com.nextcloud.talk.adapters.messages.CallStartedViewHolder
 import com.nextcloud.talk.adapters.messages.CommonMessageInterface
 import com.nextcloud.talk.adapters.messages.IncomingLinkPreviewMessageViewHolder
 import com.nextcloud.talk.adapters.messages.IncomingLocationMessageViewHolder
@@ -268,7 +270,8 @@ class ChatActivity :
     VoiceMessageInterface,
     CommonMessageInterface,
     PreviewMessageInterface,
-    SystemMessageInterface {
+    SystemMessageInterface,
+    CallStartedMessageInterface {
 
     var active = false
 
@@ -376,6 +379,8 @@ class ChatActivity :
     var typingTimer: CountDownTimer? = null
     var typedWhileTypingTimerIsRunning: Boolean = false
     val typingParticipants = HashMap<String, TypingParticipant>()
+
+    var callStarted = false
 
     private val localParticipantMessageListener = object : SignalingMessageReceiver.LocalParticipantMessageListener {
         override fun onSwitchTo(token: String?) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -4232,10 +4268,38 @@ class ChatActivity :
...
+                    unreadChatMessage.actorDisplayName = name
+                    unreadChatMessage.actorType = recent.actorType
+                    unreadChatMessage.timestamp = chatMessageList[0].timestamp
+                    unreadChatMessage.message = null
+                    adapter?.addToStart(unreadChatMessage, false)
+                    callStarted = true
+                }
+            } // add CallStartedMessage with id -2
+            ChatMessage.SystemMessageType.CALL_ENDED,
+            ChatMessage.SystemMessageType.CALL_MISSED,
```
mahibi, line 4294: ChatMessage.SystemMessageType.CALL_TRIED is missing?

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -3189,6 +3205,26 @@ class ChatActivity :
...
                             }
 
+                            try {
+                                val mostRecentCallSystemMessage = adapter?.items?.first {
+                                    it.item is ChatMessage &&
+                                        (it.item as ChatMessage).systemMessageType != null &&
+                                        (it.item as ChatMessage).systemMessageType!!.ordinal <=
+                                        ChatMessage.SystemMessageType.CALL_TRIED.ordinal &&
+                                        (it.item as ChatMessage).systemMessageType!!.ordinal >=
+                                        ChatMessage.SystemMessageType.CALL_STARTED.ordinal
```
mahibi, line 3215: i don't like this solution with ordinal too much.
if anything changes in the enum order it could introduce strange bugs someday.
better would be to check for every single CALL_ item?

rapterjet2004: Yeah, you're right. It would be better to have it not rely on any arbitrary placement of ordinals.

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
rapterjet2004: I left it out intentionally, because the docs had this
- `call_tried` - You tried to call {user}

which, I think, implies that the current user started a call and therefore wouldn't need a CallStartedMessage to be rendered in the first place.

mahibi: start a call from mobile
do not accept on web
hang up again on mobile

--> the CallStartedMessage remains visible but the call is over.
Including ChatMessage.SystemMessageType.CALL_TRIED would fix this.

AndyScherzinger (approved):

PR #3319, opened 2023-09-16
FIxed deletion of voice, video, image, contact and location messages

Smarshal21:
### 🖼️ Screenshots
🏚️ Before | 🏡 After

RESULTS

### 🚧 TODO

- [X] Fix the deletion of voice, video, image, contact and location messages
### 🏁 Checklist

- [X] ⛑️ Tests (unit and/or integration) are included or not needed
- [X] 🔖 Capability is checked or not needed 
- [X] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [X] 📅 Milestone is set
- [X] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)


Smarshal21:
@AndyScherzinger PTAL

AndyScherzinger (changes requested):
Breaks the calculation of the delete action by overwritting it. So the original place in the ChatActivity needs to be adapted yet not as simply as done here since there are further rules in place which determine if a chat message can be deleted.

So I think this needs some further work and alignment with @nickvergessen

mahibi:
@Smarshal21 
in ChatActivity in method isShowMessageDeletionButton, delete the following lines in the return statement:

message.hasFileAttachment() -> false       # deleting this line should allow deletion of voice, video, image, contact
OBJECT_MESSAGE == message.message -> false    # deleting this line should allow deletion of location

AndyScherzinger:
@mahibi 

wouldn't 

```
message.hasFileAttachment() -> false       # deleting this line should allow deletion of voice, video, image, contact
```

also apply for any other file type then? like "documents" ? Just asking to be sure.

mahibi:
One more change i suggest:

Shared files from the Nextcloud instance are handled differently. Currently with a long click it's only possible to "Open in files app", other options like deletion are not available.
To change this, i suggest to remove the
`onMessageViewLongClick`
in `PreviewMessageViewHolder`
and instead of  

```
                clickView!!.setOnLongClickListener { l: View? ->
                    onMessageViewLongClick(message)
                    true
                }
```
just open the MessageActionDialog via

```
                clickView!!.setOnLongClickListener { l: View? ->
                    previewMessageInterface!!.onPreviewMessageLongClick(message)
                    true
                }
```
This would allow reactions and all the other options just like in web.

@Smarshal21 just in case anything is confusing for you, just let me know if i should do these commits

Smarshal21:
No i would like to handle that

app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
```diff
@@ -128,7 +128,7 @@ abstract class PreviewMessageViewHolder(itemView: View?, payload: Any?) :
                     )
                 }
                 clickView!!.setOnLongClickListener { l: View? ->
-                    onMessageViewLongClick(message)
+                    previewMessageInterface!!.onPreviewMessageLongClick(message)
```
mahibi, line 131: as the method 
`private fun onMessageViewLongClick(message: ChatMessage)..`
is no longer used it can be deleted.
https://github.com/nextcloud/talk-android/pull/3319/files#diff-d097b26ec3d15c322bf4cf32bdb89ba37a00fd8c41bde2109848e68b0e559729R283

app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
```diff
@@ -104,7 +104,7 @@ class MessageActionsDialog(
                 hasUserActorId(message) &&
                 currentConversation?.type != ConversationType.ROOM_TYPE_ONE_TO_ONE_CALL
         )
-        initMenuDeleteMessage(showMessageDeletionButton)
+        initMenuDeleteMessage(ChatMessage.MessageType.SYSTEM_MESSAGE != message.getCalculateMessageType())
```
mahibi, line 107: this needs to be reverted to be `showMessageDeletionButton` again so the logic from ChatActivity to show/hide the deletion button is used.
https://github.com/nextcloud/talk-android/blob/288ec688349dd55e885c25f330c79c21e232277a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt#L4172

app/src/main/java/com/nextcloud/talk/adapters/messages/PreviewMessageViewHolder.kt
Smarshal21: Done

app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
Smarshal21: Done

mahibi:
@Smarshal21 please make sure to always sign your commits
https://github.com/nextcloud/talk-android/blob/master/CONTRIBUTING.md#sign-your-work

If you're using Android Studio to commit, you can use this checkbox in the UI once you set it up in git-config:
https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git

image

Smarshal21:
> @Smarshal21 please make sure to always sign your commits https://github.com/nextcloud/talk-android/blob/master/CONTRIBUTING.md#sign-your-work
> 
> If you're using Android Studio to commit, you can use this checkbox in the UI once you set it up in git-config: https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git
> 
> image

Sorry for that 
Certainly, I will ensure to do so from now on 👍🏻

mahibi:
to fix the missing sign-offs in this branch retrospectively,
you could rebase on origin master and then squash all your commits into one commit which is signed off.
After this, you can do a force push to overwrite the origin commits.
Just let me know if you have any questions here.

PR #3333, opened 2023-09-21
Save filter state

tobiasKaminsky:
Remember filter states

### 🏁 Checklist

- [x] 📅 Milestone is set


app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -272,6 +276,50 @@ class ConversationsListActivity :
 
         showSearchOrToolbar()
     }
+    fun filterConversation() {
+        filterState[FilterConversationFragment.UNREAD] = appPreferences.isUnreadFilterEnabled
```
AndyScherzinger, line 280: as discussed, should be bound to the account instead of a global setting

AndyScherzinger (approved):
Works for me 👍

PR #3336, opened 2023-09-22
make text in preview selectable, so that user can copy it

tobiasKaminsky:
Needs testing, currently in train and no local docker 😇

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
works.
thx

PR #3342, opened 2023-09-25
Error 429 "Too Many Requests" is now handled

### LINKED ISSUE #3340: handle '429 Too Many Requests' for phonebook sync
show an info to the user when  '429 Too Many Requests' is returned for for phonebook sync.

rapterjet2004:
- fixes #3340 

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
@@ -32,6 +32,7 @@ import android.net.Uri
 import android.os.RemoteException
 import android.provider.ContactsContract
 import android.util.Log
+import android.widget.Toast
 import androidx.core.content.ContextCompat
 import androidx.core.os.ConfigurationCompat
 import androidx.work.Data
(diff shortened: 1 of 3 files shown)


mahibi (approved):

app/src/main/java/com/nextcloud/talk/jobs/ContactAddressBookWorker.kt
```diff
@@ -143,9 +144,22 @@ class ContactAddressBookWorker(val context: Context, workerParameters: WorkerPar
                     }
 
                     override fun onNext(foundContacts: ContactsByNumberOverall) {
-                        val contactsWithAssociatedPhoneNumbers = foundContacts.ocs!!.map
-                        deleteLinkedAccounts(contactsWithAssociatedPhoneNumbers)
-                        createLinkedAccounts(contactsWithAssociatedPhoneNumbers)
+                        when (foundContacts.ocs?.meta?.statusCode) {
+                            HTTP_CODE_TOO_MANY_REQUESTS -> {
+                                Toast.makeText(
```
mahibi, line 149: fine for me to use a toast here as we have no view. 
just an edge case here which might never(?) be triggered..
otherwise we prefer snackbar

PR #3349, opened 2023-09-27
use MVVM for Geocoding

Smarshal21:
**Implemented LiveData for GeoLocation RecyclerView**

https://github.com/nextcloud/talk-android/assets/99678760/0e772e34-5f77-4253-8f2e-40e8b90b5e84

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/GeocodingAdapter.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/GeocodingAdapter.kt
@@ -20,6 +20,7 @@
 
 package com.nextcloud.talk.adapters
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.view.LayoutInflater
 import android.view.View
@@ -35,6 +36,8 @@ class GeocodingAdapter(private val context: Context, private var dataSource: Lis
     interface OnItemClickListener {
         fun onItemClick(position: Int)
     }
+
+    @SuppressLint("NotifyDataSetChanged")
     fun updateData(data: List<Address>) {
         this.dataSource = data
         notifyDataSetChanged()
--- a/app/src/main/java/com/nextcloud/talk/adapters/GeocodingAdapter.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/GeocodingAdapter.kt
@@ -29,11 +29,15 @@ import androidx.recyclerview.widget.RecyclerView
 import com.nextcloud.talk.R
 import fr.dudie.nominatim.model.Address
 
-class GeocodingAdapter(private val context: Context, private val dataSource: List<Address>) :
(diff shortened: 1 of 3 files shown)


Smarshal21:
@AndyScherzinger PTAL and suggest me the changes 
Excuse me for my poor internet connection 😅😅

mahibi:
there are still the methods
`searchLocation()`
and
`executeGeocodingRequest()`
in the GeocodingActivity and they are still used by `onResume()`

remove these 2 methods and take the viewModel for it.

As it's not necessary to trigger the geocoding on screen rotation in onResume (viewModel will survive..), maybe there should be some check to only do this when the adapter has no items. So it should only happen initially when the query is passed from LocationPickerActivity. 

As 
`query = intent.getStringExtra(BundleKeys.KEY_GEOCODING_QUERY)`
is always set again in onCreate in GeocodingActivity, this can lead to a bug when rotating the screen. 
After searching for other queries, you would still set the old query that was initially passed to GeocodingActivity after rotating the screen.
Lets try to initially set the query that was passed to GeocodingActivity into the viewModel. Then forget about the bundleKey or ignore it whenever the query is already set inside the viewModel.
And where ever you need to get the query, get it from the viewModel, for example in onCreateOptionsMenu.

Smarshal21:
> there are still the methods `searchLocation()` and `executeGeocodingRequest()` in the GeocodingActivity and they are still used by `onResume()`
> 
> remove these 2 methods and take the viewModel for it.
> 
> As it's not necessary to trigger the geocoding on screen rotation in onResume (viewModel will survive..), maybe there should be some check to only do this when the adapter has no items. So it should only happen initially when the query is passed from LocationPickerActivity.
> 
> As `query = intent.getStringExtra(BundleKeys.KEY_GEOCODING_QUERY)` is always set again in onCreate in GeocodingActivity, this can lead to a bug when rotating the screen. After searching for other queries, you would still set the old query that was initially passed to GeocodingActivity after rotating the screen. Lets try to initially set the query that was passed to GeocodingActivity into the viewModel. Then forget about the bundleKey or ignore it whenever the query is already set inside the viewModel. And where ever you need to get the query, get it from the viewModel, for example in onCreateOptionsMenu.

Done 👍🏻
PTAL and suggest me the changes

mahibi:
@Smarshal21 i pushed some changes, hope that's okay for you.
e.g. to avoid using instanceState, as this can be handled by the view model.
If you have any questions regarding my commit just let me know.

Smarshal21:
> @Smarshal21 i pushed some changes, hope that's okay for you. e.g. to avoid using instanceState, as this can be handled by the view model. If you have any questions regarding my commit just let me know.

yeah i got 
Its fine

rapterjet2004 (approved):

PR #3396, opened 2023-10-19
Bump room to 2.6.0 

AndyScherzinger:
and rename paramter to match super class in case of use of names parameters

### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
Tested successfully

Issue #1578, opened 2021-09-10
Handshake failed


eachtime:
## Steps to reproduce
1. open a newly installed (via f-droid) talk app on android 9 phone
2. try to connect to a self hosted Nextcloud instance (next.hitch.fr)

### Expected behaviour
switch to an account connection screen

### Actual behaviour
Error message : Handshake failed

## Talk app

**Talk app version:** 11.3.1
**Talk android app version:** 12.2.1

## Server configuration

**Web server:** Apache2
SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1 -TLSv1.2

**Nextcloud Version:** 21.0.4

> Every other apps like notes, deck working well on the android 9 device and talk seems to work fine with an IOS device with the same instance of Nextcloud

> I don't have any Apache or Nextcloud log when the error pops up

mahibi:
seems https://github.com/nextcloud/talk-android/commit/9f296996abb465fd223ae089230d64c5369420a2 still needs to be ported to the current master (see https://github.com/nextcloud/talk-android/issues/672 )
But anyway that should only fix the handshake error for android 7.
- [ ] fix ssl handshake exception for android 7

This issue here is about android 9 so there's another problem. Maybe already closed issues about handshake errors could contain some useful hints. https://github.com/nextcloud/talk-android/issues?q=is%3Aissue+handshake+failed+is%3Aclosed

eachtime:
I didn't find any useful hints in previous issues but I don't understand what does this solution refer to. https://github.com/nextcloud/talk-android/issues/603#issuecomment-526914966
maybe someone can enlighten me on this.

I also tried with the last apk provided on github and I have the same issue.

FaribaKhandani:
I checked with two below devices and couldn't log in because of handshake failed error 
Samsung galaxy s9+ sm-g965f  - Android 10 (API level 29)
motorola one (xt1941-4) - Android 9 (API level 28)

mahibi:
i was able to reproduce this with Android 9 in Emulator when connecting to sermo.

PR #3416, opened 2023-10-30
feature/2331/saveFiles

FaribaKhandani:
### The feature of saving media in personal storage space

**Access to feature**
Access to this feature is possible through the menu in full screen media mode or long click on media.

**process of saving**
In this feature, if the media is already stored in the cache, it is copied to the download folder, and if the media is not available in the cache, it is first downloaded to the cache and then copied to the download folder.

**Permitions**
Due to the limitations of memory access, if there is no prior permission from the user's side, the user's permission will be asked first, and like other features in this app, the media will be saved in the next attempt.
According to the descriptions mentioned in the related issue, every time an attempt is made to save the media, a warning is sent to the user that other programs have access to the saved media.

###  TODO

- [Physical device Test ] After fetch this project and running it through Android Studio on two types of physical device , the "Handshake" error occurred when logging into the account.
For this reason, this feature has only been tested on virtual Android devices and needs to be tested by the QA team on a physical device.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/FullScreenImageActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/FullScreenImageActivity.kt
@@ -26,6 +26,8 @@
 
 package com.nextcloud.talk.activities
 
+import android.annotation.SuppressLint
+import android.content.DialogInterface
 import android.content.Intent
 import android.os.Bundle
 import android.util.Log
@@ -33,6 +35,7 @@ import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup.MarginLayoutParams
+import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.FileProvider
 import androidx.core.view.ViewCompat
@@ -41,20 +44,28 @@ import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
 import androidx.core.view.updateLayoutParams
 import androidx.core.view.updatePadding
+import androidx.work.Data
+import androidx.work.OneTimeWorkRequest
+import androidx.work.WorkInfo
+import androidx.work.WorkManager
 import com.google.android.material.snackbar.Snackbar
 import com.nextcloud.talk.BuildConfig
 import com.nextcloud.talk.R
 import com.nextcloud.talk.databinding.ActivityFullScreenImageBinding
+import com.nextcloud.talk.jobs.SaveFileToStorageWorker
+import com.nextcloud.talk.ui.theme.ViewThemeUtils
 import com.nextcloud.talk.utils.BitmapShrinker
 import com.nextcloud.talk.utils.Mimetype.IMAGE_PREFIX_GENERIC
 import pl.droidsonroids.gif.GifDrawable
 import java.io.File
+import java.util.concurrent.ExecutionException
 
 class FullScreenImageActivity : AppCompatActivity() {
     lateinit var binding: ActivityFullScreenImageBinding
     private lateinit var windowInsetsController: WindowInsetsControllerCompat
     private lateinit var path: String
     private var showFullscreen = false
+    lateinit var viewThemeUtils: ViewThemeUtils
 
     override fun onCreateOptionsMenu(menu: Menu): Boolean {
         menuInflater.inflate(R.menu.menu_preview, menu)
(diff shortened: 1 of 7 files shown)


mahibi:
Hi @FaribaKhandani,

thanks for your PR! With my test device it works as expected :+1: 

Regarding the handshake failed error: can you comment on https://github.com/nextcloud/talk-android/issues/1578 by adding the devices and versions you used?

Regarding the code i have some minor remarks, i will just add some comments.
If you would like and have some time, feel free to solve my remarks. Otherwise just let me know if i should do it, that's totally fine.

app/src/main/java/com/nextcloud/talk/activities/FullScreenImageActivity.kt
```diff
@@ -188,6 +228,51 @@ class FullScreenImageActivity : AppCompatActivity() {
         }
     }
 
+    @SuppressLint("LongLogTag")
+    private fun saveImageToStorage(
+        fileName: String,
+        funToCallWhenDownloadSuccessful: (() -> Unit)
```
mahibi, line 234: the second parameter (funToCallWhenDownloadSuccessful) is not used so it could also just be deleted.
So just call the method with saveImageToStorage(fileName) and delete the observer.

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -2018,6 +2021,63 @@ class ChatActivity :
...
+        val fileId = message.selectedIndividualHashMap!!["id"]
+
+        // check if download worker is already running
+        val workers = WorkManager.getInstance(
+            context
+        ).getWorkInfosByTag(fileId!!)
+        try {
+            for (workInfo in workers.get()) {
+                if (workInfo.state == WorkInfo.State.RUNNING || workInfo.state == WorkInfo.State.ENQUEUED) {
+                    Log.d(TAG, "Download worker for $fileId is already running or scheduled")
```
mahibi, line 2050: just replace "Download Worker" with SaveFileToStorageWorker, so the log message makes sense again.

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -4101,12 +4161,55 @@ class ChatActivity :
...
+        if (permissionUtil.isFilesPermissionGranted()) {
+            saveImageToStorage(message ,  false){
+            }
+        }
+        else{
+             UploadAndShareFilesWorker.requestStoragePermission(this@ChatActivity)
+        }
+    }
+
+     fun showWarningDialog(message: ChatMessage) {
```
mahibi, line 4181: maybe better find a name that tells a bit more about what this dialog does. For example "showSaveToStorageWarning"

app/src/main/java/com/nextcloud/talk/jobs/SaveFileToStorageWorker.kt
```diff
@@ -0,0 +1,94 @@
...
+            setProgressAsync(Data.Builder().putInt(PROGRESS, 0).build())
+        }
+
+        try {
+            val fileName = inputData.getString(KEY_FILE_NAME)
+            val sourceFilePath = inputData.getString(KEY_SOURCE_FILE_PATH)
+
+            val cacheFile = File(sourceFilePath)
+
+            val downloadDirectory = "storage/self/primary/Download"
```
mahibi, line 55: i think it would be better to use MediaStore to store the files.
https://developer.android.com/training/data-storage/shared

app/src/main/java/com/nextcloud/talk/jobs/SaveFileToStorageWorker.kt
```diff
@@ -0,0 +1,94 @@
...
+            val destFile = File( downloadDirectory, fileName)
+
+            Log.d("amds" , cacheFile.path)
+            Log.d("amdd" , destFile.path)
+
+            if (destFile.exists()){
+                destFile.delete()
```
mahibi, line 62: deleting is not the best choice here i guess. Instead there could be number appended. But that's maybe out of scope for this PR.

FaribaKhandani: Hello 

All of your remarks are fixed and pushed .
 Media Store approach for saving file is added (previous way for saving files is commented within the code )

Please let me know what is the next step?

mahibi:
Hi Fariba,

thanks again for the PR and also thanks to solve the remarks.
I did some more modifications, basically:
- remove some unnecessary params for saveImageToStorage
- make some methods private
- do some code formatting

Please feel free to ask if any of my changes are unclear.

Issue #3324, opened 2023-09-19
Only 3-4 group participants show up


jakobroehrl:
### Steps to reproduce

Open a conversation with a lot of participants

### Expected behaviour

see all of them in a scrollable list

### Actual behaviour

there is no scroll effect, so you see only 3-4

### Device brand and model

Samsung and Pixel

### Android version

13

### Nextcloud Talk app version

17.1.0

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

Dtrieb:
Same here.

Smarshal21:
i implemented it here - https://github.com/nextcloud/talk-android/pull/3441

jakobroehrl:
@mahibi 
This is still an issue, maybe you could look over it?

PR #3441, opened 2023-11-10
Made the Participant List in Conversation Info Scrollable

Smarshal21:
Resolves - https://github.com/nextcloud/talk-android/issues/3324
Made the Participant List in Conversation Info Scrollable, allowing users to navigate through the list of participants. 

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
@@ -441,7 +441,7 @@ class ConversationInfoActivity :
         binding.recyclerView.layoutManager = layoutManager
         binding.recyclerView.setHasFixedSize(true)
         binding.recyclerView.adapter = adapter
-        binding.recyclerView.isNestedScrollingEnabled = false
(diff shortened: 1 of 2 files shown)


Smarshal21:
**is it better if i change the colour of the background for better visibility ??**

mahibi:
Without to test yet: You want to make the list itself scrollable while the screen itself is also scrollable? I don't think this is the is the right solution.

I can't reproduce the bug, but will have a closer look soon.

Smarshal21:
> Without to test yet: You want to make the list itself scrollable while the screen itself is also scrollable? I don't think this is the is the right solution.
> 
> I can't reproduce the bug, but will have a closer look soon.

Yeah , but what exactly might be the solution for this??

mahibi:
> Yeah , but what exactly might be the solution for this??

not sure yet, first i would suggest some research if `isNestedScrollingEnabled = false` could fail for some reason and if/why the list is cropped. Could even be that recycler view is not the best choice here..

mahibi:
will close this one in favor of https://github.com/nextcloud/talk-android/pull/3762

PR #3443, opened 2023-11-14
Added a ProgressBar to display while uploading Pics/Files

### LINKED ISSUE #2500: Add a placeholder/progress bar for uploading pics/files
Sometimes the user Don't know if the photo was sucessful because the app gives no response especially on slow connections

Smarshal21:
Resolves #2500 

## Overview
This pull request introduces a ProgressBar to enhance the user experience during file uploads. The ProgressBar provides visual feedback on the upload progress, making it more intuitive for users.

### Upload Functionality:

- Modified the upload function to handle ProgressBar visibility.
- The ProgressBar is set to VISIBLE during the upload process and GONE when the upload is complete.

### WorkManager Integration:

- Utilized WorkManager for managing the background file upload task.
- Enqueued unique work to ensure proper handling of the upload task.

### Observing WorkManager State:

- Implemented an observer for the WorkManager to track the state of the file upload.
- The ProgressBar is hidden once the upload is completed, and appropriate notifications are displayed.

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -143,6 +143,7 @@ import com.nextcloud.talk.extensions.loadAvatarOrImagePreview
 import com.nextcloud.talk.jobs.DownloadFileToCacheWorker
 import com.nextcloud.talk.jobs.ShareOperationWorker
 import com.nextcloud.talk.jobs.UploadAndShareFilesWorker
+import com.nextcloud.talk.jobs.UploadAndShareFilesWorker.Companion.upload
 import com.nextcloud.talk.location.LocationPickerActivity
 import com.nextcloud.talk.messagesearch.MessageSearchActivity
 import com.nextcloud.talk.models.domain.ConversationModel
@@ -3254,12 +3255,15 @@ class ChatActivity :
 
         try {
             require(fileUri.isNotEmpty())
-            UploadAndShareFilesWorker.upload(
+            upload(
+                context,
                 fileUri,
                 room,
                 currentConversation?.displayName!!,
                 metaData
-            )
+            ) { showProgressBar: Boolean ->
+                binding.idprogressbar.visibility = if (showProgressBar) View.VISIBLE else View.GONE
+            }
         } catch (e: IllegalArgumentException) {
             context.resources?.getString(R.string.nc_upload_failed)?.let {
                 Snackbar.make(binding.root, it, Snackbar.LENGTH_LONG)
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -3256,7 +3256,6 @@ class ChatActivity :
         try {
             require(fileUri.isNotEmpty())
             upload(
-                context,
                 fileUri,
                 room,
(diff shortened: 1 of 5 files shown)


mahibi (changes requested):
I like the callback approach, however i think the solution is not yet fully developed.

I agree that we need to show the loading progress in the chat, but we should introduce placeholder messages for that, which is a bit more work.

With the current solution, the loading bar now floats above all messages which is a bit irritating, especially when uploads take longer. It is not obvious that it is the upload progress but it feels more like there are problems loading the chat.

Smarshal21:
> Placeholder messages seem to be necessary for this one. There should be a shimmer effect in place while the file is loading. We're already using this [loading library](https://github.com/elye/loaderviewlibrary) in some of our xml.
> 
> We could have it so that when the file message is sent, a temporary placeholder message is added to the chat list, before being removed when the file message arrives from the server.
> 
> This also allows us a bit of freedom when it comes to design. We could implement a custom Nextcloud themed loading animation in the future, if we ever need to.

what would be better @mahibi @rapterjet2004  ? use the existing shimmer which appears while loading the chat or a custom shimmer for uploading ?

mahibi:
> > Placeholder messages seem to be necessary for this one. There should be a shimmer effect in place while the file is loading. We're already using this [loading library](https://github.com/elye/loaderviewlibrary) in some of our xml.
> > We could have it so that when the file message is sent, a temporary placeholder message is added to the chat list, before being removed when the file message arrives from the server.
> > This also allows us a bit of freedom when it comes to design. We could implement a custom Nextcloud themed loading animation in the future, if we ever need to.
> 
> what would be better @mahibi @rapterjet2004 ? use the existing shimmer which appears while loading the chat or a custom shimmer for uploading ?

i would say a loading circle (spinner) inside a placeholder message.

Fyi i rebased on master and fixed minor merge conflicts

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -3212,12 +3213,14 @@ class ChatActivity :
...
             require(fileUri.isNotEmpty())
-            UploadAndShareFilesWorker.upload(
+            upload(
                 fileUri,
                 room,
                 currentConversation?.displayName!!,
                 metaData
-            )
+            ) { showProgressBar: Boolean ->
+                binding.idprogressbar.visibility = if (showProgressBar) View.VISIBLE else View.GONE
```
rapterjet2004, line 3222:
```suggestion
                binding.idprogressbar.visibility = if (showProgressBar) View.VISIBLE else {
                    chatViewModel.refreshChatParams(
                        setupFieldsForPullChatMessages(
                            true,
                            globalLastKnownFutureMessageId,
                            true
                        )
                    )
                    View.GONE
                } 
```

Breaks with the recent refactoring of `ChatActivity`. Insert this line to fix it, so that uploaded files show up after sending.

rapterjet2004 (changes requested):

mahibi:
Thanks for your work @Smarshal21 :+1: and sorry we did not find the time to bring this PR forward together. 
The approach of the PR is not how the feature should be implemented so it will close it.
Meanwhile, for the next major release there is a solution to have temporary messages for text. This approach should also be used for other message types. However before developing we should have a clear roadmap on it.

Thanks again for your efforts!

PR #3466, opened 2023-11-27
Bugfix/noid/fix save file dialog

mahibi:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_20231127_142107 geaendert | Screenshot_20231127_141507 geaendert

### 🚧 TODO

- [ ] material3??

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/dialog/SaveToStorageDialogFragment.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/dialog/SaveToStorageDialogFragment.kt
@@ -26,9 +26,6 @@ import android.app.Dialog
 import android.content.DialogInterface
 import android.os.Bundle
 import android.util.Log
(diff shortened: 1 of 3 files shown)


AndyScherzinger (approved):
Don't think it's about m3, I think we usually dynamically theme all colorings including the text, I think.

AndyScherzinger (approved):

mahibi:
> Don't think it's about m3, I think we usually dynamically theme all colorings including the text, I think.

yeah m3 is unrelated here, just thought about it as the textsizes of the dialog are different when open it via context menu with longclick on the file, or when you open it via FullScreen*Activity.

image


image


I temporarily applied m3 design to FullScreen*Activity which worked, but iirc then the colors of the appbar menu are wrong in dark mode..

PR #3506, opened 2023-12-11
DatabaseStorageModule not working correctly

rapterjet2004:
Room was overwriting old data in `ArbitraryStorage`, because there was a conflict detected in `ArbitraryStoragesDao`
`@Insert(onConflict = OnConflictStrategy.REPLACE)`. This was because the composite primary key wasn't including the conversation token as well. `primaryKeys = ["accountIdentifier", "key"]`

## ToDo
- [x] Create a Migration from Dual Key to Triple Key
- [x] Test Migrations
image

image

### Before
https://github.com/nextcloud/talk-android/assets/69230048/1ce4d174-4d09-4418-86c2-fcdfad60f942

### After
https://github.com/nextcloud/talk-android/assets/69230048/741bbf93-a13a-4768-8c9a-ea756b7fa816

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/schemas/com.nextcloud.talk.data.source.local.TalkDatabase/9.json
+++ b/app/schemas/com.nextcloud.talk.data.source.local.TalkDatabase/9.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 9,
-    "identityHash": "666fcc4bbbdf3ff121b8f1ace8fcbcb8",
+    "identityHash": "250a3a56f3943f0d72f7ca0aac08fd1e",
     "entities": [
       {
         "tableName": "User",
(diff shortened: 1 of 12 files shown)


AndyScherzinger:
@rapterjet2004 without having taken a deeper look into it, it seems you didn't implement a DB migration to "9" which is needed I think either an auto migration or an actual one.

app/src/main/java/com/nextcloud/talk/data/source/local/Migrations.kt
```diff
@@ -124,4 +131,29 @@ object Migrations {
         // Change the table name to the correct one
         db.execSQL("ALTER TABLE ArbitraryStorage_dualPK RENAME TO ArbitraryStorage")
     }
+
+    fun migrateToTriplePrimaryKeyArbitraryStorage(db: SupportSQLiteDatabase) {
+        db.execSQL(
+            "CREATE TABLE ArbitraryStorage_triplePK (" +
```
mahibi, line 137: why is the order now changed to:
accountIdentifier
value
\"key\"
object

it makes more sense to use this order? (so the primary key fields come first):
accountIdentifier
\"key\"
object
value

app/schemas/com.nextcloud.talk.data.source.local.TalkDatabase/9.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 9,
-    "identityHash": "666fcc4bbbdf3ff121b8f1ace8fcbcb8",
+    "identityHash": "250a3a56f3943f0d72f7ca0aac08fd1e",
```
mahibi, line 5: shouldn't there be a file called 10.json instead to modify 9.json ?

rapterjet2004: Yeah that's weird, I'll just add it in manually I guess. Seems to pass the test.

image

mahibi:
i suggest to not merge this PR in favor of https://github.com/nextcloud/talk-android/issues/3569
extending the primary key might make it more difficult to implement data migrations etc.

Anyway good catch Julius :+1: :+1:

PR #3533, opened 2023-12-19
🤖Android 1️⃣4️⃣

AndyScherzinger:
Currently missing is a research on foreground service changes, see https://developer.android.com/about/versions/14/changes/fgs-types-required.

We currently seem to not be calling [startForeground()](https://developer.android.com/reference/android/app/Service#startForeground(int,%20android.app.Notification)) - so we should not be facing the corresponding security exception. However we do have some services in our manifest that might need a type attributed to it.

### 🚧 TODO

- [x] Test against Android 14

### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
@mahibi @sowjanyakch - fancy taking a look? I didn't have any issues on Android 13 but the important scenario is the behavior on 14 of course. The foreground aspect is quite important here and also an issue in terms of publishing to GPlay. @tobiasKaminsky and the team already went through this for the files app also had to provide a video to Google to proof the need for foreground services.

So just wanted to raise this to your attention to not run into a situation where publishing and sdk bumps are needed but then get blocked by Google. Hence I though better tackle this now than close to a new major release.

sowjanyakch (approved):
works on Android 14.

mahibi (approved):

mahibi:
i removed `foregroundServiceType="phoneCall"`
for NCFirebaseMessagingService, see https://github.com/nextcloud/talk-android/pull/3533/commits/e4a80429d3c820d7b7760f992442c77121e27c43

Tested on Android 14 and works, can be merged.

AndyScherzinger:
Thank you both for taking a look and for sporting the issue on the different flavor and fixing it 🙏

PR #3535, opened 2023-12-20
Handle audio focus and becoming noisy

parneet-guraya:
- [x] Handle Audio focus
- [x] Handle Audio becoming noisy

### Handle audio focus -->

The player will handle the audio focus shifts. Basically It would:
- Pause the already playing media when this player is started.
- Or If some other media starts this player would pause itself.
This would avoid two media's playing at the same time.

https://github.com/nextcloud/talk-android/assets/111801812/da9d2fcc-f389-42d9-b075-95be6b8dadf8

https://github.com/nextcloud/talk-android/assets/111801812/978668e7-1082-40bd-9c00-33a29dbd2417

### Handle audio becoming noisy -->

> When a headset is unplugged or a Bluetooth device disconnected, the audio stream automatically reroutes to the built-in speaker. If you listen to music at a high volume, this can be a noisy surprise.
> 
> Users usually expect apps that include a music player with onscreen playback controls to pause playback in this case. Other apps, like games that don't include controls, should keep playing. The user can adjust the volume with the device's hardware controls.
>
[See More](https://developer.android.com/guide/topics/media/platform/output#becoming-noisy)

https://github.com/nextcloud/talk-android/assets/111801812/c6a18a00-d9b3-4b3c-907a-4f5531812687

You can see here after turning of bluetooth the playback pauses.

https://github.com/nextcloud/talk-android/assets/111801812/bc25ccfc-6717-4eef-8c3b-aaada64300a8

Here after removing earphones playback pauses.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -42,7 +42,9 @@ import android.database.Cursor
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
+import android.media.AudioFocusRequest
 import android.media.AudioFormat
+import android.media.AudioManager
 import android.media.AudioRecord
 import android.media.MediaPlayer
 import android.media.MediaRecorder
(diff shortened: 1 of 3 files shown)


mahibi (approved):

mahibi:
tested and works.
great improvements @parneet-guraya :+1: :+1:

PR #3550, opened 2023-12-27
Handle forward message back button

### LINKED ISSUE #3538: back button in "Forward message" closes the app
## reproduce

- go to a chat
- long click on a message and select "forward"
- click back button

## expected
old chat is opened again

## actual
app closes

sowjanyakch:
Resolves #3538 

This pull request handles the back button of forward message. 

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
@@ -206,7 +206,11 @@ class ConversationsListActivity :
 
     private val onBackPressedCallback = object : OnBackPressedCallback(true) {
         override fun handleOnBackPressed() {
-            finishAffinity()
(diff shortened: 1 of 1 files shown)


mahibi:
It's the correct section where the fix should be made.
However finishAffinity() makes sense most of the time and finish() should only be only used when a message forwarding is used.

ConversationsListActivity can be opened from different activities. For example from AccountVerificationActivity after initially adding an account.
So after adding an account and opening ConversationsListActivity, the back back button should close the app instead to open the AccountVerificationActivity again.

I think this should work (not tested): Just check if forwardMessage is true. If it is, call finish, otherwise finishAffinity.

An alternative might be to use special flags for the intents that define how the stack behaves, but just checking forwardMessage should do the job with low effort.

mahibi (approved):

Issue #2373, opened 2022-09-16
grouping for upload notifications


mahibi:
notifications should be grouped when uploading many files at the same time.
https://developer.android.com/develop/ui/views/notifications/group

hint: notification behaviour may be different for android versions and manufacturers..

PR #3567, opened 2024-01-09
Grouping upload notification

parneet-guraya:
Fix: #2373 

### What this PR includes? 

- [x] Grouping notification
- [x] pushing new notification (when failure) so that the progress indicator (part of the original notification) won't show

### 🖼️ Screenshots

https://github.com/nextcloud/talk-android/assets/111801812/9b3ba71f-d376-4e4a-994d-4d79eece4d47

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
@@ -268,7 +268,7 @@ class UploadAndShareFilesWorker(val context: Context, workerParameters: WorkerPa
         mNotifyManager?.cancel(notificationId)
         // summary notification would not get dismissed automatically if child notifications are cancelled programmatically
         // so check if only 1 notification left if yes then cancel it because that would be summary notification
-        if(getActiveUploadNotifications() == 1){
+        if (getActiveUploadNotifications() == 1) {
             mNotifyManager?.cancel(NotificationUtils.GROUP_SUMMARY_NOTIFICATION_ID)
         }
     }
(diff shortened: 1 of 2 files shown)


parneet-guraya:
Sometimes the first upload notification may show separately from group notification. A little demo to support this fact -->

Basically if second notification is pushed after >=400ms then the first one would be separate and rest of the notifications would be part of group and after this second notification time gap wouldn't matter new notifications would always be part of the group. (See video 1)

If second notification is pushed after <400ms then it would be the part of the group from beginning.

Video 1

https://github.com/nextcloud/talk-android/assets/111801812/41b6f8b7-bba0-4e3d-872e-8ae8352fdccd

Video 2

https://github.com/nextcloud/talk-android/assets/111801812/8de64d16-f2e0-4f20-97fc-1227c3329f72

You can also see in both of these cases a third notification is pushed after 4 seconds which is always the part of group which shows time gap doesn't matter here.

rapterjet2004:
> Sometimes the first upload notification may show separately from group notification. A little demo to support this fact -->

Yes, I've replicated it too. Is there anything you can do about this? Other than that, it looks good. Just fix the `ktlintChecks`. You can find them by running this on Windows. Also works for `detekt` too.

` ./gradlew ktlintCheck `

parneet-guraya:
> > Sometimes the first upload notification may show separately from group notification. A little demo to support this fact -->
> 
> Yes, I've replicated it too. Is there anything you can do about this? 

I think it is more of a behaviour than an issue. As I said if the second notification takes more than 400ms it would show the first one separate and the rest of them will go into group. One thing could be done if our worker can trigger notification quickly then it may solve it. That is on my to-do so will see if that works for us.

mahibi (approved):

mahibi:
> > > Sometimes the first upload notification may show separately from group notification. A little demo to support this fact -->
> > 
> > 
> > Yes, I've replicated it too. Is there anything you can do about this?
> 
> I think it is more of a behaviour than an issue. As I said if the second notification takes more than 400ms it would show the first one separate and the rest of them will go into group. One thing could be done if our worker can trigger notification quickly then it may solve it. That is on my to-do so will see if that works for us.

Can confirm this behaviour but it's okay in my opinion.
Worked for me so will merge..
Also: depending on devices, automatic grouping might already done by android.

Issue #3576, opened 2024-01-13
i18n: Need context for build flavour


rakekniven:
https://github.com/nextcloud/talk-android/blob/a3d77f75632941bab325b0a6e87fbc082df15cb6/app/src/main/res/values/strings.xml#L214

Please enlight us (the translators).

Add it to the sources.

AndyScherzinger:
It basically tells you the type of build. F-droid or. Play, debug or. Release, QA-build.

I Would say should be marked as translatable false.

AndyScherzinger:
But leave it to @mahibi to decide

AndyScherzinger:
@rakekniven how can context be provided for Android Strings?

mahibi:
> I Would say should be marked as translatable false.

sure, lets do this https://github.com/nextcloud/talk-android/pull/3579

PR #3579, opened 2024-01-15
set string nc_diagnose_build_flavor to translatable=false

mahibi:
fis https://github.com/nextcloud/talk-android/issues/3576

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
> Lgtm while I would rename it to diagnosis

well, all the new key for diagnosis screen begin with "nc_diagnose_*".
So change all of them? That might lead to frustration for translators as they already began to translate...

AndyScherzinger:
Than let's leave them as is 👍

AndyScherzinger (approved):

Issue #3572, opened 2024-01-12
18.0.1:  conversation popup: flip from horizontal to vertical kills app


jakobroehrl:
1. open a conversation
2. take a picture
3. tip send
4. the add to conversation popup appears:
![image](https://github.com/nextcloud/talk-android/assets/24530680/8875ad2c-afee-4c23-874d-155e9d4651ea)
7. flip from horizontal to vertical or vise versa
8. the app is killed or the conversation and the pic is lost

### Device brand and model

Pixel 4a

### Android version

14

### Nextcloud Talk app version

18.0.1

parneet-guraya:
I can reproduce this issue. picking this up.

PR #3674, opened 2024-02-29
Fix: File Attachment Preview Dialog crashes ChatActivity

parneet-guraya:
Fixes: #3572

### Before

https://github.com/nextcloud/talk-android/assets/111801812/a4a59bb8-5548-43f0-8a0b-58b48f28349d

### After

https://github.com/nextcloud/talk-android/assets/111801812/b0bf27d9-ac5e-4695-98fa-d2b25722ced3

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -4385,7 +4385,7 @@ class ChatActivity :
             val lon = data["longitude"]!!
             metaData =
                 "{\"type\":\"geo-location\",\"id\":\"geo:$lat,$lon\",\"latitude\":\"$lat\"," +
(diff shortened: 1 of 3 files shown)


parneet-guraya:
Hi @mahibi, could you give it a look?

The reason for crashing was that this dialog fragment does not have a default constructor. So, when config changes occured the activity does not know how to initialize a parameterized constructor and looks for empty constructor, hence the crash. Instead use empty constructor with arguments passed to it as bundle.

I have seen some more fragment implementations like this, will create an issue for that and fix it.

parneet-guraya:
@mahibi, Also I'm doubtful about using scroll view in a dialog. The buttons were going out of the screen, so I added the scroll view. Let me know if it's correct to use it here.

mahibi (approved):

mahibi:
:+1:

PR #3630, opened 2024-02-06
Preparing for federation - Developing data layer for chats

rapterjet2004:
Makes `ChatActivity` in pure MVVM. This creates a solid foundation of infrastructure for federation and for any future offline support. TBH this has been on the books for some time now, but I suppose federation gave me an excuse to finally dedicate some time to implementing it.

Essentially does this
1. Exposes the entirely of `ChatActivity` as an interface in `ChatRepository` with state handled in the `ChatViewModel`. In this way, handling errors from the server, doing checks upon JSON response, and developing a offline-first app become simpler.

- [x] Repository Interface
- [x] Repository Implementation
- [x] ViewModel Implementation
- [x] Activity Implementation
   - [x] Functionality
   - [x] Bugs
      - [x] pull chat messages duplicating (finally 🙏🏾 )
      - [x] old chat messages not showing up
      - [x] Crash on empty conversation
      - [x] Reactions causes crash? 

### TODO (followup PRs)

- [ ] Deal with Unread message popup not showing up 
- [ ] photos/videos not showing up after sending

Related
- #3569 
- #3529 

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [x] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #3675, opened 2024-02-29
Follow up fixes to ChatActivity w/ recent refactoring

rapterjet2004:
follow up to #3630 

- Popup bubbles now show up
- Fixed bugs with images
- added new string resource

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -193,7 +193,7 @@ import com.nextcloud.talk.ui.recyclerview.MessageSwipeActions
 import com.nextcloud.talk.ui.recyclerview.MessageSwipeCallback
 import com.nextcloud.talk.utils.ApiUtils
 import com.nextcloud.talk.utils.AudioUtils
-import com.nextcloud.talk.utils.SpreedFeatures
+import com.nextcloud.talk.utils.CapabilitiesUtil
 import com.nextcloud.talk.utils.ContactUtils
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -861,27 +861,17 @@ class ChatActivity :
...
                             }
 
-                            var lastAdapterId = 0
-                            if (adapter?.items?.size != 0) {
-                                val item = adapter?.items?.get(0)?.item
-                                if (item != null) {
-                                    lastAdapterId = (item as ChatMessage).jsonMessageId
+                            var lastAdapterId = getLastAdapterId()
+
+                            if (chatMessageList[0].jsonMessageId > lastAdapterId) { // Weird off by one bug
```
mahibi, line 866: this will destroy to fetch message from the past as processMessagesNotFromTheFuture won't be called when scrolling up.

Keep it like before will fix it:

```
     if (
          state.lookIntoFuture &&
          lastAdapterId != 0 &&
          chatMessageList[0].jsonMessageId > lastAdapterId
      ) {
          processMessagesFromTheFuture(chatMessageList)
      } else if (!state.lookIntoFuture) {
          processMessagesNotFromTheFuture(chatMessageList)
          collapseSystemMessages()
      }
```

mahibi (changes requested):
see comment about processMessagesNotFromTheFuture

mahibi (approved):

PR #3682, opened 2024-03-01
Renaming strings from circles to teams.

### LINKED ISSUE #3651: Renaming "Circles" to "Teams"
### Refs
- https://github.com/nextcloud/circles/issues/1505
- https://github.com/nextcloud/server/issues/42433

sowjanyakch:
Resolves #3651

This PR renames strings from circles to teams.

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
Renaming keys of strings will cause problems for the tooling around transifex translation service.
As written in the top of strings.xml: 
"- Never change the key of a translation. If it really has to be changed, delete it and create a new entry instead."

So the complete lines in strings.xml have to be deleted and a new line with a different key has to be added.
For the translators, this means they have to translate again, but they have do anyway because the value also changes.

Other than they, should be fine.

AndyScherzinger:
So from my POV PR should worjk since all Strings also have the ID changed, so this should then be just fine

mahibi (approved):

PR #3698, opened 2024-03-07
hide some features if federation is not available

mahibi:
hide some features that are not available (yet) for federation.

### 🏁 Checklist

- [x] 📅 Milestone is set


sowjanyakch (approved):
Looks good to me

PR #3712, opened 2024-03-13
fix to hide open conversations after returning from search

mahibi:
How to test:

pre-requirement:
There are open conversations

without this commit:
1. go to search screen
2. go back to conversation list -> all open conversations are listed below own conversations

with this commit:
1. go to search screen
2. go back to conversation list -> only own conversations are listed

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch (approved):
Tested and works fine.

Issue #3737, opened 2024-03-20
Dialog's Text Field does not adapt to dark/light modes and does not retain state.


parneet-guraya:
### Steps to reproduce

- go to settings.
- click on phone number integration toggle.

- See the dialog's text field have darker color even in the dark mode. This does not text properly.
- Also dialog does not retain configuration changes.

### Expected behaviour

- Text field should adapt to the system theme.
- Should retain state i.e Typed phone number

### Actual behaviour

- does not show proper text field color according to the system them (dark/light).
- does not retain state.

https://github.com/nextcloud/talk-android/assets/111801812/dd49615d-eea3-4536-a17f-5bfbb00eec2b

### Device brand and model

OnePlus 9RT

### Android version

14

### Nextcloud Talk app version

v 19.0.0 Alpha 03

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

PR #3738, opened 2024-03-20
Use DialogFragment for Phone Number Integration

parneet-guraya:
Fixes: #3737 

### 🖼️ Screenshots

https://github.com/nextcloud/talk-android/assets/111801812/d2f59153-8402-49fd-9eab-94624a3f81c6

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/settings/SettingsActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/settings/SettingsActivity.kt
@@ -34,10 +34,8 @@ import android.annotation.SuppressLint
 import android.app.KeyguardManager
 import android.content.Context
 import android.content.DialogInterface
-import android.content.DialogInterface.OnShowListener
 import android.content.Intent
 import android.content.pm.PackageManager
-import android.content.res.ColorStateList
 import android.graphics.PorterDuff
 import android.graphics.drawable.ColorDrawable
 import android.media.RingtoneManager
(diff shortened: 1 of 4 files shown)


mahibi:
Good question from @parneet-guraya copied from public chat:

"Hi 👋 . In some places we are using AlertDialog Builder to show a dialog instead of a DialogFragment. As a result neither they retain state nor they survive configuration changes. I was wondering if we should convert those to use DialogFragment as well?
Or should we keep the simpler ones (yes/no) as is (to avoid creatingseparate files but again they won't survive config change) and convert those which have some state to retain?"

mahibi:
> Good question from @parneet-guraya copied from public chat:
> 
> "Hi 👋 . In some places we are using AlertDialog Builder to show a dialog instead of a DialogFragment. As a result neither they retain state nor they survive configuration changes. I was wondering if we should convert those to use DialogFragment as well? Or should we keep the simpler ones (yes/no) as is (to avoid creatingseparate files but again they won't survive config change) and convert those which have some state to retain?"

I am still undecided. Maybe this behavior could be fixed when switching to more modern solutions like Jetpack Compose etc.?
So maybe it's worth it to keep it like it is until then, instead to create more xml layouts.
Anyone with opinions on this?

rapterjet2004 (approved):

parneet-guraya:
> I am still undecided. Maybe this behavior could be fixed when switching to more modern solutions like Jetpack Compose etc.? So maybe it's worth it to keep it like it is until then, instead to create more xml layouts. Anyone with opinions on this?

I think we should atleast use DialogFragment for Dialogs those have some state. Otherwise it is bad experience for the user if the dialog loose state for eg. Dialogs with Text fields like this one. It should retain whatever user has typed in. 

let me know your thoughts :-)

sowjanyakch:
@parneet-guraya I agree that it is a bad user experience. 

If there are many AlertDialogs, I prefer to use Jetpack Compose. In Jetpack Compose, we can reuse the composable with custom styles and content. Managing state also becomes easier. 

The approach you are mentioning creates many xml layouts. These files need to be converted to Jetpack Compose eventually.

mahibi:
Lets merge this one as it definitely improves the dialog.
I did not have a look how many other dialogues may be affected. From my side it's not high priority and i would prefer to wait to check this after we migrated to Jetpack Compose.

PR #3743, opened 2024-03-21
i18n: Change to capital letter

Valdnet:


Valdnet:
@mahibi After removing spaces in a text string:
`<string name="nc_edited_by">"Edited by"</string>`

You need to add a space at the end of this string:
https://github.com/nextcloud/talk-android/blob/e98f60d8846d35b7d5df39ca2ec15ecc60ecd25a/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt#L390

app/src/main/res/values/strings.xml
```diff
@@ -797,8 +797,8 @@ How to translate with transifex:
     <string name="nc_conversation_not_found">Conversation not found</string>
     <string name="add_to_notes">Add to Notes</string>
     <string name="nc_edited_by_admin">Edited by admin</string>
-    <string name="nc_edited_by">"Edited by "</string>
-    <string name="nc_clear_edit_button">clear Edit Button</string>
+    <string name="nc_edited_by">"Edited by"</string>
```
mahibi, line 800: @sowjanyakch 
seeing this now, i suggest to change how this is shown, because for some languages the "Edited by" part might make more sense after the username.

Could you change it to

<string name="nc_edited_by">Edited by %1$s</string>

and then when using this key in code you can use String.format like this:

String.format(
            context.getString(R.string.nc_edited_by),
            message.lastEditActorDisplayName
        )

With this solution the whitespace is also better handled as translators won't think "oh a whitespace at the end of a string, let's delete it"

mahibi:
> @mahibi After removing spaces in a text string: `<string name="nc_edited_by">"Edited by"</string>`
> 
> You need to add a space at the end of this string:
> 
> https://github.com/nextcloud/talk-android/blob/e98f60d8846d35b7d5df39ca2ec15ecc60ecd25a/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt#L390

ah seeing this only now. I suggested to change it [like this](https://github.com/nextcloud/talk-android/pull/3743#discussion_r1534132713)

app/src/main/res/values/strings.xml
sowjanyakch: Agreed. I will make the changes.

sowjanyakch:
I could not perform string formatting on the string "nc_edited_by" (modified its value to "Edited by %1$s") for some reason. I added a new string "message_last_edited_by" and did the formatting. It worked.

mahibi (approved):

mahibi:
> I could not perform string formatting on the string "nc_edited_by" (modified its value to "Edited by %1$s") for some reason. I added a new string "message_last_edited_by" and did the formatting. It worked.

I guess that could be because of already translated strings that don't contain %1$s. Did you use some other language for testing?
Because of this it's a good idea to use a new key :+1:

sowjanyakch:
Edit View

It is the content description of the view "X".

sowjanyakch:
Clicking on the button "X", clears the text in the message field (edited text) and closes the edit view. It is like canceling the edit. 
Clicking on the button :heavy_check_mark: saves the edited text and closes the edit view.

Issue #3547, opened 2023-12-25
Note to self icon does not display


louim-lbs:
### Steps to reproduce

Open Talk

### Expected behaviour

Note to self icon

### Actual behaviour

Black hole

![SmartSelect_20231225_130322_Talk](https://github.com/nextcloud/talk-android/assets/82395416/9b470c93-57ff-489b-8fff-0751ebdb8801)

### Device brand and model

Samsung S20 FE

### Android version

13

### Nextcloud Talk app version

18.0.1

### Nextcloud server version

28.0.1

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

louim-lbs:
Still happening in 18.1.0

mahibi:
Is it shown on web?
I don't have any ideas off the top of my head, except possibly to clear the cache for the app (requires new login)

louim-lbs:
Hello,
No problem on web. Clearing the cache does nothing...
BTW, here is a screenshot.
![img](https://github.com/nextcloud/talk-android/assets/82395416/9d8257bc-39d6-4d63-abe4-28f69e417ec1)

mahibi:
Found out this fails as there is a hardcoded check if the string is "note to self" which fails for translations.
Will change this check!
Workaround for now would be to use english

PR #3812, opened 2024-04-08
share conversation link to other apps

### LINKED ISSUE #2702: Can not share a conversation to another one or copy the link
Just now I tried to guide someone from the community chat to the designer chat.
The problem is there is no way to post a link to the other conversation. I tried switching over but was also unable to find a "copy link" there.
I ended up logging in on firefox navigating with the painful mobile web view to the chat to copy the link.

I think we could help on multiple levels there:

* Have a copy link in the conversation settings (basic)
* Have it in the long press menu of the conversation list (level 2)
* Have a conversation picking option in the share menu (very advanced)

sowjanyakch:
Resolves #2702 

- [x] Have a copy link of a conversation in the conversation settings. 
- [x] Have a conversation link in the long press menu of the conversation list.
- [x] Don't share the conversation link of Note to self.
 

### 🖼️ Screenshots

share_link_option

share_link_to_other_apps

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
Lets make "copy link of a conversation in the conversation settings" available to all users, not only moderators. 
Let's replace the title "Message expiration" with "Conversation settings". And in Conversation settings then add sharing and message expiration as items. On iOS it looks like this (also with more options that may not yet be implemented for android):

grafik

I suggest to also extract the code for sharing the link out of GuestAccessHelper and move it to ShareUtils, so it can be used for ConversationList and ConversationSettings.

mahibi (changes requested):
see overall comment

mahibi:
The settings category is currently hidden whenever message expiration is not available (see method `initExpiringMessageOption`).

As the share button is now also in settings, the settings category should not be hidden, so let's only hide the expiring options now if necessary but keep the settings category visible.

sowjanyakch:
I did not notice this side effect. I was focusing only on sharing the link. 

I made the necessary changes now and also worked on the lint warnings.

app/src/main/res/values/strings.xml
```diff
@@ -785,4 +784,7 @@ How to translate with transifex:
     <string name="nc_edit_icon">Edit Icon</string>
     <string name="get_invitations_error">Failed to fetch pending invitations</string>
     <string name="message_last_edited_by">Edited by %1$s</string>
+    <string name="share_link_to_conversation">Join conversation %1$s at %2$s</string>
+    <string name="note_to_self">Note to self</string>
```
mahibi, line 788: This won't work because translations for one language might differ (web vs app).

I saw this was also done wrong when "note to self" feature was implemented, see https://github.com/nextcloud/talk-android/pull/3351/files#diff-ae89ba240692c2742cc545fd2f57ae82d78e9fce03e3bedc5cfc1af8acbcc1a0R93

(This also causes https://github.com/nextcloud/talk-android/issues/3547)

I will fix this to not use translations at all for these checks and then adapt your PR.

mahibi (approved):

mahibi:
Not sure how "partially verified" happens for commits from @sowjanyakch 
For the backport, these commits become unverified...

We will have to check this again..

PR #3830, opened 2024-04-12
Bugfix/3315/fix empty call notification screen

### LINKED ISSUE #3315: call notification without caller/chat name, not possible to decline
from @ShGKme : I had a very strange call notification from Talk Android right now. There was no caller/chat name and I couldn't decline it.

grafik

mahibi:
fix #3315

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi:
> Thanks!

thanks for reporting!

Issue #1552, opened 2021-08-26
get rid of forked ChatKit


mahibi:
in the past the fork https://github.com/nextcloud/ChatKit was created from https://github.com/stfalcon-studio/ChatKit
The modifications and the useage of https://github.com/nextcloud/ChatKit are too complicated.

The best solution would be to implement the needed designs etc on our own so we are more flexible.
Or at least we should use some well maintained lib (possibly switch back to original https://github.com/stfalcon-studio/ChatKit)

jld3103:
The upstream ChatKit hasn't been updated in 1.5 years, we should probably not use it and just switch away from ChatKit completely.

tobiasKaminsky:
Great idea, but which…?
I searched a bit, but did not find a good alternative…

jld3103:
I think the plan was to build it all on our own, but that is a lot of work...

andyxheli:
Would anything from this link help ?

https://android.libhunt.com/chatkit-alternatives

jld3103:
All the suggestions are not related to chat ui libraries...

AndyScherzinger:
None of the libs in the link provide anything related to a chat view 😔

AndyScherzinger:
@jld3103 yeah, I am unsure if completely building it on our own is worth it or if looking into how to upgrade to their latest version/master would be possible since the lib actually offers some thing we want in their latest version like image or map chat bubbles.

jld3103:
Forking the latest version and then building on top of that sounds good. Some initial work to update it, but a lot less work implementing the basic stuff.

AndyScherzinger:
I quickly talked about it today mentioning this also to @mahibi
Issue is that it is also meshed together with the fresco fork I think. I tried it a while ago but failed due to time constraints on my side.

jld3103:
I looked at the forked ChatKit and Fresco isn't used as a dependency in it. Not sure why everyone thinks this is the case, maybe it's just used together.

mahibi:
yes it's just used together. didn't have a closer look now what exactly is related.
but regarding fresco: there is the need anyway to get rid of the forked fresco because of https://github.com/nextcloud/talk-android/issues/2376 (@timkrueger wants to work on this next i think..)
so maybe we should first try to replace fresco and then chatkit.

timkrueger:
[...]

PR #3839, opened 2024-04-16
Fix onLoadMore not refreshing instantly

rapterjet2004:
I couldn't reproduce the bug that caused the condition to be implemented, but I do faintly recall that `totalItemCount` and `page` were the same when the bug occurred, so I kept this new conditional in place to play it safe.

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -3968,7 +3968,7 @@ class ChatActivity :
     }
 
     override fun onLoadMore(page: Int, totalItemsCount: Int) {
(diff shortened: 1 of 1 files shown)


mahibi:
grafik

this seems to introduce duplicated messages again. Can not reproduce this constantly however

rapterjet2004:
Ok yeah that's weird, I tested it out with the original implementation and it works fine, even though it was bugging out earlier. 

I'm a bit lost, not sure what's causing this inconsistent behavior. Might have to take a look a ChatKit's source code. I really don't like this library. Closing for now.

mahibi:
Tested and works.
This is a prime example of the problems we have with ChatKit, we need to get rid of it!! https://github.com/nextcloud/talk-android/issues/1552

PR #3855, opened 2024-04-18
fix to keep text of message-editing after 30sec

mahibi:
Without this fix, the text when editing a message was lost after pulling chat messages (30sec), because inputEditText was initialized again with the initial text

As a fix, message editing is only initialized once when the edit button was clicked.

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch (approved):
Tested and it works as expected :+1:

Issue #3784, opened 2024-03-29
Refactor MessageInput logic into it's own fragment


rapterjet2004:
- Related https://github.com/nextcloud/talk-android/pull/3704#issuecomment-1999927317

`MessageInput` has grown in complexity. It's unreasonable to have it in `ChatActivity`. It should be moved into it's own fragment to simplify the logic, and prevent future bugs from occurring.

PR #3918, opened 2024-05-17
fix to not set clipboard text repeatedly

### LINKED ISSUE #3897: Unsended text inside the "write text windows" is overwriten by clipboard text, when a 30seconds of time pass
### Steps to reproduce

1.- Share some text to App Nextcloud Talk via the "share to" feature
2.- Enter on a room or personal conversation.
3.- Write some text on the text windows.... AND DON'T SEND IT
4.- Wait some seconds (or a minute).... and...

### Expected behaviour

The text on the text windows doesn't dissapear, being overwriting by the clipboard text.

### Actual behaviour

The text on the text window doesn't change (not get ovewritten by the clipboard text).

### Device brand and model

Xiaomi mi Note 10 (and Samsung s24)

### Android version

14

19.0.0

### Nextcloud server version

28.0.5

### Talk version

19.0.0

No

Yes

Yes

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

mahibi:
fix #3897

initMessageInputView() is called repeatedly when capabilities are received. This is not a good solution but was done because capabilities are needed inside initMessageInputView().

As sharedText is set inside initMessageInputView(), this caused the bug that text was repeatedly inserted in to the message input field.

All in all, there is only the need to initialize initMessageInputView() once. With this fix, it is for now done when there is no filter set yet. As a long term solution, the architecture must be changed so that initMessageInputView() is not called repeatedly, but all UI elements that need capabilities etc are updated by livedata. 

The fix should be seen as a quickfix for 19.0.1, but should be obsolete when architecture continues to improve..

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -1156,92 +1156,94 @@ class ChatActivity :
     }
 
     private fun initMessageInputView() {
-        val filters = arrayOfNulls<InputFilter>(1)
-        val lengthFilter = CapabilitiesUtil.getMessageMaxLength(spreedCapabilities)
+        if (binding.messageInputView.inputEditText?.filters?.isEmpty() == true) {
+            val filters = arrayOfNulls<InputFilter>(1)
+            val lengthFilter = CapabilitiesUtil.getMessageMaxLength(spreedCapabilities)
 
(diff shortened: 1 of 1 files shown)


rapterjet2004 (approved):
> the architecture must be changed so that initMessageInputView() is not called repeatedly, but all UI elements that need capabilities etc are updated by livedata.

I'll see if I can implement this in #3784

PR #4129, opened 2024-08-30
fixes share link

### LINKED ISSUE #4010: Link from Android phone is not transferred to talk
### Steps to reproduce

I would like to tranfer a link from a browser in my Samsung Galaxy Phone to a chat in nextcloud talk, It doesn't work.

### Expected behaviour

Nextcloud Talk 19.1.0 RC1

Hello,

I'm slowly starting to despair of the whole cloud and everything that surrounds it. Constant errors and they cost a lot of time... but I still stick with this system.

Maybe someone here can help me:

For a few days now, I can no longer share links from another app in Talk on my Samsung phone (Android 13). I can select the app talk to share and then it appears, but it just doesn't happen.

Also, I can't forward a notification from one chat to another.

I this a bug?

### Actual behaviour

***

### Device brand and model

Samsung Galaxy 

### Android version

13

19.0.1 RC1

### Nextcloud server version

_No response_

### Talk version

_No response_

None

None

None

### Android logs

_No response_

### Server log

_No response_

### Additional information

_No response_

RamReso:
# Resolves #4010 

## The problem was:
- when sharing a link with nextcloud talk the link disappears.

## General Observations:
- Depending on the Type that is shared with nextcloud talk the behavior of the app differs:
  - e.g. when sharing a text it is copied into the `sharedText` in `ConversationsListActivity` which then opens the `chatActivity`  which then opens `MessageInputFragment`, where then the `sharedText` is set into the `binding.fragmentMessageInputView.inputEditText`, but then not send.
  - when sharing a photo `showSendFilesConfirmDialog` gets called in `ConversationsListActivity` which then uploads the photo
   
## How it is fixed
- just send the message with the `sharedText` in `MessageInputFragment`

## Problems with  the fix
- there is no confirmation dialog, if one really wants to share a link with a certain chat (which could be bad for a person like me, that likes to accidentally press the wrong chat) 
- it is not possible to share a link without notification

### 🚧 TODO

- [ ] get some feedback

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/MessageInputFragment.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/MessageInputFragment.kt
@@ -347,6 +347,7 @@ class MessageInputFragment : Fragment() {
 
         if (chatActivity.sharedText.isNotEmpty()) {
             binding.fragmentMessageInputView.inputEditText?.setText(chatActivity.sharedText)
+            submitMessage(false)
         }
 
         binding.fragmentMessageInputView.setAttachmentsListener {


mahibi:
Thank you @RamReso 

The expected behavior is that the shared text is just pasted into the textfield, so it's possible to edit it and there is no need for a confirm dialog.

The main problem here was that the shared text was set into the textfield, but it was immediately overwritten with "" in the `restoreState` method.

I created the PR https://github.com/nextcloud/talk-android/pull/4139 which should fix it.

I will close your PR but thanks again to have a look!

PR #4194, opened 2024-09-11
sleep one second before websocket reconnect

mahibi:
otherwise it's an endless loop without delay which may stress the devices...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt
+++ b/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt
@@ -41,6 +41,7 @@ import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import java.io.IOException
+import java.lang.Thread.sleep
 import javax.inject.Inject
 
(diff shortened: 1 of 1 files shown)


SystemKeeper:
> otherwise it's an endless loop

Which would raise the question if there should be an error at some point? But yes, we are doing something similar in ios and increase the timeout up to a certain threshold:
https://github.com/nextcloud/talk-ios/blob/83bed85d475e2c38fb63db1caa3eb4e040fcb2ce/NextcloudTalk/NCExternalSignalingController.m#L225-L239

SystemKeeper (approved):

mahibi:
> > otherwise it's an endless loop
> 
> Which would raise the question if there should be an error at some point? But yes, we are doing something similar in ios and increase the timeout up to a certain threshold: https://github.com/nextcloud/talk-ios/blob/83bed85d475e2c38fb63db1caa3eb4e040fcb2ce/NextcloudTalk/NCExternalSignalingController.m#L225-L239

If internet connection itself is lost then the user is informed. You suggest to still inform about if the websocket reconnect fails (maybe also for other reasons than lost internet?).

Increase the timeout will also be done :+1:

Issue #4173, opened 2024-09-09
Should we have passcode timeout `Immediate`?


parneet-guraya:
As a user I'd like to have the immediate option for passcode timeout. can we have it? @mahibi

mahibi:
fine for me

Issue #3815, opened 2024-04-09
Screen lock does not behave predictably


ASerbinski:
### Steps to reproduce

Enable Settings --> Screen lock, and set an inactivity timeout of 30 seconds.

When going in and out of the application, it does not consistently prompt for authentication, even if the timeout period has definitely elapsed.

The issue appears to be tied to SecurityUtils.checkIfWeAreAuthenticated(timeout)

The comment in the function says 
```
            // Try encrypting something, it will only work if the user authenticated within
            // the last AUTHENTICATION_DURATION_SECONDS seconds.
```

I do not believe that this works as expected.

### Expected behaviour

The user should be consistently prompted for authentication if the timeout has expired.

### Actual behaviour

The user may or may not be prompted for authentication even though the timeout has expired.

### Device brand and model

Any

### Android version

14

### Nextcloud Talk app version

master

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

mahibi:
I believe the implementation of this feature can be improved, but for me this is working as expected.
The unlock can also be done on system level, so whenever you unlock your phone, this also counts for the talk app. 
This might explain why you expect the lock screen to be shown when opening the talk app, but it's not shown because you recently unlocked your phone?

mahibi:
I agree that in the scenario with "handing the phone to kids" it would make sense to have to app locked while the phone is unlocked.

On the other hand people may expect there is no app lockscreen when the phone was just unlocked. 
If there is always a lockscreen which they don't expect they might get annoyed of the feature and disable it completely.

So #3816 would make sense so people can decide which behavior they want.

ASerbinski:
> On the other hand people may expect there is no app lockscreen when the phone was just unlocked.
> If there is always a lockscreen which they don't expect they might get annoyed of the feature and disable it completely.

I'm not understanding why anyone would expect that behavior, its different than every other lockable application that I've ever come across.

In fact, my experience has been the complete opposite, that in fact the inconsistency of being prompted to unlock causes people to believe that the feature is unreliable and broken (I have several users who have expressed this issue to me). Why use something at all if you can't trust that it will work when you need it to?

Now here is the problem as I see it; it you're passing off application unlocking to the lockscreen, wouldn't it be just as effective to just use the lockscreen? Just tap the power button and its locked. The purpose of locking the application is to guarantee that the application won't open even though the phone is unlocked.

mahibi:
[...]

PR #4204, opened 2024-09-13
Added option `Immediate` for passcode timeout.

parneet-guraya:
Fix:  #4173

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/values-v28/arrays.xml
+++ b/app/src/main/res/values-v28/arrays.xml
@@ -26,6 +26,7 @@
     </array>
 
     <array name="screen_lock_timeout_descriptions">
+        <item>@string/nc_screen_lock_timeout_immediate</item>
         <item>@string/nc_screen_lock_timeout_30</item>
         <item>@string/nc_screen_lock_timeout_60</item>
         <item>@string/nc_screen_lock_timeout_300</item>
@@ -33,6 +34,7 @@
     </array>
 
     <array name="screen_lock_timeout_entry_values">
+        <item>@string/nc_screen_lock_timeout_immediate</item>
         <item>@string/nc_screen_lock_timeout_thirty</item>
         <item>@string/nc_screen_lock_timeout_sixty</item>
         <item>@string/nc_screen_lock_timeout_three_hundred</item>
@@ -40,6 +42,7 @@
     </array>
 
     <array name="screen_lock_timeout_entry_int_values">
+        <item>1</item>
         <item>30</item>
         <item>60</item>
         <item>300</item>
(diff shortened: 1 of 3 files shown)


parneet-guraya:
> @parneet-guraya - I tested this and lock screen is always active on my Samsung A52 (because inactivity time here is set 0, which practically means no inactive time?). It works when inactivity time is set to 1 sec.

@sowjanyakch 
Hmm... If by always active you mean that if we leave and comeback to the app immediately , lock screen shows? then it is what I aimed for (basically no time window (timeout) i.e 0, lock will be active immediately). Let me know if you meant something else :)

parneet-guraya:
> the screen should lock as soon as you stop interacting with the app.

@sowjanyakch , okay this got me, 'by not interacting' I interpreted that the app's still in foreground but user's not using it. But, this is not what you meant right? Just to be on the same page, this is how the functionality should work ->

- App goes to background when either user press the recent screen button/swipe and hold gesture or completely goes back to home. On return the lock should show. 

-------------

It's a bit strange behavior that you're seeing. I have tested on these devices  and works fine-->
- Oneplus (Android 14)
- Realme (Android 12)
- Google pixel 6a (Android 14)

Although there's another bug that executes the dialog multiple times , this is how you can reproduce ->
- When prompted for password try loosing the focus by either clicking on the home button or the one that shows recent apps. Now this happens regardless of what timeout preference we have.

Is that what you're seeing? 


https://github.com/user-attachments/assets/ae1dd46f-02fb-4172-9944-27798f47efd0

parneet-guraya:
> Open app and set screen lock inactivity timeout to _"Immediate"_ and press recent screen button. Immediately open the app from the recents screen - I see this bug when I try to unlock screen. It happens only when timeout option is set to Immediate.

Strange..., okay when you set immediate option it doesn't show the fingerprint dialog just after setting it? Because for me this bug only occurs when dialog is showing and then I try to loose focus by going back. Also, if it's possible could you post the video of the behavior too?

mahibi:
What's the state for this PR?
@sowjanyakch @parneet-guraya IIRC setting the value to
`<item>1</item>`
and just call it 
`<item>@string/nc_screen_lock_timeout_immediate</item>`
solved the problem that lock screen blocks the app and everything was just working as expected?
Is thee any reason against this?

parneet-guraya:
@mahibi , It worked fine on the devices I was testing on except @sowjanyakch 's samsung device. So, my idea was to fix the Lock screen first. Since, you proposed the idea (#3815) to integrate this functionality in the library. Until that finishes, it is blocked.

mahibi (approved):

mahibi:
we merge with setting immediate option to 1 sec and test this in the next RC
Be aware the behavor will change in the future, see comment in https://github.com/nextcloud/talk-android/pull/4204#issuecomment-2401555587

Issue #3843, opened 2024-04-17
call buttons are shown when they shouldn't ("call buttons were null when trying to disable them")


mahibi:
### Steps to reproduce

open a conversation where the call buttons should be disabled. For example "Note to self".

### Expected behaviour

call buttons are disabled

### Actual behaviour

call buttons are enabled

Log: "call buttons were null when trying to disable them"
`conversationVoiceCallMenuItem` and `conversationVideoMenuItem` are null. 
Handling related to `onPrepareOptionsMenu` must be improved?

I can imagine that sometimes it works and sometimes not, as there could be a race condition.

### Nextcloud Talk app version

19.1.0 Alpha 04 (most likely since ever)

mahibi:
should be done when capabilities are loaded from DB (offline support), so closing for now

mahibi:
still happening with 20.0.1 
Also: for federated rooms, the call buttons disappear after 30 sec because "call buttons were null..."

mahibi:
seems to be explained by the 
`if (this::spreedCapabilities.isInitialized) {.... `
behavior which may be sometimes false when it's supposed to be true.

PR #4239, opened 2024-09-18
fix to not remove call buttons for federated conversations after 30 seconds

mahibi:
After 30 seconds (when the capabilities were updated) the call buttons of federated conversations were sometimes removed (the logic was added back then when fed calls were not implemented).

However this happened not always because of the check `if (this::spreedCapabilities.isInitialized) {....` 
It seems this check sometimes is false when it's supposed to be true. This has be to further investigated and has to be be simplified/improved by a cleaner architecture.

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
while testing, i reopened https://github.com/nextcloud/talk-android/issues/3843 which occured again and seems to be explained by the `if (this::spreedCapabilities.isInitialized) {....` behaviour.

PR #4281, opened 2024-09-26
fix file name of voice recording

### LINKED ISSUE #1564: Check file name of audio recordings
- [ ] Filter out special chars which are not allowed on the filesystem, e.g. `/`, `\` and `:` and replace them with a space
- [ ] Replace multiple spaces with 1
- [ ] Make sure the full filename is not exceeding 250 chars, so trim the name to 246 chars before ammending the filesuffix

Server ref https://github.com/nextcloud/spreed/issues/6174

sowjanyakch:
Fix #1564

- [x] Filter out special chars which are not allowed on the filesystem, e.g. `/`, `\` and `:` and replace them with a space
- [x] Replace multiple spaces with 1 space
- [x] Make sure the full filename is not exceeding 150 chars, so trim the name to 146 chars before ammending the filesuffix

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/utils/FileUtils.kt
+++ b/app/src/main/java/com/nextcloud/talk/utils/FileUtils.kt
@@ -27,7 +27,6 @@ object FileUtils {
     private val TAG = FileUtils::class.java.simpleName
     private const val RADIX: Int = 16
     private const val MD5_LENGTH: Int = 32
-    const val FILE_MAX_LENGTH = 146
 
     /**
      * Creates a new [File]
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/utils/FileUtils.kt
```diff
@@ -27,6 +27,7 @@ object FileUtils {
     private val TAG = FileUtils::class.java.simpleName
     private const val RADIX: Int = 16
     private const val MD5_LENGTH: Int = 32
+    const val FILE_MAX_LENGTH = 146
```
mahibi, line 30: the max length is set to 146 instead 150 because of the suffix ".mp3". This is not obvious in the FileUtils class and may lead to confusion. So i suggest to define the max length inside MediaRecorderManager and call it "VOICE_MESSAGE_PREFIX_MAX_LENGTH"

mahibi (approved):

PR #3481, opened 2023-11-30
fix to enable links in markdown

mahibi:
fix https://github.com/nextcloud/talk-android/issues/3633

fix to enable links in markdown when no top level domain was included in the link description.

Examples (without TLD):

[Hacker News](https://news.ycombinator.com/item?id=38457953)
[AlternativeTo](https://alternativeto.net/news/2023/11/nextcloud-welcomes-open-source-webmail-software-roundcube-to-its-platform/)
[Phoronix](https://www.phoronix.com/news/Roundcube-Nextcloud)
[BNN-Breaking](https://bnn.network/bnn-newsroom/roundcube-and-nextcloud-join-forces-in-unprecedented-open-source-partnership/)
[OSBA – Open Source Business Alliance](https://osb-alliance.de/news/nextcloud-bietet-neues-zuhause-fuer-den-e-mail-giganten-roundcube)

Examples (with TLD):

[LinuxNews.de](https://linuxnews.de/roundcube-fusioniert-mit-nextcloud/)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Bildschirmfoto vom 2023-11-30 11-02-24 geaendert | Bildschirmfoto vom 2023-11-30 11-02-33 geaendert

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not needed
- [x] 🔖 Capability is checked or not needed 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set
- [x] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #4297, opened 2024-10-01
handle link clicks properly

### LINKED ISSUE #1789: Link activity
## Steps to reproduce

1. Receive or send a link to e.g. a video.
2. Tap and hold on the **link address**.

### Expected behaviour
Only the context menu appears.

### Actual behaviour
The appearance of the context menu and the window with the selection of the application to run the link.

### Screenshot
obraz

## Device information

**Android version:** 10

**Talk version:** 13.0.0 and 13.1.0 Alpha 09

sowjanyakch:
fix #1789

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingLinkPreviewMessageViewHolder.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingLinkPreviewMessageViewHolder.kt
@@ -9,7 +9,12 @@ package com.nextcloud.talk.adapters.messages
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.text.SpannableString
+import android.text.Spanned
 import android.text.TextUtils
+import android.text.style.UnderlineSpan
 import android.util.Log
 import android.view.View
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingLinkPreviewMessageViewHolder.kt
```diff
@@ -92,6 +97,16 @@ class IncomingLinkPreviewMessageViewHolder(incomingView: View, payload: Any) :
 
         binding.messageText.text = processedMessageText
 
+        val link = SpannableString(processedMessageText)
```
mahibi, line 100: `processedMessageText` can be mixed text with hyperlinks. So "test http://anysite.com" can't be directly converted to a link but the link has to be extracted and only this should be underlined and clickable.
Otherwise there could be a ActivityNotFoundException if the link is not valid.

mahibi: to extract links, code from https://github.com/nextcloud/talk-android/blob/7ca68b334a625e82a3865b5d1a8d7d50343a5cfe/app/src/main/java/com/nextcloud/talk/chat/data/model/ChatMessage.kt#L167
will be helpful. This is the code which decides if the *LinkPreviewMessageViewHolder is applied.

app/src/main/res/layout/activity_chat.xml
```diff
@@ -154,7 +154,6 @@
             app:outcomingTextLinkColor="@color/high_emphasis_text"
             app:outcomingTextSize="@dimen/chat_text_size"
             app:outcomingTimeTextSize="12sp"
-            app:textAutoLink="all"
```
mahibi, line 157: removing 
`app:textAutoLink="all"`
works and may be mandatory to fix the bug.

However this will change the behavior of all ChatMessage types and automatic link detection won't work as described in
https://github.com/nextcloud/talk-android/pull/3481#issuecomment-2099087795

app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingLinkPreviewMessageViewHolder.kt
mahibi: However thinking about that there can be multiple links inside a message this gets more complicated. Then there might have to be more complex logic similar like it's used for the chip mentions: https://github.com/nextcloud/talk-android/blob/9d62a6f7456d5980433a1e4d58a35e2f4dbbccdb/app/src/main/java/com/nextcloud/talk/utils/DisplayUtils.kt#L220

PR #4318, opened 2024-10-07
group chat messages

### LINKED ISSUE #1747: Messages grouped wrong
Messages are grouped wrong in 1-to-1 and group conversations:

| Android | Spreed |
| :-: | :-: |
| image | image |

sowjanyakch:
fix #1747

Align with web on grouping messages.

- [x] Group messages from same author sent on the same day and time difference should be less than or equal to 5 minutes.
- [x] Never group bot messages (apart from changelog bot)
- [x] Also group system messages from different users
- [x] Never group edited messages.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2667,7 +2667,7 @@ class ChatActivity :
 
             adapter?.let {
                 val previousChatMessage = adapter!!.items[1].item
-                if(previousChatMessage is ChatMessage){
+                if (previousChatMessage is ChatMessage) {
                     chatMessage.isGrouped = groupMessages(chatMessage, previousChatMessage)
                 }
                 chatMessage.isOneToOneConversation =
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2666,9 +2666,10 @@ class ChatActivity :
             chatMessage.activeUser = conversationUser
 
             adapter?.let {
-                val previousChatMessage = adapter!!.items[1].item as ChatMessage
-                chatMessage.isGrouped = groupMessages(chatMessage, previousChatMessage)
-
+                val previousChatMessage = adapter!!.items[1].item
+                if(previousChatMessage is ChatMessage){
+                    chatMessage.isGrouped = groupMessages(chatMessage, previousChatMessage)
+                }
                 chatMessage.isOneToOneConversation =
                     (currentConversation?.type == ConversationEnums.ConversationType.ROOM_TYPE_ONE_TO_ONE_CALL)
                 chatMessage.isFormerOneToOneConversation =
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -2600,11 +2600,10 @@ class ChatActivity :
             chatMessage.activeUser = conversationUser
 
             adapter?.let {
-                chatMessage.isGrouped = (
-                    it.isPreviousSameAuthor(chatMessage.actorId, -1) &&
-                        it.getSameAuthorLastMessagesCount(chatMessage.actorId) %
-                        GROUPED_MESSAGES_SAME_AUTHOR_THRESHOLD > 0
-                    )
+                val lastItemIndex = adapter!!.itemCount - 1
+                val previousChatMessage = adapter!!.items[lastItemIndex].item as ChatMessage
```
mahibi, line 2604: `previousChatMessage` is not calculated correctly.

When querying
`adapter!!.itemCount - 1`

you would get some old messages (the oldest one available in the adapter).

The newest message has index 0

So it's just
`val previousChatMessage = adapter!!.items[0].item as ChatMessage`

But also it must be checked if the message is a ChatMessage. Otherwise there will be cast exceptions, for example when it's a date-message (the ones that are shown between the messages, saying "Today", "Yesterday", ...). Just look into the implementation of `isPreviousSameAuthor` to see the checks that may be necessary.

mahibi (changes requested):

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -2600,11 +2600,9 @@ class ChatActivity :
             chatMessage.activeUser = conversationUser
 
             adapter?.let {
-                chatMessage.isGrouped = (
-                    it.isPreviousSameAuthor(chatMessage.actorId, -1) &&
-                        it.getSameAuthorLastMessagesCount(chatMessage.actorId) %
-                        GROUPED_MESSAGES_SAME_AUTHOR_THRESHOLD > 0
-                    )
+                val previousChatMessage = adapter!!.items[1].item as ChatMessage
```
mahibi, line 2603: As said, there must be checks if the previousChatMessage is a ChatMessage.
To reproduce this crashes when the check is not in place, write a message from webuser to androiduser in a chat that does not yet have any message for this day.

mahibi (changes requested):

mahibi:
when testing i got
grafik

while for server it is
grafik

This happened when i sent messages while being in the chat, so must be caused in `processMessagesFromTheFuture`

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -2688,6 +2677,40 @@ class ChatActivity :
...
+                )
+        ) {
+            return false
+        }
+
+        val timeDifference = dateUtils.getTimeDifferenceInSeconds(
+            message2.timestamp,
+            message1.timestamp
+        )
+        val isLessThan5Min = timeDifference > 300
```
mahibi, line 2703: 300 should be a const val FIVE_MINUTES_IN_SECONDS

sowjanyakch:
> when testing i got 
grafik
> 

@mahibi - Does this happens for you all the times or sometimes?

mahibi (approved):

mahibi:
> > when testing i got
> > grafik
> 
> @mahibi - Does this happens to you all the times or only sometimes?

tested again and cant reproduce. For me it's fine to merge then

PR #4327, opened 2024-10-09
show deck card in messages

### LINKED ISSUE #1025: deck card link is show as {object}

sowjanyakch:
fix #1025

- [x] show deck card in the chat messages when user shared card via "Post to conversation".
- [x] Clicking on the deck card chat message will navigate to the deck card on the web.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
Playing around how deck cards in more detail for the first time. It feels quite confusing...:

The handling how deck cards are viewed are quite depending on who sent it in which way, if the board is shared to the other user, and also it's different by platforms.

See numbers in the first screenshot of web...

1&2 were received when the board was not yet shared.

1) was received from other user who sent it from deck via "Post to a conversation"
2) was received from other user by smartpicker (however if the board is shared beforehand, it displayed via preview(OpenGraph?))
3) was sent to other user via "Post to a conversation"
4) was sent to other user by smartpicker (it's a link that has a preview by OpenGraph)


This is how it looks in web:
grafik

Android (with this PR):
grafik

iPhone:
grafik
grafik

## Todos:

  - [ ] todo for server people?: share board when a card is posted to a chat???
  - [ ] we should decide if cards are shown by own design or by OpenGraph? But mixed it's confusing?
  - [ ] todo for android?: visualize OpenGraph link when it's a Deck Card? Is this how it's done on iOS and web? 
  - [ ] todo for android: check if links are opened in Android Nextcloud Deck app / if not whats needs to be done?

pinging some people if anyone want to share opinions..
@nickvergessen @Ivansss @SystemKeeper @juliusknorr

app/src/main/res/values/strings.xml
```diff
@@ -415,6 +416,7 @@ How to translate with transifex:
     <string name="nc_sent_poll_you">You sent a poll.</string>
     <string name="nc_sent_location_you">You sent a location.</string>
     <string name="nc_sent_voice_you">You sent a voice message.</string>
+    <string name="nc_sent_deck_card_you">You: %1$s</string>
```
mahibi, line 419: You sent a deck card.

app/src/main/res/values/strings.xml
```diff
@@ -407,6 +407,7 @@ How to translate with transifex:
     <string name="nc_sent_poll" formatted="true">%1$s sent a poll.</string>
     <string name="nc_sent_location" formatted="true">%1$s sent a location.</string>
     <string name="nc_sent_voice" formatted="true">%1$s sent a voice message.</string>
+    <string name="nc_sent_deck_card" formatted="true">%1$s: %2$s</string>
```
mahibi, line 410: %1$s sent a deck card.

SystemKeeper:
> 1 was received from other user who sent it from deck via "Post to a conversation"

We try to use the link that included in the richObject to get additional information from the reference provider (not opengraph). If that is not possible, we render the deck card with the information we have in the richObject. Actually the last one is not correctly working, as our check is wrong, you found a bug :) https://github.com/nextcloud/talk-ios/issues/1843. Without the bug, we should render the same as android.

> 2 was received from other user by smartpicker (however if the board is shared beforehand, it displayed via preview(OpenGraph?))

Since it's just a link, we try to get reference data. If there's one for "deck-card" we use that: https://github.com/nextcloud/talk-ios/blob/25ebf7eca1f1043ab89b7ff4791a4eac34dff075/NextcloudTalk/ReferenceView.swift#L126-L135. Otherwise we fall back to opengraph data, which most of the time is just the link.

> 3 was sent to other user via "Post to a conversation"

Should be the same as 1. Not totally sure, why it's rendered in this case. Maybe the reference provider behaves differently, but need to check.

> 4 was sent to other user by smartpicker (it's a link that has a preview by OpenGraph)

Same as 2.

app/src/main/res/values/strings.xml
sowjanyakch: @mahibi - The string is modified to align with web

app/src/main/res/values/strings.xml
sowjanyakch: The string is modified to align with web

sowjanyakch:
@nickvergessen - Is the current implementation of deck card on android okay for you or do you want to make any changes to the way the deck card is shown?

app/src/main/res/values/strings.xml
nickvergessen: The web is always saying `You: …` but that's not the best. Use what Marcel suggested, to align with the other messages of the same type that exist atm in Android

app/src/main/res/values/strings.xml
nickvergessen: Same as above

nickvergessen:
Regarding the questions/comments in https://github.com/nextcloud/talk-android/pull/4327#issuecomment-2422519579

> was received from other user by smartpicker (however if the board is shared beforehand, it displayed via preview(OpenGraph?))

When the board/card is not publicly accessible, OpenGraph will not be able to return any information, so that's expected.

> todo for server people?: share board when a card is posted to a chat???

It was decided that automatic sharing should not happen, as otherwise people might expect that quickly afterwards deleting the message would also undo the share, but we wouldn't know if we just shared or not. In addition we would need to understand and interact with quite a lot 3rdparty app APIs in those cases, so that's a "No" for now.

> we should decide if cards are shown by own design or by OpenGraph? But mixed it's confusing?

There is also a significant difference when you share a file to chat or when you post `/f/` link or even a public share link. So yes, for now the difference is expected.

> todo for android: check if links are opened in Android Nextcloud Deck app / if not whats needs to be done?

While that would be cool, we can also simply raise a follow up issue and keep this PR here scoped.

mahibi (approved):

PR #4341, opened 2024-10-15
Fixed crashes with AudioFocusRequestManager on Android 7.1

### LINKED ISSUE #4338: Crash when opening any conversation on Android 7.1
### Steps to reproduce

1. Open Talk
2. Click on any conversation (confirmed with Talk updates and a group chat)
3. Crash

### Expected behaviour

The conversation should open.

### Actual behaviour

Crashes because `java.lang.NoClassDefFoundError: Failed resolution of: Landroid/media/AudioFocusRequest$Builder;`

Culprit based on the logs: https://github.com/nextcloud/talk-android/blob/e9fc23a5611406179ed6677f72f8304722358b07/app/src/main/java/com/nextcloud/talk/chat/data/io/AudioFocusRequestManager.kt#L65

### Device brand and model

samsung SM-J320F

### Android version

25

v20.1.0 Alpha 08

### Nextcloud server version

29.0.8

### Talk version

19.0.10

No

No

No

### Android logs

[logcat-relevant.txt](https://github.com/user-attachments/files/17369145/logcat-relevant.txt)

### Server log

_No response_

### Additional information

Same behavior on latest stable (v20.0.2).

rapterjet2004:
- fixes #4338 

https://github.com/nextcloud/talk-android/issues/4338#issuecomment-2414012168

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/data/io/AudioFocusRequestManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/data/io/AudioFocusRequestManager.kt
@@ -70,24 +70,17 @@ class AudioFocusRequestManager(private val context: Context) {
      * Requests the OS for audio focus, before executing the callback on success
      */
     fun audioFocusRequest(shouldRequestFocus: Boolean, onGranted: () -> Unit) {
-        if (isPausedDueToBecomingNoisy) {
+        if (isPausedDueToBecomingNoisy || Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
             onGranted()
             return
         }
 
-        val isGranted: Int = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-            if (shouldRequestFocus) {
-                audioManager.requestAudioFocus(focusRequest)
-            } else {
-                audioManager.abandonAudioFocusRequest(focusRequest)
-            }
+        val isGranted: Int = if (shouldRequestFocus) {
+            audioManager.requestAudioFocus(focusRequest)
         } else {
-            if (shouldRequestFocus) {
-                audioManager.requestAudioFocus(audioFocusChangeListener, AudioManager.STREAM_MUSIC, duration)
-            } else {
-                audioManager.abandonAudioFocus(audioFocusChangeListener)
-            }
+            audioManager.abandonAudioFocusRequest(focusRequest)
         }
+
         if (isGranted == AudioManager.AUDIOFOCUS_REQUEST_GRANTED) {
             onGranted()
             handleBecomingNoisyBroadcast(shouldRequestFocus)
--- a/app/src/main/java/com/nextcloud/talk/chat/data/io/AudioFocusRequestManager.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/data/io/AudioFocusRequestManager.kt
@@ -14,7 +14,6 @@ import android.content.IntentFilter
 import android.media.AudioFocusRequest
 import android.media.AudioManager
 import android.os.Build
-import androidx.annotation.RequiresApi
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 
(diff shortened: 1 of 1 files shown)


MmAaXx500:
Unfortunately, this does not fix the issue, and it seems to me that on versions before android O, the audio focus request got removed.

On my fork, I created a commit that fixes this issue, but didn't have the time to submit a pr. https://github.com/MmAaXx500/talk-android/tree/api25-audiofocus
If you want, I can open a pr, or you can take the code from there.

MmAaXx500:
Now there is no crashing, but waits indefinitely on opening a conversation.

If you have no specific reason to remove the `requestAudioFocus` and `abandonAudioFocus` then you can drop your first commit (1dfdb8387fa2456606c64b908b3570646b740151) and use my suggested changes only.

sowjanyakch:
@MmAaXx500  - thank you for testing and mentioning that there is no crash. I could not test this PR because I don't have a device that runs on android 7.1 or below. I referred android documentation, https://developer.android.com/media/optimize/audio-focus#audio-focus-7-1, I see that it's okay to use requestFocus() and abandonFocus(). 

One more suggestion always use safe call  with let '?.let' whenever possible instead of '!!' to avoid unnecessary crashes. 
Instead of 
audioManager.requestAudioFocus(focusRequest!!)
we can use
focusRequest?.let { request ->
    audioManager.requestAudioFocus(request)
}

mahibi:
Sorry all, I should have looked directly when the issue came up.
I just calculated the percentage usage of android 7.0 and 7.1 for our user base.

For 7.0 it is 0,3%
For 7.1 it is 0,19%

We have to rule to ditch android version lower than 0,5%.

So instead to introduce more logic here, we should create a PR that removes support for these versions and bump the minSdkVersion to 26 (Android 8.0) which currently has an install base of 0,7%.

Sorry if this is disappointing if you're a user with android version lower than 8, but we are not able to support endless scenarios with the cost of complexity.

ksihota:
I have 20 LG phones that are unable to be upgraded past Android 7.0 
I tried installing an older version of the Talk App that supports Android 7 but they all crash when opening a conversation. Is there any way to make an older version of the App work on the latest NextCloud Talk server? 
I have also tried to find a custom ROM that would allow me to upgrade the phones but the only two I could find appear to cause a boot loop on these phones. I can access the Talk conversations through the browser but it it pretty cumbersome on the phone.
Any suggestions would be appreciated.

ksihota:
I have 20 LG phones that are unable to be upgraded past Android 7.0 
I tried installing an older version of the Talk App that supports Android 7 but they all crash when opening a conversation. Is there any way to make an older version of the App work on the latest NextCloud Talk server? 
I have also tried to find a custom ROM that would allow me to upgrade the phones but the only two I could find appear to cause a boot loop on these phones. I can access the Talk conversations through the browser but it it pretty cumbersome on the phone.
Any suggestions would be appreciated.

MmAaXx500:
Hi @ksihota, I have a fork of this repo that I patched to work on Android 7, however I only did it for 20.1.0 so far, but seems to be working fine with an updated NC 31 installation.
If you can compile yourself, you can pull the `v20.1.0-api24` branch from here: https://github.com/MmAaXx500/talk-android/tree/v20.1.0-api24

If you need 21.0 I may do an update next week.

ksihota:
Thanks for the link @MmAaXx500. I'll try and learn how to compile it. Right now I'm just trying to see if I can get any Talk app to work on these phones so your work looks the most promising to me. Thanks again.

MmAaXx500:
Glad to hear it's working!

The APKs are saved under `build/outputs/apk`, depending on the variant you built.

Just a small note about build variants: push notifications are only supported in the `gplay` variant. This is [a limitation in Talk itself](https://github.com/nextcloud/talk-android/blob/master/docs/notifications.md#-developerstesters), not something introduced by my patches.
You can switch the build variant if you need via `Build -> Select Build Variant...` in Android Studio.

ksihota:
Thanks sooooo much. I stumbled upon both of those last night and was able to create a working APK. Its good to know that what I did was correct and when I get the chance will have to explore my choices of options more carefully. I saw the gplay option but didn't try it yet.
I am curious to know what changes you had to do to make it work with Android 7. I'll have to look more carefully at your fork to see what you have done.
Thanks again for all your assistance.

PR #4346, opened 2024-10-18
Fix ignored blank lines in Javadocs

danxuliu:
Given that there is no risk in backporting this I suggest doing it to ease other future backports that might touch the Javadocs too.


mahibi (approved):
fine for me

PR #4360, opened 2024-10-22
Last moderator should not leave the room

### LINKED ISSUE #490: Show error when last moderator tries to leave a room
https://github.com/nextcloud/talk-ios/pull/207

sowjanyakch:
resolve #490

### 🖼️ Screenshots
Screenshot 2024-10-28 at 09 21 09

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


app/src/main/java/com/nextcloud/talk/jobs/LeaveConversationWorker.kt
```diff
@@ -0,0 +1,89 @@
...
+                .subscribe(object : Observer<GenericOverall?> {
+                    override fun onSubscribe(d: Disposable) {
+                    }
+
+                    override fun onNext(p0: GenericOverall) {
+                    }
+
+                    override fun onError(e: Throwable) {
+                        Log.e(TAG, "Failed to remove self from room", e)
+                        if (e.message?.contains("HTTP 400") == true) {
```
mahibi, line 70: For errors other than http400, Result.failure() would not be thrown which could cause some bugs (worker won't finish...).

So we have to make sure to differentiate between the errors. We can do this by passing data to the result, 
so define a consts like:
```
  const val ERROR_NO_OTHER_MODERATORS_OR_OWNERS_LEFT = "NO_OTHER_MODERATORS_OR_OWNERS_LEFT"
  const val ERROR_OTHER = "ERROR_OTHER"
```
and set it as data for error_type (which could also be a const string):

Also instead to analyze the string, let's check the error code

```
  override fun onError(e: Throwable) {
      Log.e(TAG, "Failed to remove self from room", e)
      val httpException = e as? HttpException
      val errorData = if (httpException?.code() == 400) {
          Data.Builder()
              .putString("error_type", ERROR_NO_OTHER_MODERATORS_OR_OWNERS_LEFT)
              .build()
      } else {
          Data.Builder()
              .putString("error_type", ERROR_OTHER)
              .build()
      }
      result.set(Result.failure(errorData))
  }
```

When getting the work info, read the data to identify the error type:

```
WorkInfo.State.FAILED -> {
    val errorType = workInfo.outputData.getString("error_type")

    if (errorType == LeaveConversationWorker.ERROR_NO_OTHER_MODERATORS_OR_OWNERS_LEFT) {
        Snackbar.make(
            binding.root,
            R.string.nc_last_moderator_leaving_room_warning,
            Snackbar.LENGTH_LONG
        ).show()
    } else {
        Snackbar.make(
            binding.root,
            R.string.nc_common_error_sorry,
            Snackbar.LENGTH_LONG
        ).show()
    }
}
```

mahibi (changes requested):

mahibi (approved):

PR #4386, opened 2024-10-25
Add support for all parameter in leave call endpoint

### LINKED ISSUE #4271: Implement short/long press options for "End call"/"Leave call"
Clarifying expectation:
- One-to-one:
  - Short/Normal press: "End call" (Label (if shown) is short, but behaviour is "End call for everyone")
  - Long press: "Leave call"
- Group:
  - Short/Normal press: "Leave call"
  - Long press: "End call for everyone"

---

- iOS PR https://github.com/nextcloud/talk-ios/pull/1629

sowjanyakch:
Resolve #4271

One-to-one:

Short/Normal press: "End call" ("End call for everyone")
Long press: "Leave call"

Group:

Short/Normal press: "Leave call"
Long press: "End call for everyone"

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -143,7 +143,6 @@ import io.reactivex.Observable
 import io.reactivex.Observer
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.Disposable
-import io.reactivex.observables.GroupedObservable
 import io.reactivex.schedulers.Schedulers
 import okhttp3.Cache
 import org.apache.commons.lang3.StringEscapeUtils
@@ -679,8 +678,21 @@ class CallActivity : CallBaseActivity() {
             }
         }
 
+        if(isOneToOneConversation && isModerator){
+            binding!!.hangupButton.setOnClickListener {
+                showPopupMenu()
+            }
+        }
+
+        if(isOneToOneConversation){
+            binding!!.hangupButton.setOnLongClickListener {
+                hangup(true, null)
+                true
+            }
+        }
+
         if (isModerator && isGroupConversation) {
-           binding!!. hangupButton.setOnLongClickListener {
+           binding!!.hangupButton.setOnLongClickListener {
                 showPopupMenu()
                 true
             }
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -112,13 +112,12 @@ import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_CONVERSATION_PASSWORD
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_FROM_NOTIFICATION_START_CALL
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_IS_BREAKOUT_ROOM
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_IS_MODERATOR
-import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_IS_ROOM_GROUP
-import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_IS_ROOM_ONE_TO_ONE
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_MODIFIED_BASE_URL
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_AUDIO
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_VIDEO
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_RECORDING_STATE
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_ID
+import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_ONE_TO_ONE
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_TOKEN
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_START_CALL_AFTER_ROOM_SWITCH
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_SWITCH_TO_ROOM
@@ -385,8 +384,7 @@ class CallActivity : CallBaseActivity() {
         canPublishAudioStream = extras.getBoolean(KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_AUDIO)
         canPublishVideoStream = extras.getBoolean(KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_VIDEO)
         isModerator = extras.getBoolean(KEY_IS_MODERATOR, false)
-        isOneToOneConversation = extras.getBoolean(KEY_IS_ROOM_ONE_TO_ONE, false)
-        isGroupConversation = extras.getBoolean(KEY_IS_ROOM_GROUP, false)
+        isOneToOneConversation = extras.getBoolean(KEY_ROOM_ONE_TO_ONE, false)
 
         if (extras.containsKey(KEY_FROM_NOTIFICATION_START_CALL)) {
             isIncomingCallFromNotification = extras.getBoolean(KEY_FROM_NOTIFICATION_START_CALL)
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/nextcloud/talk/utils/bundle/BundleKeys.kt
```diff
@@ -80,4 +80,6 @@ object BundleKeys {
     const val KEY_CREDENTIALS: String = "KEY_CREDENTIALS"
     const val KEY_FIELD_MAP: String = "KEY_FIELD_MAP"
     const val KEY_CHAT_URL: String = "KEY_CHAT_URL"
+    const val KEY_IS_ROOM_ONE_TO_ONE: String = "KEY_IS_ROOM_ONE_TO_ONE"
```
mahibi, line 83: there is already a bundle key with 
KEY_ROOM_ONE_TO_ONE

just check if it's set correctly in ChatActivity or CallNotificationActivity set so you can be sure it's in place when you get it in CallActivity.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -381,6 +385,9 @@ class CallActivity : CallBaseActivity() {
         canPublishAudioStream = extras.getBoolean(KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_AUDIO)
         canPublishVideoStream = extras.getBoolean(KEY_PARTICIPANT_PERMISSION_CAN_PUBLISH_VIDEO)
         isModerator = extras.getBoolean(KEY_IS_MODERATOR, false)
+        isOneToOneConversation = extras.getBoolean(KEY_IS_ROOM_ONE_TO_ONE, false)
```
mahibi, line 388: i think it's enough to just check for the one-to-one key.
So no need to introduce KEY_IS_ROOM_GROUP. If it's not a one-to-one conversation do the handling for the click listeners as it's a group conversation.

app/src/main/java/com/nextcloud/talk/utils/bundle/BundleKeys.kt
mahibi: The KEY_ROOM_ONE_TO_ONE must also be set in CallNotificationActivity, maybe also be passed through it..
Otherwise the clickListeners may not be triggered when the bundle keys are not set whenever a call is received via notification.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -663,7 +675,38 @@ class CallActivity : CallBaseActivity() {
...
+
+        if (isGroupConversation) {
+            binding!!.hangupButton.setOnClickListener {
+                hangup(true, null)
+            }
+        }
+
+        if (isOneToOneConversation && isModerator) {
+            binding!!.hangupButton.setOnClickListener {
+                showPopupMenu()
```
mahibi, line 687: > One-to-one:
> Short/Normal press: "End call" (Label (if shown) is short, but behaviour is "End call for everyone")

so the action here must be
`hangup(true, true)`

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -663,7 +675,38 @@ class CallActivity : CallBaseActivity() {
...
+            }
+        }
+
+        if (isOneToOneConversation) {
+            binding!!.hangupButton.setOnLongClickListener {
+                hangup(true, null)
```
mahibi, line 693: here, show the popup button instead.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -663,7 +675,38 @@ class CallActivity : CallBaseActivity() {
...
+
+        if (isModerator && isGroupConversation) {
+            binding!!.hangupButton.setOnLongClickListener {
+                showPopupMenu()
+                true
+            }
+        }
+
+        binding!!.popupMenu.setOnClickListener {
+            hangup(true, true)
```
mahibi, line 706: Actually, also for the popup button there has to be a check if it is a one-to-one conversation or not, so the behavior of the button is different depending on the type.

mahibi (changes requested):

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -1331,6 +1378,30 @@ class CallActivity : CallBaseActivity() {
...
+
+                val buttonLeft = buttonLocation[0]
+                val buttonTop = buttonLocation[1]
+                val buttonRight = buttonLeft + popupMenuWidth
+                val buttonBottom = buttonTop + popupMenuHeight
+
+                val x = event.rawX
+                val y = event.rawY
+                if (x < buttonLeft || x > buttonRight || y < buttonTop || y > buttonBottom) {
+                    binding!!.popupMenu.visibility = View.GONE
```
mahibi, line 1398: i think analyzing pixels should be avoided just to hide the button. 
Let's try to solve this via focus handling. I guess/hope onFocusChangeListener could do the job.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -1953,7 +2024,7 @@ class CallActivity : CallBaseActivity() {
         }
     }
 
-    private fun hangup(shutDownView: Boolean) {
+    private fun hangup(shutDownView: Boolean, endCallForAll: Boolean?) {
```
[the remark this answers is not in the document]
sowjanyakch: The parameter "all" is set to true only when we need to end call for everyone. Otherwise, it is not set(so passing null should be okay here)

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
sowjanyakch: Short/Normal press of hangupButton in one-one conversation should show a button ("End call for everyone"). If you click this button, then only end call for everyone. This is how ioS does it.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
sowjanyakch: The current implementation automatically hides the button when a click occurs outside of its boundaries, like in ioS. This can be implemented by using Touch Listeners,  I don't think we can use FocusChangeListener to achieve this functionality.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
mahibi: I clarified with @nickvergessen and @Ivansss it should also show the popup button like it is implemented in iOS (it was not in the issue description but it has now been decided that this makes more sense).
So in the end the popup button must be able to handle different actions like i mentioned in it's clickListener.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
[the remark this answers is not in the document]
mahibi: I think when there are only 2 possibilities we should make use of true/false, so don't make it nullable.
If the API needs null in case it's false, then implement it just before the API call.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -699,6 +729,10 @@ class CallActivity : CallBaseActivity() {
         binding!!.lowerHandButton.setOnClickListener { l: View? -> raiseHandViewModel!!.lowerHand() }
     }
 
+    private fun showPopupMenu() {
+        binding!!.popupMenu.visibility = View.VISIBLE
```
mahibi, line 733: popupMenu should be renamed to indicate what it is about. Lets call it `leaveCallPopupMenu` ?
Same for the function name.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
mahibi: Then lets do it the other way round and hide the button when other layouts are touched.

There is already code for

```
        binding!!.gridview.setOnTouchListener { _, me ->
            val action = me.actionMasked
            if (action == MotionEvent.ACTION_DOWN) {
                animateCallControls(true, 0)
            }
            false
        }
        binding!!.conversationRelativeLayout.setOnTouchListener { _, me ->
            val action = me.actionMasked
            if (action == MotionEvent.ACTION_DOWN) {
                animateCallControls(true, 0)
                binding!!.popupMenu.visibility = View.GONE
            }
            false
        }
```

Just add a line to hide the button inside the `action == MotionEvent.ACTION_DOWN` blocks

nickvergessen:
What's the status here?

mahibi (approved):

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
mahibi: renamed it

PR #4389, opened 2024-10-28
User now knows if no offline messages are saved when entering a conversation

### LINKED ISSUE #4377: Info about "No offline messages available"
When being offline and entering a chat that has no downloaded messages, instead of the loading spinner there should be an info that no messages are available yet offline.

Otherwise it feels like messages are loading, which is not true.

rapterjet2004:
- fixes #4377 

Screenshot 2024-11-13 at 7 59 29 AM

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi (changes requested):
This also shows the snackbar when being online and loading messages for the first time. It should only be shown when being offline and no messages are available.

Instead of a snackbar, it would be nice to have the info as a background image in the chat (instead to see the animated progress spinner)

mahibi (approved):

PR #4251, opened 2024-09-19
Showing temporary messages when queued

rapterjet2004:
### 🖼️ Screenshots
Screenshot 2024-09-19 at 9 26 21 AM

### 🚧 TODO
- [x] Got UI working
- [ ] Maintain state on lifecycle changes
- [ ] Allowing editing
- [ ] Allowing deleting
- [ ] handle replies

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #4422, opened 2024-11-06
Feature/4378/add temporary messages while sending

### LINKED ISSUE #4378: Show pending indicator for messages that are being sent (incl retry)
Messages that are being sent should be visible inside the chat, but there should be an indicator that sending is still in progress.
Or if sending failed, there should be an option to re-trigger sending.

Also see https://github.com/nextcloud/talk-android/pull/1259 for an old draft. Now with offline support this may have to be handled differently.

Also make sure to set reference id's for messages that are sent when being offline: https://github.com/nextcloud/talk-android/pull/4251.

## Basic steps:
(discussed with @Ivansss)

When sending message (also while being offline):
- create temporary message
- set a reference id
- put it into chatlist and show an icon (spinner or clock..)

When receiving send-response or get messages
- check all reference ids
   - if reference id from a received message is equal to a reference id of temporary message
     - replace temporary message with received message
     - delete temorary message from DB

## Links:
- https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/BaseChatViewController.swift#L480
- https://nextcloud-talk.readthedocs.io/en/latest/chat/#sending-a-new-chat-message

mahibi:
resolve #4378

### 🖼️ Screenshots

When sending, a clock icon is shown:

When being offline, an icon is shown:

When sending failed, a warning icon is shown:

On long click, actions are offered for temporary messages:
- "Copy" button is always available.
- When sending failed or connection is gone, "Delete" and "Edit" is available.
- When sending failed, "Resend" button is available.

### 🚧 TODO

- [x] recycler view problem
- [x] implement resend logic
- [x] change implementation of temporary messages in offline mode (dataStore -> Room)
   - [x] will have to change a lot from https://github.com/nextcloud/talk-android/pull/4251
- [x] change usage of TemporaryMessageViewHolder (use for sending failed messages and "no-internet-connection" mode)
- [x] check "Unread messages" bug
- [x] fix duplicate "Today"-bug (or any other date indicator)
- [x] add logic to show/hide Resend button
- [x] add "copy" logic
- [x] uncomment retry logic (for now commented out to get failed messages on bad connection)
- [x] undo ReadStatus logic??
- [x] fix that temp messages are swapped for a moment
- [x] when editing temp message, keep it's position
- [x] comment in openHelperFactory
- [x] DB migration ok?
- [x] check if parent message handling still works
- [x] change click handling
- [x] isSilent (also add to DB migration)
- [x] try to remove flickering of status (caused by coroutine scope for networkMonitor.isOnline)

for followup (maybe not directly related to this PR, but happened during developing):
- [ ] add logic to check for bad internet connection. Show a warning?
    - e.g. to add solution for StreamResetException when pulling chat messages
- [ ] fix loading spinner overlay with chat messages (not sure if related to this PR)

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
> Super nice! :) Small design feedback:
> 
>     * [ ]  Instead of a loading spinner, we could use a [schedule symbol](https://fonts.google.com/icons?icon.query=time&selected=Material+Symbols+Outlined:schedule:FILL@0;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23222222) to make it less distracting (similar to Whatsapp and Telegram, they show a very similar icon)
> 

We have loading spinners on web, desktop app and iOS, so i think we should have it for android as well..?


>     * [ ]  Instead of the warning icon in the screenshot, we can use the [error symbol](https://fonts.google.com/icons?icon.query=warning&selected=Material+Symbols+Outlined:error:FILL@0;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23222222), also for consistency with other apps (eg. SMS app in Android)
> 

fine for me

>     * [ ]  I'm wondering if we could the same error symbol for even the "no connection" state, as it is just another type of error where you can edit/resend etc. The red banner is already very eye catching, and I haven't seen a similar icon in any other apps, so I'm thinking it might confuse users. What do you think? :)

In "no connection" state, the resend button is not available. When using the same icon for both states it would be confusing when the resend button is available or not. So i suggest to have seperate icons. 
Not sure if the schedule icon suggested by @alperozturk96 is the best suggestion for it.. We have to keep in mind that at some point we want to have scheduled messages and we also need an icon for it..
The iOS app uses also a loading spinner when being offline, but i think all the movements are too much when having multiple messages pending..

Maybe @nickvergessen has opinions on this?

nimishavijay:
> We have loading spinners on web, desktop app and iOS, so i think we should have it for android as well..?

I would say for Android we follow the standard of other apps on Android so it looks as native as possible :) also because of what you mentioned 

> but i think all the movements are too much when having multiple messages pending

So I'd suggest the schedule icon instead of the loading spinner 

> We have to keep in mind that at some point we want to have scheduled messages and we also need an icon for it..

Taking a quick look at Signal and Telegram, a calendar icon is used in those, so we could do the same

> the resend button is not available. When using the same icon for both states it would be confusing when the resend button is available or not. So i suggest to have seperate icons.

Oh I see! Then it makes sense.

nickvergessen:
> > Instead of a loading spinner, we could use a [schedule symbol](https://fonts.google.com/icons?icon.query=time&selected=Material+Symbols+Outlined:schedule:FILL@0;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23222222) to make it less distracting (similar to Whatsapp and Telegram, they show a very similar icon)
>
> We have loading spinners on web, desktop app and iOS, so i think we should have it for android as well..?

I think we shouldn't use the schedule icon as scheduling is still a planned feature and it would be confusing?

nimishavijay:
> I think we shouldn't use the schedule icon as scheduling is still a planned feature and it would be confusing?

Also my concern with using a schedule icon for scheduled messages is that in other apps the schedule icon is usually used for messages that have not been sent. So it might cause the user to be confused about whether the message is scheduled or not sent, do you know what I mean?

nickvergessen:
I just checked some other messengers and the results are ... mixed...

Threema: ⬆️ Loading ✉️ Sent to server 📥 Received by contact
Signal: ◌ (Spinner) Loading ✔ Sent to server ✔✔ Received by contact
WhatsApp: 🕒 Loading ✔ Sent to server ✔✔ Received by contact

jancborchardt:
> I just checked some other messengers and the results are ... mixed...
> 
> Threema: ⬆️ Loading ✉️ Sent to server 📥 Received by contact
> Signal: ◌ (Spinner) Loading ✔ Sent to server ✔✔ Received by contact
> WhatsApp: 🕒 Loading ✔ Sent to server ✔✔ Received by contact

Also talked about this in the design team call – it’s good to follow standards here, and WhatsApp and Signal seem to be basically the same, so would be good to follow this.

nickvergessen:
> it’s good to follow standards here, and WhatsApp and Signal seem to be basically the same, so would be good to follow this.

We are only discussing the first icon atm as we are adding/touching that feature, and that's exactly the one that is different.
But okay let's go for the clock then

nickvergessen (approved):
:elephant:

mahibi:
not backported.
will be released with v21.x

PR #4436, opened 2024-11-12
fix to provide all call emojis

mahibi:
As the number of provided emojis grew (from 10 to 12), there was a bug that only one emoji was shown.

Putting all 12 emojis in one fixed row would have been too close, so it's implemented to scroll them horizontally

🏚️ Before 

grafik

🏡 After

[A](https://github.com/user-attachments/assets/6dac5349-cbbb-47d1-b0e8-b1d1ed5468c7)

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
I tested this with my Samsung A52 and it shows all emoji's and they are packed. Please see the screenshot.

all_emoji_in_call

sowjanyakch (approved):

PR #4448, opened 2024-11-14
Bugfix/noid/fix websocket connection for chat

mahibi:
### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2448,7 +2448,7 @@ class ChatActivity :
         currentlyPlayedVoiceMessage?.let { stopMediaPlayer(it) } // FIXME, mediaplayer can sometimes be null here
 
         adapter = null
-        Log.d(TAG, "inConversation was set to false!")
+        disposables.dispose()
     }
 
     private fun joinRoomWithPassword() {
@@ -2513,7 +2513,7 @@ class ChatActivity :
                 ?.observeOn(AndroidSchedulers.mainThread())
                 ?.subscribe(object : Observer<SignalingSettingsOverall> {
                     override fun onSubscribe(d: Disposable) {
-                        // unused atm
+                        disposables.add(d)
                     }
 
                     override fun onNext(signalingSettingsOverall: SignalingSettingsOverall) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt
```diff
@@ -391,7 +391,7 @@ class WebSocketInstance internal constructor(
                 roomToken == currentRoomToken &&
                 normalBackendSession == currentNormalBackendSession &&
                 federation?.roomId == currentFederation?.roomId &&
-                federation?.nextcloudServer == federation?.nextcloudServer
+                federation?.nextcloudServer == currentFederation?.nextcloudServer
```
danxuliu, line 394: :see_no_evil: What I was thinking on... Thanks!

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -2514,57 +2503,78 @@ class ChatActivity :
...
-                    if (signalingSettingsOverall.ocs!!.settings!!.externalSignalingServer == null ||
-                        signalingSettingsOverall.ocs!!.settings!!.externalSignalingServer?.isEmpty() == true
-                    ) {
-                        return
+            )
+                .subscribeOn(Schedulers.io())
+                ?.observeOn(AndroidSchedulers.mainThread())
+                ?.subscribe(object : Observer<SignalingSettingsOverall> {
+                    override fun onSubscribe(d: Disposable) {
+                        disposables.add(d)
```
danxuliu, line 2516: Just to be sure, that means that if the signaling settings are received once the activity was closed `onNext` will not be executed (so the Websocket instance will not be got, not the room joined, nor the listeners added), is that correct? If not then that should be somehow prevented, but I guess that is exactly the purpose of the disposable :-)

danxuliu (approved):
I have not tested, but the changes make sense and look good :+1: 

Approved, as I guess that the disposable does what I expected it to do, but please check the comment before merging :-)

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
mahibi: yes that's the purpose. won't be necessary anymore soon when we switch from rxJava to coroutines

PR #4463, opened 2024-11-18
Save push notified messages to the database without entering the app

### LINKED ISSUE #4379: Load messages and conversation in background when push message is received
when a push message is received
- the offline messages should be updated immediately
- the offline conversations should be updated immediately
in background.

So when entering the chat, there is no need to fetch messages from server but the initial loading from server can be skipped. This will improve user experience when entering a chat as it feels like everything is already in place and doesn't have to be loaded.
Also, this will be useful to have data offline as soon as possible in case the connection is lost.

This will require to move some logic from OfflineFirstChatRepository to ViewModel or a DomainModel so the Repository can also be used from NotificationWorker.

- [ ] discuss if fetching messages and conversations for every push message is causing too much load for the server?
- [ ] is fetching conversation list necessary? Clicking notification directs to the chat and skips the conversation list

rapterjet2004:
- fixes #4379 

### 🚧 TODO

- [ ] Fix 412 error on pulling chatMessages

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/data/ChatMessageRepository.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/data/ChatMessageRepository.kt
@@ -10,6 +10,7 @@ package com.nextcloud.talk.chat.data
 import android.os.Bundle
 import com.nextcloud.talk.chat.data.io.LifecycleAwareManager
 import com.nextcloud.talk.chat.data.model.ChatMessage
+import com.nextcloud.talk.data.database.model.ChatMessageEntity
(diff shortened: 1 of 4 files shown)


mahibi:
@rapterjet2004 Sorry, i haven't seen in time you picked this issue. 🙈
I fear working here now might cause problems as i plan to continue with https://github.com/nextcloud/talk-android/pull/4422 in the next days. There would be complicated merge conflicts. 
Could you pause this and work on other parts not related to the offline feature?

PR #4474, opened 2024-11-21
Replace Popup bubble with Material Button

### LINKED ISSUE #4391: get rid of com.nextcloud.ui.popupbubble.PopupBubble
In PR https://github.com/nextcloud/talk-android/pull/4390 (commit https://github.com/nextcloud/talk-android/pull/4390/commits/f4340f3f97c14de89597f8721582cb7e09057a8b)

the com.nextcloud.ui.popupbubble.PopupBubble (forekd from https://github.com/webianks/PopupBubble)

was removed to show the unread messages button.

It is still used in ConversationsListActivity. It should be replaced so we can get rid of it (it's very outdated)

sowjanyakch:
Resolve #4391

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi (changes requested):
the behavior of the button is now different for me. When clicked, the list will scroll as expected but the button won't disappear.

Only after a pull to refresh it seems to work that the button disappears for the next time it's clicked..

mahibi:
please remove the dependency of the popup bubble

mahibi:
> the behavior of the button is now different for me. When clicked, the list will scroll as expected but the button won't disappear.
> 
> Only after a pull to refresh it seems to work that the button disappears for the next time it's clicked..

still the same for me

mahibi:
works for me now.
I added an animation for the button and changed it's design a bit so it feels similar to like it was before..

mahibi (approved):

mahibi:
So.. we can delete repo https://github.com/nextcloud/PopupBubble soon (just wait for the major release i would say)

PR #4476, opened 2024-11-22
Add permanent warning when notifications are not set up correctly

mahibi:
Show a warning whenever something regarding notifications is not set up correctly.

known bug is https://github.com/nextcloud/talk-android/issues/4484 (won't be fixed for now if no one comes up with an idea..)

### 🖼️ Screenshots/ Videos

Warning when notifications are not set up correctly

grafik

- On first login, ask directly with popups for notification permission and to ignore battery optimization.
- If this is declined or any other setting is wrong, show a permanent warning. 
   - Also: When using mobile device management MDM, it may happen that some permissions are not set up while preparing the phones and users won't see any dialogs on first use. With this change, they now see the permanent warning and are guided what changes need to be made.
- When clicking on the warning, settings are opened and things that need to be changed will blink and have red subtitles.
- When any permission is revoked it will show a warning again in conversation list.
- Allow user to disable the warning in conversation list if anything is wrong.
- If a server has no notification app installed, this will be shown for the regarding account.

https://github.com/user-attachments/assets/caca4e1b-dcb7-440d-b5ba-7030d15d4e32

- When logging in and allowing all permissions, no warning is necessary.

https://github.com/user-attachments/assets/e74d7189-9dff-467d-bf24-855f0119a711

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -101,7 +101,6 @@ How to translate with transifex:
     <string name="nc_settings_no_talk_installed">Talk app is not installed on the server you tried to authenticate against</string>
     <string name="nc_settings_account_updated">Your already existing account was updated, instead of adding a new one</string>
     <string name="nc_account_scheduled_for_deletion">The account is scheduled for deletion, and cannot be changed</string>
-    <string name="nc_settings_notification_sounds">Notification sounds</string>
     <string name="nc_settings_notification_sounds_post_oreo">Notifications</string>
     <string name="nc_settings_call_ringtone">Calls</string>
     <string name="nc_settings_call_ringtone_key" translatable="false">call_ringtone</string>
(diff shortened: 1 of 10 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -181,6 +181,8 @@ How to translate with transifex:
 
     <string name="nc_ignore_battery_optimization_dialog_title">Ignore battery optimization</string>
     <string name="nc_ignore_battery_optimization_dialog_text">Battery optimization is not ignored. This should be changed to make sure that notifications work in the background! Please click OK and select \"All apps\" -> %1$s -> Do not optimize</string>
+    <string name="nc_show_ignore_battery_optimization_hint_title">Show battery optimization hint</string>
+    <string name="nc_show_ignore_battery_optimization_hint_description">When the battery optimization is not ignored, show a hint</string>
```
nickvergessen, line 185:
```suggestion
    <string name="nc_show_ignore_battery_optimization_hint_description">When the battery optimization is not disabled or ignored, show a hint</string>
```

nickvergessen (commented):
Should it say "not disabled" or "not ignored"?
I think the warning should say disabled?

app/src/main/res/values/strings.xml
Antreesy: Don't know about other devices, but mine (Xiaomi) suggests following for 'Battery saver`:
- `No restrictions`: Battery saver doesn't restrict app's activity
- `Battery saver`: ... saver will treat important apps selectively

So I would expect a wording like this:
`When the battery optimization is enabled (not ignored) for this app, show a hint`

mahibi:
> Should it say "not disabled" or "not ignored"? I think the warning should say disabled?

i will change the warning so it will summarize if anything is wrong with notifications, not only battery optimization. So it will include if notification are disabled in general or if notification app is not installed on server..
So the wording of the hint will be different in the end.

nickvergessen:
Can you update the screenshots?

rapterjet2004 (approved):

sowjanyakch (approved):

nickvergessen (approved):

PR #4482, opened 2024-11-25
Add playback speed control to voice messages

arkascha:
This is an implementation of the feature request in issue 3786: 
"Allow voice messages to be played at different speeds"

This PR should not be considered final. There are a few questions to be answered first, I cannot do that myself, however. I need someone from inside the project for that. So this is more a "request for comment" than a PR ;-)

The current implementation does not at all offer a fancy design. The UI is as simplistic as possible. 
I could very well imagine a button that extends when clicked and offers the options in parallel instead of cycling through the offered options as it is currently implemented. But that certainly is something a design expert has to decide. 

The current implementation has 4 speeds it offers (opposed to WhatsApp offering 3): 0.8x, 1.0x, 1.5x, 2.0x. I added the slow variant since that could sometimes help to understand messages with bad acoustics. 
The chosen speed is persisted on a per-user base. This appeared intuitive to me, since usually individual users keep their typical talking speed. Which means that if there is a need to change the playback speed that change probably makes sense for all voice messages of that user. 

I would need some advice about the testing strategy in this project ... When trying to implement tests I found that next to no tests and test structure exist. I assume I miss something here ...

Seems I found a bug in how the message view holders are handled: their onBind() method is called over and over again when a voice message is played. That might be triggered by updating the progress slider. This behaviour repaints the view holder millions of times which actually creates quite a load on the system ...

I am thankful for any feedback on this! 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
screenshot_nextcloud-andoird-talk-voice-message-before | screenshot_nextcloud-andoird-talk-voice-message-after

### 🚧 TODO

- [x] decide about design requirements or requests for the new button
- [x] clearify state of unit and instrumented tests
- [x] decide what to do with pre existing onBind() bug in message holders
- [x] Translations

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingVoiceMessageViewHolder.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/messages/IncomingVoiceMessageViewHolder.kt
@@ -1,6 +1,7 @@
 /*
  * Nextcloud Talk - Android Client
  *
+ * SPDX-FileCopyrightText: 2024 Christian Reiner <foss@christian-reiner.info>
  * SPDX-FileCopyrightText: 2021 Andy Scherzinger <info@andy-scherzinger.de>
  * SPDX-FileCopyrightText: 2021 Tim Krüger <t@timkrueger.me>
  * SPDX-FileCopyrightText: 2021 Marcel Hibbe <dev@mhibbe.de>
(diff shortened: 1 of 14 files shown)


arkascha:
@rapterjet2004 It seems we had a mid air collision ... I received an email message pointing to a commit you added, but I cannot find that commit. I am afraid I suppressed it by a force push I made inside my repository from where I created this PR. Sorry for that, I did not expect it is possible to add a commit to a PR, that is a new concept for me. Could you please cherry-pick that commit again? 
Sorry!

arkascha:
Is there anything that needs to be done on my side? 
As far as I can tell the failed checks are caused by issues completely independent from this branch.

AndyScherzinger:
I guess no, just waiting for @rapterjet2004 or @mahibi

The failing check will be fixed once #4502 gets merged

AndyScherzinger (approved):
Tested successfully

arkascha:
As I understand it you are accepting this change without any implemented tests. A bit surprising, that is why I asked about your test strategy when initially opening the PR. 
Now, each project has it's own strategy. I would like to understand yours, though. Is it true that you do not implement tests inside the application itself?

mahibi:
works great. Fine for me to merge! Thank you @arkascha !

PR #4497, opened 2024-11-28
Add clear icon for search view in open conversations

### LINKED ISSUE #4490: Add ✖️ to clear search/filter input in open conversations
## Filtering conversations list
Image

## Filtering open conversations list
Image

sowjanyakch:
Resolve #4490

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/openconversations/ListOpenConversationsActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/openconversations/ListOpenConversationsActivity.kt
@@ -58,7 +58,7 @@ class ListOpenConversationsActivity : BaseActivity() {
         setContentView(binding.root)
         setupSystemColors()
         viewThemeUtils.platform.colorImageView(binding.searchOpenConversations, ColorRole.ON_SURFACE)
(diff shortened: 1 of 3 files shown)


mahibi (changes requested):
Could you have a closer look how much work it is to use a SearchView (https://developer.android.com/reference/com/google/android/material/search/SearchView) here? I think it would make more sense as it's intended to use with the appbar.

If SearchView is not possible for whatever reason, then you should wrap the 
TextInputEditText inside a TextInputLayout which will also give you the possibility to use built in cancel button etc.

mahibi (changes requested):
when typing, the result list should also be updated if there searched string doesn't match anymore.

For example there are 3 rooms called "Talk".
When i search for "Tal", all 3 rooms are shown.
But when i add a letter that doesn't match, the results should disappear. So when i type "Talx", the old results are still shown.

sowjanyakch:
I will implement the search bug in a different PR.

mahibi (approved):

PR #4521, opened 2024-12-05
Out of office

### LINKED ISSUE #4477: Show "Out of office" message in 1-1 conversations
- Ref https://github.com/nextcloud/spreed/issues/10599#issuecomment-1810359864
- iOS implementation https://github.com/nextcloud/talk-ios/pull/1890

sowjanyakch:
Resolve #4477

out_of_office_view

1. Scroll through the office view to find the absence message if the user has set a long absence message.
2. Show the vacation replacement only if the user has set one otherwise hide the view.
3. Clicking on the vacation replacement user’s avatar/ name will allow you to join one-on-one conversation with the designated vacation replacement.
4. Add start and end dates of vacation.
5. If the user is out of office for only one day, display the message “User is out of office today” without showing the start and end dates of the vacation.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -1080,15 +1080,16 @@ class ChatActivity :
                 }
                 is ChatViewModel.OutOfOfficeUIState.Success -> {
                     binding.outOfOfficeContainer.visibility = View.VISIBLE
-                    binding.outOfOfficeContainer.findViewById<TextView>(R.id.status).text = String.format(
+                    binding.outOfOfficeContainer.findViewById<TextView>(R.id.userAbsenceShortMessage).text = String.format(
                         context.resources.getString(R.string.user_absence),
                         uiState.userAbsence.userId
                     )
-                    binding.outOfOfficeContainer.findViewById<TextView>(R.id.replacement).text = String.format(
-                        context.resources.getString(R.string.user_absence_replacement),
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -3808,6 +3919,24 @@ class ChatActivity :
...
     }
 
+    fun joinOneToOneConversation(userId: String) {
+        val apiVersion =
+            ApiUtils.getConversationApiVersion(conversationUser!!, intArrayOf(ApiUtils.API_V4, 1))
+        val retrofitBucket = ApiUtils.getRetrofitBucketForCreateRoom(
+            apiVersion,
+            conversationUser?.baseUrl!!,
+            "1",
+            "users",
```
mahibi, line 3929: can you add constants for
"1"
and
"users"

and also use them in the other places in this class where
getRetrofitBucketForCreateRoom is used?

(better would be to have these types centralized somewhere but doesn't have to be done in this PR)

mahibi (approved):

rapterjet2004 (changes requested):
Looks good!

1. Is that the replacement views are still there even if I didn't set a replacement. This makes it seem like it's still loading even when they're not
Screenshot 2024-12-13 at 7 46 17 AM

rapterjet2004 (approved):

PR #3670, opened 2024-02-27
add compatibility for janus 1.x

mahibi:
fix #2704

Ref:
- server PR: https://github.com/nextcloud/spreed/pull/11660
- iOS PR: https://github.com/nextcloud/talk-ios/pull/1557

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #4536, opened 2024-12-11
Improve handling of data channels

danxuliu:
Follow up to #3670

This was part of an upcoming pull request but was extracted for easier reviewing. It adjusts some data channel related code that is either needed for the other pull request (queuing messages if the data channel was not open yet), was incorrect (not disposing of all data channels, sending messages on data channels other than _status_) or could be improved (missing logs, documentation and tests). Please refer to the individual commit messages for details.

As data channels involve several threads (the events are received in the so called WebRTC signaling thread (unrelated to the thread were signaling messages are received), but the PeerConnectionWrapper can be used from RxJava executors, the main thread, the WebRTC signaling thread itself...) synchronization was needed in certain points. There are probably better ways to do the synchronization, but for simplicity I just resorted to synchronized methods/blocks and catching exceptions. Similarly the unit tests for the possible synchronization problems were done in a brute force approach, just repeating the test many times, so all this has a lot of room for improvements... In any case, thread safety should be kept in mind for future changes, as there are probably other areas of the code that might need to be adjusted :shrug:

Besides that, adding the logs unveiled a bug that seems to have been there since forever: when the HPB is used the nick is tried to be sent even after the call activity is destroyed (and the `PeerConnectionWrapper` is also leaked). There is something wrong with the use of [`interval` and `repeatUntil`](https://github.com/nextcloud/talk-android/blob/f820277779ced24c9fea877cb6df3ef853bfca2e/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt#L2552-L2553), but that will be rewritten anyway in the upcoming pull request, so for now it stays as is.

Finally, one thing that still should be implemented (but it will not be implemented either in the other pull request) is disabling data channels so they are not even negotiated when not needed (for example, screen connections). This is not a big deal, however, as the data channel will be blocked by the other end, and even if it is opened it should not cause any trouble unless the Android app sends conflicting messages :-)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -1174,12 +1174,12 @@ class CallActivity : CallBaseActivity() {
         if (isConnectionEstablished && othersInCall) {
             if (!hasMCU) {
                 for (peerConnectionWrapper in peerConnectionWrapperList) {
-                    peerConnectionWrapper.sendChannelData(DataChannelMessage(isSpeakingMessage))
+                    peerConnectionWrapper.send(DataChannelMessage(isSpeakingMessage))
                 }
             } else {
                 for (peerConnectionWrapper in peerConnectionWrapperList) {
                     if (peerConnectionWrapper.sessionId == webSocketClient!!.sessionId) {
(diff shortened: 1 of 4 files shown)


mahibi (approved):

mahibi:
Thank you @danxuliu !!!

PR #4579, opened 2025-01-07
Abstracting away media player

### LINKED ISSUE #4563: Abstract away the MediaPlayer logic in ChatActivity to MediaPlayerManager
Will likely require rewriting the MediaPlayerManager to be compatible with the more complicated functionality in ChatActivity (cycling through messages, marking messages as played, restoring state etc). I think a state machine (with an exposed enum) and some callbacks to observe should do the trick.

Will be helpful for eventually removing ChatKit, and replacing it with a full compose solution. The more functionality handled outside of ChatActivity the better.

Functionality to replicate:
- #3704
- #4549 
- #4482

rapterjet2004:
- fixes #4563 

### 🖼️ Screenshots
Example of background play

### 🚧 TODO

- [x] Add functionality to MediaPlayerManager
- [x] Create Background Voice Message View
- [x] Rewrite ChatActivity to use MediaPlayerManager
    - [x] play
    - [x] stop
    - [x] pause
    - [x] seek
    - [x] Bug: pausing and replaying causes the message to loop
    - [x] Bug: On completion the message is not reset 
- [x] Resumes playback of voice messages on screen rotation
- [x] Continue playing a voice message in background
    - [x] Bug: Enter into the original conversation after lifecycle change. Seek Bar not updated.
    - [x] Bug: After orientation change. Seek Bar not updated. Scope is canceled.
    - [x] Add a reference to the conversation that the msg is in too. Maybe avatar or name
    - [x] Bug: Occasionally, card shows up when played messaged is paused. Need to look into this
- [x] Implement Voice Message card
    - [x] play
    - [x] close
    - [x] pause
    - [x] avatar
    - [x] animation
    - [x] Bug: Pausing it destroys it
- [x] Auto play consecutive audio messages
    - [x] prefetch audio messages
    - [x] show fetched audio durations
    - [x] temporarily mark as listened audio messages
    - [x] Bug: Durations no longer count down, also they should match the playback speed
    - [x] Bug: pausing resets the duration for some reason
- [x] Get playback speed controls working
    - [x] bug: Works, but UI doesn't update ( check cycling too ) 
- [x] Bug: It crashes when playing the Voice Recording preview

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
great :+1: 
won't be backported to 21.0

While it's of course an improvement to have the player inside the conversation list, i wonder if it's much more work to have it as a notification instead (Like Whatsapp and telegram have it..)?

rapterjet2004:
> great :+1: 
> won't be backported to 21.0
> 
> While it's of course an improvement to have the player inside the conversation list, i wonder if it's much more work to have it as a notification instead (Like Whatsapp and telegram have it..)?

Could be done,  but it would require adding a new dependency and a bit more boilerplate to work with notification.  It should be in a separate PR

mahibi (approved):

PR #4591, opened 2025-01-10
fix translation for polls_option_delete

mahibi:
### 🏁 Checklist

- [ ] 📅 Milestone is set


tobiasKaminsky (approved):

mahibi:
it's expected that analysis fails as translation string changed and needs to be sync when this PR is merged

Issue #4609, opened 2025-01-15
Migrate UsersRepository+UserManager to coroutines


mahibi:
Classes related to Users, e.g.
- UsersRepository
- UserManager 
- CurrentUserProviderImpl
have to be migrated from RxJava to coroutines.

Similar to the current implementation with RxJava, in CurrentUserProviderImpl a coldFlow to hotFlow conversion has to be made so the database is not queried every time.

userManager.currentUser.blockingGet() (or it's coroutine pendent) should not be queried throughout the code but CurrentUserProviderImpl should be used instead.

PR #4611, opened 2025-01-15
hopefully fix NPE for currentUser in addToConversationItems

### LINKED ISSUE #4610: NPE for currentUser in ConversationsListActivity
```
Exception java.lang.NullPointerException:
  at com.nextcloud.talk.conversationlist.ConversationsListActivity.addToConversationItems (ConversationsListActivity.kt:851)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity.access$addToConversationItems (ConversationsListActivity.kt:151)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invokeSuspend (ConversationsListActivity.kt:394)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invoke (Unknown Source:8)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invoke (Unknown Source:4)
```

v20.1.0

mahibi:
fix #4610

It seems there could have been a race condition because currentUser was initialized too late.
 I was not able to reproduce but hopefully this fixes it:
 - initialize currentUser earlier (moved from onResume to onCreate)
 - use currentUserProvider instead userManager

NPEs were reported via gplay:

1)
```
Exception java.lang.NullPointerException:
  at com.nextcloud.talk.conversationlist.ConversationsListActivity.addToConversationItems (ConversationsListActivity.kt:851)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity.access$addToConversationItems (ConversationsListActivity.kt:151)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invokeSuspend (ConversationsListActivity.kt:394)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invoke (Unknown Source:8)
  at com.nextcloud.talk.conversationlist.ConversationsListActivity$initObservers$5$1.invoke (Unknown Source:4)
```

2)
```
Exception java.lang.RuntimeException:
  at android.app.ActivityThread.performResumeActivity (ActivityThread.java:5427)
  at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:5508)
  ...
Caused by java.lang.NullPointerException:
  at com.nextcloud.talk.conversationlist.ConversationsListActivity.shouldShowNotificationWarning (ConversationsListActivity.kt:1557)
```

---

Long term solution: https://github.com/nextcloud/talk-android/issues/4609 + better usage of MVVM

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
@@ -235,6 +235,8 @@ class ConversationsListActivity :
         super.onCreate(savedInstanceState)
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
 
+        currentUser = currentUserProvider.currentUser.blockingGet()
+
         conversationsListViewModel = ViewModelProvider(this, viewModelFactory)[ConversationsListViewModel::class.java]
 
         binding = ActivityConversationsBinding.inflate(layoutInflater)
@@ -284,7 +286,7 @@ class ConversationsListActivity :
         if (!eventBus.isRegistered(this)) {
             eventBus.register(this)
         }
-        currentUser = userManager.currentUser.blockingGet()
+
         if (currentUser != null) {
             if (isServerEOL(currentUser!!.serverVersion?.major)) {
                 showServerEOLDialog()
@@ -309,7 +311,7 @@ class ConversationsListActivity :
             fetchRooms()
             fetchPendingInvitations()
         } else {
-            Log.e(TAG, "userManager.currentUser.blockingGet() returned null")
+            Log.e(TAG, "currentUser was null")
             Snackbar.make(binding.root, R.string.nc_common_error_sorry, Snackbar.LENGTH_LONG).show()
         }
 
@@ -421,7 +423,7 @@ class ConversationsListActivity :
     }
 
     fun filterConversation() {
-        val accountId = UserIdUtils.getIdForUser(userManager.currentUser.blockingGet())
+        val accountId = UserIdUtils.getIdForUser(currentUser)
         filterState[FilterConversationFragment.UNREAD] = (
             arbitraryStorageManager.getStorageSetting(
                 accountId,
@@ -1365,7 +1367,7 @@ class ConversationsListActivity :
                     val conversation = clickedItem.model
                     conversationsListBottomDialog = ConversationsListBottomDialog(
                         this@ConversationsListActivity,
-                        userManager.currentUser.blockingGet(),
+                        currentUser!!,
                         conversation
                     )
                     conversationsListBottomDialog!!.show()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -235,6 +235,8 @@ class ConversationsListActivity :
         super.onCreate(savedInstanceState)
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
 
+        currentUser = currentUserProvider.currentUser.blockingGet()
```
alperozturk96, line 238: Rather than blocking the thread, you can observe the result on main thread.

`private var currentUserDisposable: Disposable? = null`


```
override fun onCreate(savedInstanceState: Bundle?) {
        NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)

        currentUserDisposable = currentUserProvider.currentUser
            .subscribeOn(Schedulers.io())
            .observeOn(AndroidSchedulers.mainThread())
            .subscribe { user ->
                currentUser = user

                conversationsListViewModel = ViewModelProvider(this, viewModelFactory)[ConversationsListViewModel::class.java]

                binding = ActivityConversationsBinding.inflate(layoutInflater)
                setupActionBar()
                setContentView(binding.root)
                setupSystemColors()
                viewThemeUtils.material.themeCardView(binding.searchToolbar)
                viewThemeUtils.material.themeSearchBarText(binding.searchText)

                forwardMessage = intent.getBooleanExtra(KEY_FORWARD_MSG_FLAG, false)
                onBackPressedDispatcher.addCallback(this, onBackPressedCallback)

                initObservers()
            }
    }
```

```
public override fun onDestroy() {
        super.onDestroy()
        dispose(null)
        if (searchViewDisposable != null && !searchViewDisposable!!.isDisposed) {
            searchViewDisposable!!.dispose()
        }

        currentUserDisposable?.dispose()
    }
```

alperozturk96 (changes requested):

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
mahibi: The 
[CurrentUserProviderImpl](https://github.com/nextcloud/talk-android/blob/f5a4274e545f506861dc7baf7f92ec5e73d674f6/app/src/main/java/com/nextcloud/talk/utils/database/user/CurrentUserProviderImpl.kt)
was implemented so that retrieving the user via

`currentUserProvider.currentUser.blockingGet()`

is reasonable.

Given the amount of how many times the user is queried in code, adding observers and disposables in all places would cause so much boilerplate.

I agree that it should be replaced with more modern coroutine handling, but until then i think usage of 
currentUserProvider.currentUser.blockingGet()
is okay.

Does this make sense or is this a no go for you @alperozturk96 ?

alperozturk96: [blockingGet()](https://reactivex.io/RxJava/2.x/javadoc/2.0.8/io/reactivex/Maybe.html#blockingGet()) does not use any particular Schedulers. It means it can run on MainThread and **may** cause an issue.

Ideally, CurrentUserProviderImpl should handle this and do the necessary jobs. You don't have to repeat for every class.

This is how you should handle RxJava code. I just wanted to share.

But since this PR focuses on the NPE fix, we can move forward.

alperozturk96 (approved):

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
mahibi: Thanks to clarify, we can have a closer look together maybe when replacing the CurrentUserProviderImpl...

PR #4624, opened 2025-01-16
Follow up fixes for 'Archive Conversations

### LINKED ISSUE #4607: Archive conversation option follow ups
- It should be possible to archive from the context menu on the conversation list
- The option somehow does not look like a option, maybe because other things around it are red and participant list?
- The option can not be toggled multiple times: After turning it on, you need to leave the details in order to be able to turn if off again. Pressing it a second time without leaving and coming back does not do anything.

### LINKED ISSUE #4608: Archive conversation follow ups #2
1. After filtering for archived conversations and forgetting about it, i was shocked for some seconds why the conversations don't load. Turned out i had the filter enabled but no conv is archived. In this case we should give a hint about it as a text in middle of the screen. 
Please check how it's done on iOS.

2. the "Unread mentions" bubble is not updated when filtering

rapterjet2004:
- fixes #4607
- fixes #4608  

### 🚧 TODO

- [x] It should be possible to archive from the context menu on the conversation list
- [x] The option somehow does not look like a option, maybe because other things around it are red and participant list?
- [x] The option can not be toggled multiple times: After turning it on, you need to leave the details in order to be able to turn if off again. Pressing it a second time without leaving and coming back does not do anything.
- [x] After filtering for archived conversations and forgetting about it, i was shocked for some seconds why the conversations don't load. Turned out i had the filter enabled but no conv is archived. In this case we should give a hint about it as a text in middle of the screen.
- [x] the "Unread mentions" bubble is not updated when filtering.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
@@ -820,6 +820,7 @@ class ConversationInfoActivity :
                     binding.archiveConversationTextHint.text = resources.getString(R.string.unarchive_hint)
                 }
             }
+            viewModel.getRoom(conversationUser, conversationToken)
         }
 
         if (conversation!!.hasArchived) {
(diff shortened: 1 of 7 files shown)


mahibi:
@rapterjet2004 works fine for me. Just to let you know, i made some changes to the PR:


before/after






before/after






added snackbar for archive/unrachive action from bottomDialog

mahibi (approved):

PR #4627, opened 2025-01-17
Rework can leave/delete conversation

### LINKED ISSUE #4570: Check `canLeaveConversation` and `canDeleteConversation` handling
If the attribute `canLeaveConversation` is there it should be respected no matter any other things. Same for `canDeleteConversation`. (Required for an upcoming feature where we change those values e.g. one-to-one can no longer be left, but only deleted: https://github.com/nextcloud/spreed/pull/14064 )

Also make sure that `canDeleteConversation` is used to determine if "Delete all message" is shown.

mahibi:
resolve #4570

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
@@ -782,7 +782,7 @@ class ConversationInfoActivity :
             if (CapabilitiesUtil.hasSpreedFeatureCapability(
                     spreedCapabilities,
                     SpreedFeatures.CLEAR_HISTORY
-                )
+                ) && conversationCopy.canDeleteConversation
             ) {
                 binding.clearConversationHistory.visibility = VISIBLE
             } else {


mahibi:
asked in chat:

>  Not sure if i understand https://github.com/nextcloud/talk-android/issues/4570 correctly.
> So if i understand correctly, regarding
> "If the attribute canLeaveConversation is there it should be respected no matter any other things. Same for canDeleteConversation."
> there is nothing to do as this is already the case:
https://github.com/nextcloud/talk-android/blob/35f9e339c9fb5da4ac21211407575d1e4ef954f9/app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt#L858-L868
> And regarding
"Also make sure that canDeleteConversation is used to determine if "Delete all message" is shown."
i made the change
https://github.com/nextcloud/talk-android/pull/4627/files
> Is that's all for the issue or did i get something wrong?

answer from Joas:

> then it's good. The web accidentally had an additional condition for one-to-one conversations to hide delete

rapterjet2004 (approved):

PR #4629, opened 2025-01-17
avoid shimmer animations overlay with chat messages

### LINKED ISSUE #4626: "Loading chat" shimmer effect not hidden after chat is loaded
...so chat and shimmer animations overlay each other.

Should be related to 

```
        this.lifecycleScope.launch {
            delay(DELAY_TO_SHOW_PROGRESS_BAR)
            if (adapter?.isEmpty == true && networkMonitor.isOnline.value) {
                binding.progressBar.visibility = View.VISIBLE
            }
        }
```
or whereever progressBar is set to hidden.

reported for 21.0.0 alpha 09

Already have some ideas how to fix..

mahibi:
fix #4626

The idea to show the progress bar (aka shimmer animation) was that it show not be shown at all if loading messages happens in less than a second (this was a better UX in my opinion).

The idea was to hide the progress bar when
ChatMessageStartState was triggered.

However there can be moment when adapter is still empty after ChatMessageStartState and if in this moment the DELAY_TO_SHOW_PROGRESS_BAR is reached before the adapter is actually filled, the overlay happens. It could be a solution to move the hiding of the progress bar, however then special cases might have to be handled.

For simplicity, the logic for DELAY_TO_SHOW_PROGRESS_BAR is removed. Progress bar is always shown without a delay so it wont be triggered on a later moment and can't overlay the chat.

### 🏁 Checklist

- [ ] 📅 Milestone is set


rapterjet2004 (approved):
@mahibi Noticed a bug with offline support - approving anyway for your convenience

1. turned on airplane mode
2. entered into a conversation without any offline messages saved
3. Shimmer present when it shouldn't be

Screenshot_20250122-102517


Also when you reenter the conversation the background icon is still there. ( _TBH I like the effect, reminds me of backgrounds on Telegram and Whatsapp. Chat backgrounds a future feature_?), but I don't know if it's related to your PR or mine when I implemented the feature some time ago. 

Screenshot_20250122-102657

rapterjet2004:
Oops, didn't realize auto-merge was on. My bad.

PR #4647, opened 2025-01-23
Improving search for chat messages

### LINKED ISSUE #4433: Improve search for chat messages
The message search was never implemented sufficiently. 

A user does not always get the result he is looking for, i.e. the message he is looking for is sometimes not found in the chat.

The overview will show the matched results, but when clicking on a result it depends if the message is available in the adapter. 

For older messages this is most likely not the case (one would have to scroll very often upwards before searching to have a chance it's included in the adapter), so only the last messages of the chat are shown instead to show the searched message. 

Loading all messages into the adapter at once is not an option for performance reasons. Same for doing it in a loop. (It may differ if the message is found in the database for the latest chatblock. Then it _might_ be an option to load a lot of data into the adapter, as it's quicker than doing online requests in a loop. But this might also have limitations).

Talk iOS seems to have the same challenge. The workaround there is to show a snippet of the chat in a search window. So there the searched message is shown and maybe 30(?) messages are shown before and after it, without being able to scroll further.
For talk android this would mean that chatblocks may be created that are not connected to the latest chatblock. This shouldn't be a problem when there is no option to scroll in the search result window.

We should implement the same for android talk or discuss if there are other ways to solve it.

rapterjet2004:
- fixes #4433 

https://github.com/user-attachments/assets/e0586514-9811-4a5e-91a3-8404220498e8

### 🚧 TODO
- [x] Got the dialog set up
- [x] Get API working
- [x] Adapter?
- [x] Improving Links
- [x] System Messages
- [x] getting the conversation name
- [x] Optimization.
- [x] mentions
- [x] Shimmer
- [x] make the highlight animation play only once
- [x] Occasionally stuck on shimmer, seems to be a network issue, with some kind of infinite loop going on with OKHttp. Perhaps I just need some error management, in case I'm not able to retrieve the messages.
- [x] refactor the MutableState<Boolean> logic
- [x] Image and Caption formatting the height as needed
- [x] Mention formatting, check android view
- [x] Closer analysis shows that I'm using state wrong, causing infinite loops throughout my code. Need to fix this.
- [x] Use string resources

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/.idea/inspectionProfiles/ktlint.xml
+++ b/.idea/inspectionProfiles/ktlint.xml
@@ -68,6 +68,7 @@
     </inspection_tool>
     <inspection_tool class="PreviewParameterProviderOnFirstParameter" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
     </inspection_tool>
     <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
--- a/app/src/main/java/com/nextcloud/talk/api/NcApiCoroutines.kt
+++ b/app/src/main/java/com/nextcloud/talk/api/NcApiCoroutines.kt
@@ -8,6 +8,7 @@
 package com.nextcloud.talk.api
 
 import com.nextcloud.talk.models.json.autocomplete.AutocompleteOverall
+import com.nextcloud.talk.models.json.chat.ChatOverall
 import com.nextcloud.talk.models.json.chat.ChatOverallSingleMessage
 import com.nextcloud.talk.models.json.conversations.RoomOverall
 import com.nextcloud.talk.models.json.generic.GenericOverall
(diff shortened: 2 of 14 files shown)


mahibi:
Looking forward to have this :) From a first look:

- [ ] highlight searched message would be nice
- [ ] scroll to message is missing (the searched message is contained but may be not shown on the screen)
- [ ] a bit more messages before and after would be nice (align with iOS)
- [ ] search inside a chat (upper right menu) is missing (and then only search inside this chat)
- [ ] read checkmarks are missing
- [ ] mentions don't work

- [ ] in the list, it may be not possible to get to know the date of the messages by scolling (only when there is a new day shown inside the chat). This is an improvement that could be also done in the main chat window: showing the date as fixed bubble on the top.

- [ ] images are cropped very much

web:


android search:



- [ ] time is sometimes shown vertically (digits are unter each other) or not at all but height of message is too high. See screenshot:

mahibi:
Great job Julius! :1st_place_medal: :+1: 

From my side this can be merged and improvements/fixes can be made in followup PRs. 
Some minor things:

- Maybe we had a misunderstanding here:

> search inside a chat (upper right menu) is missing (and then only search inside this chat)

I am talking about if you go into a chat, there is the search available in the upper right corner.
When searching there (instead in the main screen), the search you implemented should also be available, but limited to the chosen chat.



Adding the disabled call buttons and the disabled menu in the search results window doesn't make sense to me (i guess this is what you understood from my description?)


- Rotating the device leads to some wrong design & search window is closed.

- Some file icons are not yet displayed as expected. For example on sermo in the "Filetype examples" (e.g. icons for zip files..)

mahibi (approved):

PR #4744, opened 2025-02-19
use ubuntu-latest for analysis and codeql

mahibi:
Not sure if there are still valid reasons to use 24.04 Basically testing if github actions run in the PR and then may merge

Using ubuntu-latest would allow that the self-hosted runners can be used instead to only use the ones from github. This should speed up the checks...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -24,7 +24,7 @@ concurrency:
 
 jobs:
     analysis:
-        runs-on: [ubuntu-latest, self-hosted]
+        runs-on: ubuntu-latest
         steps:
             -   name: Setup variables
                 id: get-vars
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -24,7 +24,7 @@ concurrency:
 
 jobs:
     analysis:
-        runs-on: ubuntu-24.04
+        runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
In case of a successful test you'll actually need to update the files in nextcloud/android-config since they get synced down daily (see first comment line in the actions)

mahibi:
> In case of a successful test you'll actually need to update the files in nextcloud/android-config since they get synced down daily (see first comment line in the actions)

ah sure, now that you say it... 🙈

Anyway, according to @tobiasKaminsky there was made "some change" in ubuntu-24.04 for the android builds back then.
Is anything known about it and if it's still needed? May the current "Set Swap Space" error be related to this?

AndyScherzinger:
Maybe Tobias knows. The thing I had to change for the latest Ubuntu was about the command how to install python packages. So not related to this one.

mahibi:
with
runs-on: [ubuntu-latest, self-hosted]
GARM is finally used. Without "self-hosted", it seems GARM was never used but i will revert "self-hosted" and trigger the actions more often and see if GARM is used

mahibi:
changes were implemented in https://github.com/nextcloud/android-config/pull/243 and will be synced back to android apps.

PR #4760, opened 2025-02-24
show a snackbar with meaningful message when server is in maintenance mode.

### LINKED ISSUE #2737: Server is in maintenance mode warning
Currently if the Nextcloud server is in maintenance and when launching the Nextcloud Talk client get a popup "There was a problem lading your chats" 

I think it would be best to have a pop stating that if the server is in maintenance mode. I would still keep that other message incase the server is not in maintenance mode and there was a problem  loading chats

sowjanyakch:
Resolve #2737

Show a snackbar with meaningful message when the server is in maintenance mode instead of dialog to perform add or remove account operations. While these options are available already in UI even when the server is in maintenance mode, a simple snackbar can be used.

server_maintenance_mode

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -656,8 +656,7 @@ How to translate with transifex:
     <string name="nc_dialog_outdated_client_description">The app is too old and no longer supported by this server. Please update.</string>
     <string name="nc_dialog_outdated_client_option_update">Update</string>
     <string name="nc_switch_account">Switch account</string>
-    <string name="nc_dialog_maintenance_mode_description">Server is currently in maintenance mode. Please try again
-        later.</string>
+    <string name="maintenance_mode_description">Server is currently in maintenance mode. Please try again later</string>
 
     <!-- Take photo -->
     <string name="take_photo">Take a photo</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -656,8 +656,8 @@ How to translate with transifex:
     <string name="nc_dialog_outdated_client_description">The app is too old and no longer supported by this server. Please update.</string>
     <string name="nc_dialog_outdated_client_option_update">Update</string>
     <string name="nc_switch_account">Switch account</string>
-    <string name="nc_dialog_maintenance_mode">Maintenance mode</string>
-    <string name="nc_dialog_maintenance_mode_description">Server is currently in maintenance mode.</string>
+    <string name="nc_dialog_maintenance_mode_description">Server is currently in maintenance mode. Please try again
+        later.</string>
```
nickvergessen, line 660: could use a different string name, so it doesn't conflict the translation sync until we backported

mahibi:
sorry, i will close in favor of https://github.com/nextcloud/talk-android/pull/4782
(the snackbar could be overseen and people still wonder whats going on...)

PR #4782, opened 2025-03-06
show maintenance warning in conv list on top

mahibi:
Resolve https://github.com/nextcloud/talk-android/issues/2737 (and replace https://github.com/nextcloud/talk-android/pull/4760)

Since offline support it makes more sense to view this as a permanent warning on top, which still allows to see all offline saved conversations. The actions are still available via account switcher.

before | after
------- | -------- 
grafik | grafik 

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
@@ -1045,6 +1045,10 @@ class ConversationsListActivity :
         binding.chatListConnectionLost.visibility = if (show) View.VISIBLE else View.GONE
     }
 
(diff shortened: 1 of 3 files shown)


nickvergessen:
So this is at the top
"You have no internet connection" in the chat view is at the bottom.

Does it make sense (because its close to the newest item) or should it be consistent

mahibi:
> So this is at the top "You have no internet connection" in the chat view is at the bottom.
> 
> Does it make sense (because its close to the newest item) or should it be consistent

Fine for me to move it up in the chat. Maybe this could wait for JetpackCompose migrations, then we could have it at the top for all screens..

sowjanyakch (approved):

PR #4807, opened 2025-03-17
update workflows

mahibi:
just testing, do not merge

### 🏁 Checklist

- [ ] 📅 Milestone is set


AndyScherzinger (approved):

mahibi:
@AndyScherzinger i'm just testing in this PR and it won't be merged.
(latestly some more things to avoid some zizmor errors, which is the cause why it failed.)

you could approve https://github.com/nextcloud/android-config/pull/247 in android-config instead

Issue #4125, opened 2024-08-29
Bluetooth headset vanishes after selecting


small1:
### Steps to reproduce

1. Start a call
2. select sound source as bluetooth
3. get a calling sound in the headphone
4. sound jumps back to the phone and bluetooth gone as a choice untill you restart a new call

### Expected behaviour

I should keep the call in the headset

### Actual behaviour

It jumps back to the phone

### Device brand and model

Google pixel 7 pro

### Android version

14

### Nextcloud Talk app version

19.0.8

### Talk version

20.0.0 RC1

### Custom Signaling server configured

Yes (specify version in Additional Information)

### Custom TURN server configured

Yes

### Custom STUN server configured

Yes

### Android logs

na

### Server log

```bash
Not applicable
```

### Additional information

Tried with several headsets. Reset phone and bluetooth.

small1:
It works on the same phone with loads of other different meeting/call apps.

mahibi:
i can not reproduce this.

@jancborchardt reported some behavior that may be related:

    > The bluetooth device is not picked by default as an output (unlike for other apps like VLC, Youtube, phone, etc)
    Even when picked, it still seems to go via the phone speaker?
    Then at some point also the entry "Bluetooth" went missing again in the list.
    I force closed the app and tried again multiple times, same result always

I will get a device with android 14 soon and then will test if it's because of the android version. 

When i have the device i will also have a closer look at
https://developer.android.com/about/versions/14/behavior-changes-14#enforce-bluetooth_connect

jancborchardt:
Just for the record, I don’t even get any sound on the bluetooth initially. It is shown in the device selector first, and then vanishes.

mahibi:
got the android 14 smartphone but can't reproduce the issues 😕 So it's maybe not related to android 14.

If you have bluetooth devices, can you try to reproduce @sowjanyakch @rapterjet2004 ?

small1:
I will make a few more tests.  I have a Jabra headset and a Sennheiser and see if it might have been fixed with the latest android patches.

sowjanyakch:
I could not reproduce this issue on my Samsung A52 (Android 14) and pixel 6a (Android 14)device.

PR #4824, opened 2025-03-24
allow bluetooth headset to be discovered and used during a call

gavine99:
bluetooth headset connection and disconnection stopped being recognised whilst in a call. in order to use a bluetooth device for audio one had to have it connected before starting a call.

i believe this functionality was originally working but a change some time ago broke the existing functionality.

this pr fixes so that a bluetooth headset (dis)connection is now recognised and properly handled during a call by adding and removing the bluetooth item on the audio devices list.

possibly closes https://github.com/nextcloud/talk-android/issues/4125

### 🖼️ Screenshots
image

with this pr, bluetooth audio device will appear/disappear as conencted/disconnected during a call.

image

### 🚧 TODO

nothing.

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi (approved):

mahibi:
Thank you @gavine99 :+1: 
Tested and works like expected!

gavine99:
thanks @mahibi.

i would like to work on making the product automagically use a bluetooth device when connected (or when a call is joined if bluetooth device is already connected) - like the default android dialer does.

do you know if this feature is that was intentionally not implemented?

would this require a user setting to toggle 'auto use bluetooth' feature on/off, or should it just do this without a setting?

gavine99:
please see https://github.com/nextcloud/talk-android/pull/4878 for automagically switching to and away from bluetooth audio if user has not actively selected audio output.
if bluetooth device is connected at call start it is used immediately - same as current behavior).

PR #4830, opened 2025-03-31
Test notifications

### LINKED ISSUE #2112: Selftest button for push notifications
In the settings there should be a button which triggers to send a push notification from server which is then expected to be received on mobile.

requires https://github.com/nextcloud/notifications/issues/1206

sowjanyakch:
Resolve #2112 

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
@@ -141,8 +141,8 @@ class DiagnoseActivity : BaseActivity() {
         setupMetaValues()
         setupPhoneValues()
         setupAppValues()
-        testPushNotification()
         setupAccountValues()
+        testPushNotification()
 
         diagnoseDataState.value = diagnoseData.toList()
     }
@@ -197,7 +197,7 @@ class DiagnoseActivity : BaseActivity() {
         ).show()
     }
 
-    private fun testPushNotification()  {
+    private fun testPushNotification() {
         addHeadline(context.resources.getString(R.string.nc_test_push_button))
     }
 
--- a/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
@@ -10,7 +10,6 @@ import android.annotation.SuppressLint
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Intent
-import android.net.Uri
 import android.os.Build
 import android.os.Build.MANUFACTURER
 import android.os.Build.MODEL
--- a/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.colorResource
@@ -88,6 +89,7 @@ class DiagnoseActivity : BaseActivity() {
         )[DiagnoseViewModel::class.java]
 
         val colorScheme = viewThemeUtils.getColorScheme(this)
+        isGooglePlayServicesAvailable = ClosedInterfaceImpl().isGooglePlayServicesAvailable
 
         setContent {
             val backgroundColor = colorResource(id = R.color.bg_default)
@@ -115,6 +117,7 @@ class DiagnoseActivity : BaseActivity() {
                         )
                     },
                     content = {
+                        val viewState = diagnoseViewModel.notificationViewState.collectAsState().value
                         Column(
                             Modifier
                                 .padding(it)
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseActivity.kt
```diff
@@ -187,6 +196,10 @@ class DiagnoseActivity : BaseActivity() {
         ).show()
     }
 
+    private fun testPushNotification() {
+        addHeadline(context.resources.getString(R.string.nc_test_push_button))
```
mahibi, line 200: by adding this via addHeadline the string will be included in diagnoseData, so it will be contained as headline in the copied markdown which doesn't make sense.

Then also the check
`if (element.headline == "Test push notifications") {`
in DiagnoseContentComposable is not necessary. 
Just combine it with ShowNotificationData in DiagnoseContentComposable to initially show the Button.

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -59,19 +115,84 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
...
-@Preview(showBackground = true)
 @Composable
-fun DiagnoseContentPreview() {
-    val state = remember {
-        mutableStateOf(
-            listOf(
-                DiagnoseActivity.DiagnoseElement.DiagnoseHeadline("Headline"),
-                DiagnoseActivity.DiagnoseElement.DiagnoseEntry("Key", "Value")
+@Suppress("LongMethod")
+fun ShowNotificationData(
```
mahibi, line 124: lets extract this to a new file

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -37,13 +65,41 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
...
+                            onClick = {
+                                diagnoseViewModel.fetchTestPushResult()
+                            }
+                        ) {
+                            Text(
+                                text = element.headline,
+                                modifier = Modifier
+                                    .fillMaxWidth()
+                                    .padding(vertical = 8.dp),
+                                color = colorResource(R.color.high_emphasis_text),
```
mahibi, line 83: Text color must not be set here. With R.color.high_emphasis_text the color would not work well together with theming

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -59,19 +115,84 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
...
+                modifier = Modifier
+                    .wrapContentSize()
+                    .padding(16.dp)
+            ) {
+                Column(modifier = Modifier.padding(16.dp)) {
+                    Text(
+                        text = stringResource(R.string.nc_test_results),
+                        style = MaterialTheme.typography
+                            .titleMedium
+                    )
```
mahibi, line 159: I think in the dialog it would make sense to add an explanation to the user that a notification with the text "Testing push notifications" should pop up. 
Otherwise some users might think that "Push notification sent successfully" means that everything is working (which is only true for the server side), even if the notification did not make it to the phone.

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -59,19 +115,84 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
                 }
             }
         }
+        ShowNotificationData(isLoading.value, showDialog.value, context, message.value, diagnoseViewModel)
     }
 }
 
-@Preview(showBackground = true)
```
mahibi, line 65: we should have previews for all Composables

mahibi (changes requested):

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -59,6 +95,124 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
...
+                    )
+                    Box(
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .weight(1f, fill = false)
+                            .verticalScroll(rememberScrollState())
+                    ) {
+                        Column(modifier = Modifier.padding(top = 12.dp)) {
+                            Text(
+                                text = stringResource(R.string.nc_push_notification_message),
```
mahibi, line 202: Do not show the "Push notification is sent successfully..." string when the response was 404

app/src/main/java/com/nextcloud/talk/diagnose/DiagnoseContentComposable.kt
```diff
@@ -59,6 +95,124 @@ fun DiagnoseContentComposable(data: State<List<DiagnoseActivity.DiagnoseElement>
                 }
             }
         }
+        ShowTestPushButton(onTestPushClick)
```
mahibi, line 98: Do not show the "Test push notifications" button when gplay services are not available
You can see the check for this in the diagnosis screen itself.
(So if users are asked to click the self test push and don't find it in diagnosis screen, we already know the problem is on device...)

If notification app is not installed it's fine for me to show the button. In this case the testresult will be 404 which is okay as long there is no "success" string shown (see other comment).

mahibi (changes requested):
see comments

mahibi (approved):

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +230,11 @@ How to translate with transifex:
     <string name="send_email">Send email</string>
     <string name="create_issue">Create issue</string>
     <string name="nc_diagnose_flavor" translatable="false">Build flavor</string>
+    <string name="nc_test_push_button">"Test push notifications</string>
```
nickvergessen, line 233: Leading quote

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +230,11 @@ How to translate with transifex:
...
+    <string name="nc_test_results">Test results</string>
+    <string name="message_copied">Message copied</string>
+    <string name="nc_push_notification_message">Push notification is sent successfully. You should now receive
```
nickvergessen, line 236: Line break inside the string is not supported

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +230,11 @@ How to translate with transifex:
...
+        a notification on this device with the title \'Testing push notifications\'  </string>
```
nickvergessen, line 237: the trailing space on translation string, please read the do's and don'ts
https://docs.nextcloud.com/server/latest/developer_manual/basics/translations.html#dos-and-don-ts

PR #4876, opened 2025-04-15
WIP targetSdkVersion 35

mahibi:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] do not use `fitsSystemWindows=true`
- [ ] design decisions that are caused by edge-to-edge support
- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -28,13 +28,13 @@ apply plugin: "org.jlleitschuh.gradle.ktlint"
 apply plugin: 'kotlinx-serialization'
 
 android {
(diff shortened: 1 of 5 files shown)


sowjanyakch (changes requested):
Tested with Android 15, pixel 6a - works well. 
Tested with Android 14, Samsung A52 - search for contacts is broken, it does not show "X" icon to delete search text.

contacts_search

PR #4878, opened 2025-04-16
automatic switch to and away from bluetooth audio if no audio actively selected by user

gavine99:
fix https://github.com/nextcloud/talk-android/issues/4920

add a new 'default' audio device option in audio manager (not in ui). this stops auto-switch to bluetooth being disabled by 'user selected audio device' being set in code and not by user.

if the user actively selects an audio option a new bluetooth connection will not automatically switch away from that selection.

### 🚧 TODO

nothing

### 🏁 Checklist

no new tests required.
backporting not required.

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -836,7 +836,7 @@ class CallActivity : CallBaseActivity() {
 
     fun setDefaultAudioOutputChannel(selectedAudioDevice: AudioDevice?) {
         if (audioManager != null) {
-            audioManager!!.selectDefaultAudioDevice(selectedAudioDevice)
+            audioManager!!.setDefaultAudioDevice(selectedAudioDevice)
             updateAudioOutputButton(audioManager!!.currentAudioDevice)
         }
     }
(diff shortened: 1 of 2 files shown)


mahibi:
Thank you @gavine99 !

Using the bluetooth connection when entering the call worked fine for me

However when being in a call and then connect the bluetooth device, it was not successful:
- First the icon switches to bluetooth (but voice can not be heard)
- Then after a few seconds, the device switches back to default microphone.
- This seems to happen because of the "cancelTimer". In 
`BluetoothHeadsetBroadcastReceiver#onReceive`
`state` is `BluetoothHeadset.STATE_DISCONNECTED`



I won't find time to have a closer look today and next week, but @sowjanyakch can review/ test/ help.

gavine99:
@mahibi and @sowjanyakch  i've added a commit that delays attempting to switch to bt audio until after a sco connection is made which should stop what @mahibi saw with a, seemingly, failed sco connection. the sco connection can still fail, but in that case the audio won't have switched and should still be using the pre-bluetooth source and sink. same for the time between bt selection and sco connection being created - pre-bt-selection source and sink should be still used during that delay.

sowjanyakch:
Thank you @gavine99 !  Tested this PR and works as expected

sowjanyakch (approved):

mahibi:
Thanks again @gavine99 

i don't have the other bluetooth device available where it didn't work 2 weeks ago (so i'm not 100% sure if it's solved by your changes or if there is a problem with the bluetooth device). But for another bluetooth device it works absolutely fine, so i would like to merge

PR #4889, opened 2025-04-21
Follow up fixes for Improving Chat Message Search

### LINKED ISSUE #3527: Message from search is not shown in chat
If search returns a message that is not yet loaded by adapter, it just won't show the message (and no info that it failed).

```
    private fun scrollToMessageWithId(messageId: String) {
        val position = adapter?.items?.indexOfFirst {
            it.item is ChatMessage && (it.item as ChatMessage).id == messageId
        }
        if (position != null && position >= 0) {
            binding.messagesListView.smoothScrollToPosition(position)
        } else {
            // TODO show error that we don't have that message?
        }
    }
```

https://github.com/nextcloud/talk-android/blob/59e8f4b550501b4ab833c44fee826829b8fa62dd/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt#L2587C64-L2587C64

This was never implemented properly, so it exists since the search feature was added.
It maybe makes sense to even wait with this until offline support is implemented.

## reproduce
search for quite old message

## expected
scrolled to message

## actual
chat opens with latest messages. searched message is not shown.

### LINKED ISSUE #4885: Followup improvements for search for chat messages
Some ideas what should be improved ad followup to https://github.com/nextcloud/talk-android/pull/4647:

- [ ] Maybe we had a misunderstanding in https://github.com/nextcloud/talk-android/pull/4647#issuecomment-2734020059:

> search inside a chat (upper right menu) is missing (and then only search inside this chat)

I am talking about if you go into a chat, there is the search available in the upper right corner.
When searching there (instead in the main screen), the search you implemented should also be available, but limited to the chosen chat.

This `else` branch will be the right place to open the search window:
https://github.com/nextcloud/talk-android/blob/a9090d4e71d02a165294cb95ff7d3dcafdc5b426/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt#L2131-L2142

So when the message is found in the chat adapter, scroll to message. If it's not found, open the search result window..

- [ ] Adding the disabled call buttons and the disabled menu in the search results window doesn't make sense to me (i guess this is what you understood from my description?)

- [ ] Rotating the device leads to some wrong design & search window is closed.

- [ ] Some file icons are not yet displayed as expected. For example on sermo in the "Filetype examples" (e.g. icons for zip files..)

- [ ] Extract Composables out of ComposeChatAdapter and add Compose Previews for each message type

### LINKED ISSUE #4915: reply to old message - click reply - message not found
### Steps to reproduce

1. Someone replies to an old message
2. you tap on the citation of the old message
3. Error "Message was not found"

### Expected behaviour

scroll to cited message

### Actual behaviour

Error, see above

### Device brand and model

google pixel 4a

### Android version

13 (graphene OS)

21.0.1

### Nextcloud server version

c.nc.com

### Talk version

_No response_

None

None

None

### Android logs

_No response_

### Server log

```bash

```

### Additional information

Let me know if you need more info. I can reproduce.

rapterjet2004:
- fixes #4885 
- fixes #4915 
- fixes #3527

---
- [x] Failure to find message id in db creates the context chat compose
- [x] Rotating the device leads to some wrong design & search window is closed.
- [x] Some file icons are not yet displayed as expected. For example on sermo in the "Filetype examples" (e.g. icons for zip files..)
- [x] Extract and add Compose Previews for each message type (will require a lot of a refactoring)
- [x] Test it out

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/components/StandardAppBar.kt
+++ b/app/src/main/java/com/nextcloud/talk/components/StandardAppBar.kt
@@ -1,7 +1,7 @@
 /*
  * Nextcloud Talk - Android Client
  *
- * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
+ * SPDX-FileCopyrightText: 2025 Marcel Hibbe <dev@mhibbe.de>
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
--- a/app/src/main/java/com/nextcloud/talk/contacts/components/SearchComponent.kt
+++ b/app/src/main/java/com/nextcloud/talk/contacts/components/SearchComponent.kt
@@ -1,7 +1,7 @@
 /*
  * Nextcloud Talk - Android Client
  *
- * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
+ * SPDX-FileCopyrightText: 2025 Julius Linus <juliuslinus1@gmail.com>
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
--- a/app/src/main/java/com/nextcloud/talk/utils/preview/ComposePreviewUtils.kt
+++ b/app/src/main/java/com/nextcloud/talk/utils/preview/ComposePreviewUtils.kt
@@ -1,7 +1,7 @@
 /*
  * Nextcloud Talk - Android Client
  *
- * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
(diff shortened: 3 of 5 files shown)


mahibi:
@rapterjet2004 what the sate of this PR? Asking as it's marked as "developing" and no reviewer is assigned.
Is it still work in progress or should it be reviewed?

mahibi (approved):

PR #4898, opened 2025-04-22
"Add" button to add participants

### LINKED ISSUE #4886: Can not add users to conversation after searching
1. Have an instance with many users, notice that the colleague you want to add is not listed.
2. Search for them in an existing conversation 
3. Tick them
4. There is no button to continue/add them

sowjanyakch:
Resolve #4886

Also resolve https://github.com/nextcloud/talk-android/pull/4898#issuecomment-2910369602

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/contacts/components/SearchComponent.kt
+++ b/app/src/main/java/com/nextcloud/talk/contacts/components/SearchComponent.kt
@@ -37,7 +37,7 @@ fun SearchComponent(
     text: String,
     onTextChange: (String) -> Unit,
     onDisableSearch: () -> Unit,
-    modifier:Modifier = Modifier
+    modifier: Modifier = Modifier
 ) {
     val keyboardController = LocalSoftwareKeyboardController.current
     TextField(
@@ -103,4 +103,3 @@ fun searchKeyboardActions(text: String, keyboardController: SoftwareKeyboardCont
             }
         }
     )
-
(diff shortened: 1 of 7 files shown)


mahibi (changes requested):
_Note to myself as i need some minutes to get it... this is about the "Add participants" button in the conversation info screen._

- [ ] When using the search now with isAddParticipants=false, the "X" button is not aligned to the right. There is a blank space where the "Add" button is in the other scenario. 

- [ ] Add button should only be enabled when at least one user is selected (otherwise when clicking add, the "Add" will be replaced by "Done", which is confusing).
If no user is selected, the button should be visible but disabled.

- [ ] When users were selected and "Add" was clicked, the filtered search results should be reset. I expect that all searchable users are listed again and the one who were selected are marked. Right now, there is no possibility to return to this state.

app/src/main/java/com/nextcloud/talk/contacts/components/SearchComponent.kt
```diff
@@ -101,3 +109,10 @@ fun searchKeyboardActions(text: String, keyboardController: SoftwareKeyboardCont
             }
         }
     )
+
+class SearchComponentCompanionClass {
+    companion object {
+        const val WIDTH_RATI0_SMALL = 0.85f
```
mahibi, line 115: This width is aligned to the length of the text string "Add", which won't work for other languages.
Changed the translation to "Addddddddddddd1" and it looks like this:

grafik

mahibi (changes requested):

PR #4899, opened 2025-04-23
Fix - Message Long press menu flickering

### LINKED ISSUE #4883: Message long press menu is "jumping"
1. Long press on a message being ready to copy it
2. Try to click the action but the menu jumps as a single entry is rendered later?

I think it's the "note to self" that takes some time?

Especially annoying, when you then switch to another conversation and long press on another message it seems to check it again and flickers again.
The fact that this option exists should be independent from any other information, so why is it "jumping in" later?

sowjanyakch:
- Resolve #4883

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
@@ -60,8 +60,7 @@ class MessageActionsDialog(
     private val currentConversation: ConversationModel?,
     private val showMessageDeletionButton: Boolean,
     private val hasChatPermission: Boolean,
-    private val spreedCapabilities: SpreedCapability,
-    private val noteToSelfRoomToken: String
+    private val spreedCapabilities: SpreedCapability
 ) : BottomSheetDialog(chatActivity) {
 
     @Inject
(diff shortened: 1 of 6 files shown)


nickvergessen (changes requested):

app/src/main/java/com/nextcloud/talk/ui/dialog/MessageActionsDialog.kt
```diff
@@ -472,13 +474,15 @@ class MessageActionsDialog(
         dialogMessageActionsBinding.menuSaveMessage.visibility = getVisibility(visible)
     }
 
-    private fun initMenuAddToNote(visible: Boolean, roomToken: String = "") {
+    private fun initMenuAddToNote(visible: Boolean) {
         if (visible) {
+            chatViewModel.checkForNoteToSelf(user!!)
```
nickvergessen, line 479: This does a request on each menu open?

sowjanyakch: This makes a request when "Add To Notes" option is visible in the menu.

nickvergessen: Yes, and as said yesterday the request should only happen when you use the action, as it otherwise creates the empty "Note to self" every time you long press any message, which is not okay.

sowjanyakch:
need to make an api call only when Add To Options is clicked. WIP.

nickvergessen:
Replaced by #4935

PR #4935, opened 2025-04-29
Bugfix/4883/fix jumping note to self entry

### LINKED ISSUE #4883: Message long press menu is "jumping"

mahibi:
fix #4883

### 🏁 Checklist

- [ ] 📅 Milestone is set


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -3295,37 +3295,61 @@ class ChatActivity :
...
+            intArrayOf(ApiUtils.API_V4, ApiUtils.API_V3, 1)
+        )
+        chatViewModel.checkForNoteToSelf(
+            ApiUtils.getCredentials(conversationUser!!.username, conversationUser!!.token)!!,
+            ApiUtils.getUrlForRooms(
+                apiVersion,
+                conversationUser!!.baseUrl
+            ),
+            false
+        )
```
nickvergessen, line 3310: As long as this is only done when the action is pressed, instead of getting all rooms and searching for the note-to-self, can you GET the room with the token `note-to-self`? That will create it in case it was (accidentally) deleted

mahibi: improved with ab19f631aeb4264e894da4075ce524131f727723

sowjanyakch (approved):

PR #4937, opened 2025-04-29
fix crash

### LINKED ISSUE #4904: App crashes on screen rotation of user conversation info
1. Navigate to conversation info screen.
2. Rotate the screen and the app crashes.

sowjanyakch:
Resolve #4904 

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi (approved):
it fixes the crash so will merge.

As a side note: It feels wrong that handleConversation is called 2 times after rotation (one time with conversation=null, next time with an actual conversation). Feels like there should be architecture improvements..

Issue #4916, opened 2025-04-25
Video streams are not restored after switching to PIP and back


Antreesy:
### Steps to reproduce

1. Install latest app version
2. Join a group call with cameras on
3. enable Picture-In-Picture
4. See PIP overlay doesn't show any video peer, only a thumbnail
5. disable PIP
6. See other participants video are no longer visible, as if they have cameras disabled

### Expected behaviour

Video streams should be restored

### Actual behaviour

Before | After
-- | --
![Image](https://github.com/user-attachments/assets/004cb190-832a-4ac9-ad76-3e26e230b973) | ![Image](https://github.com/user-attachments/assets/36142261-1f4c-454d-9582-f33b2720bb07)

### Device brand and model

Xiaomi (Redmi Note 13)

### Android version

15

### Nextcloud Talk app version

21.1.0 Alpha 15

### Nextcloud server version

31

### Talk version

21

### Custom Signaling server configured

Yes (specify version in Additional Information)

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

PR #4939, opened 2025-04-29
Improve pip mode

mahibi:
show own video in PIP mode for group calls

fix "See PIP overlay doesn't show any video peer, only a thumbnail" from https://github.com/nextcloud/talk-android/issues/4916

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
Tested this PR, it shows own video in PIP mode for group call. Tap on the video, I see blur video of one of the participants.

PR #4947, opened 2025-05-07
Improve call grid & Picture-inPicture view

### LINKED ISSUE #4916: Video streams are not restored after switching to PIP and back

mahibi:
This PR migrates the call grid to Compose. (This will avoid handling with wrong video renderer handling in the XML GridView, so it will fix #4916 )

Additionally the Design is made similar to the other platforms (Same background colors for cells...).

This PR could solve other issues related to calls as well, but architecture needs to improve further.

### How to test
- Please test different call scenarios like different amount of participants, switch to PIP mode and back, test portrait and landscape mode, voice only call, video call, ...

### 🖼️ Screenshots

Android before | Android after
------- | -------- 
grafik | grafik

to compare...

Web | iOS 
------- | -------- 
grafik | grafik

Arrangement of cells with different amount of participants and portrait/landscape mode. Scolling is enabled when there are too many participants to match on the screen

grafik

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
I tested this PR - PIP mode works as expected. Tested audio and video calls both in portrait and landspace mode. 
For audio call, the call grid is a scrollable view for both portrait and landscape mode.  
scrollable portrait mode

sowjanyakch (changes requested):
https://github.com/nextcloud/talk-android/pull/4947#issuecomment-2877384806

mahibi:
> I tested this PR - PIP mode works as expected. Tested audio and video calls both in portrait and landspace mode. For audio call, the call grid is a scrollable view for both portrait and landscape mode. 

I guess it is caused by the workaround that i used for now (either the condition fails or the hardcoded dp's mess it up):

```
    val heightForNonGridComponents = if (isVoiceOnlyCall && !isInPipMode) {
        // this is a workaround for now. It should ~summarize the height of callInfosLinearLayout and callControls
        // Once everything is migrated to jetpack, this workaround should be obsolete or solved in a better way
        240.dp
    } else {
        0.dp
    }

    val gridHeight = LocalConfiguration.current.screenHeightDp.dp - heightForNonGridComponents
```

As the calls itself work, i suggest to merge this for now and a fix will follow later today (to be released with RC1 today) or for RC2. I will have to review and merge other PRs that should make it into RC1..

sowjanyakch (approved):

PR #4980, opened 2025-05-22
Send silentFor array parameter

### LINKED ISSUE #4960: Send silentFor array parameter when switch-to an extended_conversation
- To resolve https://github.com/nextcloud/spreed/issues/14983 clients need to tell the backend to send call notifications and ignore the previous participant
- API in progress in https://github.com/nextcloud/spreed/pull/15083
- Starting point should be https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/signaling/SignalingMessageReceiver.java#L295
- API call where you need to send the silentFor is https://github.com/nextcloud/talk-android/blob/1e1a8f7f21c0b3d2f74a4278bb3f09703b0e0533/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt#L3024

The silentFor array can be set to the conversation.name of the previous conversation

rapterjet2004:
- fixes #4960 

### 🚧 TODO
- [ ] double check implementation
- [ ] set up environment and test it out

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
I don't think implementing this makes sense right now, see my comment in
see my comment in https://github.com/nextcloud/talk-android/issues/4960#issuecomment-2903530570

PR #5003, opened 2025-05-27
change some logic from event conversations PRs

mahibi:
.. while making sure logic from near/far future events remains and archived conversation handling is fixed (must not show up in main conversations list).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
@sowjanyakch please check if i oversaw something that was solved with your PRs but not with this one

sowjanyakch (changes requested):
Event conversations that take place after 16 hours are excluded from filter. If you create an event conversation that occurs after 16 hours, post a message, and mark it as unread, it won’t show up under the “Unread” filter. This is the same case for unread mentions and archived filters.

PR #5008, opened 2025-05-28
simplify call grid design logic for voice only calls

mahibi:
followup to #4947

This will remove the workaround regarding heightForNonGridComponents by using a BoxWithConstraints

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
with a bit of luck, it may also fix https://github.com/nextcloud/talk-android/pull/4947#issuecomment-2877384806 
Can you check this @sowjanyakch ? (i was never able to reproduce)

PR #5049, opened 2025-06-09
Preserve Conversation List position

### LINKED ISSUE #5041: Jump to the top after press back button
### Steps to reproduce

If you click on a contact and go back with the left "back button" you jump always to the top of your contact list.

### Expected behaviour

I would like to jump to the position were the (last) contact was. (in the contact list)

### Actual behaviour

You jump always to the top of your contact list.

### Device brand and model

Pixel 7 Pro 

### Android version

15

21.1.0

### Nextcloud server version

_No response_

### Talk version

_No response_

None

None

None

### Android logs

_No response_

### Server log

```bash

```

### Additional information

_No response_

rapterjet2004:
- fixes #5041

Conversation List jumps to position of clicked conversation item when back pressed

### ToDO
- [x] get/store position in shared preferences, not sure why I sent this by intent. Maybe I need some coffee.
- [x] Reevaluate scroll logic.

> I couldn't find a reliable way to save/restore the vertical offset from the `calculateVerticalOffset()` method. The units don't translate reliably to the ones needed for `scrollY` and the method to `scrollVerticallyBy()` requires RecyclerView.State which is internal. Restoring the recycler view state from `layoutManager.onSaveInstanceState()` also doesn't work.

Current Implementation is quite simple though, requires keeping a reference to the item's viewHolder in `ConversationItem`. It's then gets the position of the first visible item and calculates the non visible offset to scroll by. The result is the exact position the user left off. Restored/Retrieved from `AppPreferences`.

### 🖼️ Screenshots

[Screen_recording_20250609_124059.webm](https://github.com/user-attachments/assets/5cd02816-a2bd-408c-aae2-2875525b2c95)

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
@@ -345,6 +345,16 @@ class ConversationsListActivity :
         showSearchOrToolbar()
     }
 
+    override fun onPause() {
+        super.onPause()
+        val firstVisible = layoutManager?.findFirstVisibleItemPosition() ?: 0
+        val firstItem = adapter?.getItem(firstVisible)
+        val firstTop = (firstItem as ConversationItem).mHolder?.itemView?.top
+        val firstOffset = firstTop?.minus(CONVERSATION_ITEM_HEIGHT) ?: 0
+
+        appPreferences.setConversationListPositionAndOffset(firstVisible, firstOffset)
+    }
+
     // if edge to edge is used, add an empty item at the bottom of the list
     @Suppress("MagicNumber")
     private fun addEmptyItemForEdgeToEdgeIfNecessary() {
@@ -1881,12 +1891,6 @@ class ConversationsListActivity :
             bundle.putString(BundleKeys.KEY_MESSAGE_ID, selectedMessageId)
             selectedMessageId = null
         }
-        val firstVisible = layoutManager?.findFirstVisibleItemPosition() ?: 0
-        val firstItem = adapter?.getItem(firstVisible)
-        val firstTop = (firstItem as ConversationItem).mHolder?.itemView?.top
-        val firstOffset = firstTop?.minus(CONVERSATION_ITEM_HEIGHT) ?: 0
-
-        appPreferences.setConversationListPositionAndOffset(firstVisible, firstOffset)
 
         val intent = Intent(context, ChatActivity::class.java)
         intent.putExtras(bundle)


mahibi (changes requested):
This will always place the last opened conversation at the top when coming back. This is not expected but people expect the exact scroll position as when the chat was opened.

It should be solved with
`scrollToPositionWithOffset`
which gets the scroll position and the offset from when the chat was opened.

mahibi (changes requested):
I have the feeling this can be solved in a simpler way.

Right now the position is passed to ChatActivity and back to ConversationList. This should not be necessary at all but ConversationList should handle this itself.
For example this does not work when the app just goes to background and coming back.

Also, i guess that findFirstVisibleItemPosition is something that should not be necessary. It will "just" scroll to an item instead to the exact position.

Can you have a look if there are simpler ways to achieve this?
Like just using the scroll position on the recyclerview itself and just using lifecycle methods to save/restore the scroll postion, eg. from the viewmodel..?

mahibi:
@rapterjet2004 whats the state for this PR? should it be "to review"?

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -345,6 +345,16 @@ class ConversationsListActivity :
...
     }
 
+    override fun onPause() {
+        super.onPause()
+        val firstVisible = layoutManager?.findFirstVisibleItemPosition() ?: 0
+        val firstItem = adapter?.getItem(firstVisible)
+        val firstTop = (firstItem as ConversationItem).mHolder?.itemView?.top
+        val firstOffset = firstTop?.minus(CONVERSATION_ITEM_HEIGHT) ?: 0
+
+        appPreferences.setConversationListPositionAndOffset(firstVisible, firstOffset)
```
mahibi, line 355: @rapterjet2004 i moved this code to onPause, otherwise it would just work when coming back from a chat and not when the app was just in background.
Guess that's okay for you or was there a reason against onPause?

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -2221,5 +2233,6 @@ class ConversationsListActivity :
         private const val SIXTEEN_HOURS_IN_SECONDS: Long = 57600
         const val LONG_1000: Long = 1000
         private const val NOTE_TO_SELF_SHORTCUT_ID = "NOTE_TO_SELF_SHORTCUT_ID"
+        private const val CONVERSATION_ITEM_HEIGHT = 44
```
mahibi, line 2236: not the nicest solution but fine for me as the handling will be replaced with Compose anyway

PR #5065, opened 2025-06-17
Try to run analysis also for merge commits

mahibi:
Trying to fix
https://github.com/nextcloud/android-config/pull/248/files

which caused that analysis is not only blocked for forks, bit also for merge commits.

Solution: If it's not a PR the step should be omitted..

This commit will be checked when workflow runs on github. It will be merged just to test it's working..

- [ ] If it works, do the change in android-config repo

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -29,7 +29,7 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Disabled on forks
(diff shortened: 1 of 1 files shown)


mahibi:
analysis is executed for the merge commit, but results could not be pushed, see
`fatal: could not read Username for 'https://github.com/': No such device or address`
in
https://github.com/nextcloud/talk-android/actions/runs/15704373286/job/44247921124

PR #5067, opened 2025-06-18
Message threads

### LINKED ISSUE #3074: Message threads
* Overview at https://github.com/nextcloud/spreed/issues/1469
* Rough idea is that when a reply message or a message that has a reply is clicked, we filter the chat stream with that "thread" ID (root parent message).
* All newly written messages should be addresses to that root parent as well (needs a new API parameter, so that you can still reply within a thread for quoting and reactions)
* The UI of course needs to see a button to "go back" to the unfiltered list.

mahibi:
fix #3074 

This PR is not complete and contains bugs but it will be merged to master for early testing in alpha channel.
There will be followup PRs for threads support.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [x] adapt chatBlocks handling for threads
- [x] clear DB tables when migrating
- [x] add option to create a thread
- [x] add option to open a thread
- [x] threads overview
- [x] thread title incl. count of replies
- [x] add logic to only show "Reply in thread" OR "Open thread" in message context menu
- [x] show number of replies for each thread in overview
- [x] open context menu also when longclick on quoted message
- [x] scrolling in threads overview ok?
- [x] reply handling? normal vs thread..
   - [x] write a message in thread = reply to thread 
- [x] bug when posting in thread?
- [x] backpress logic
- [x] better error handling
- [ ] ~hide thread messages in normal chat~ we won't do this for the initial version
- [ ] ~replies button. Show number of replies and open thread~

## TODOS for next PRs
- [ ] fix bug that in some threads messages don't load
- [ ] show only first name of participants in threads overview?
- [ ] display mentions for messages in thread overview 
- [ ] hide "created a thread" system message in thread itself
- [ ] "subscribe" handling?
   - [ ] icon in appbar (instead call icons etc)
   - [ ] implement endpoint
- [ ] unread marker handling?
   - [ ] do not set read marker in thread
- [ ] mention handling?
- [ ] what if first message of thread is a file etc? For now "{file}" would be shown
   - [ ] show caption or show filename + icon
- [ ] show icon for thread messages in normal chat?
- [ ] check dark theme
- [ ] hide threads options when not supported by server. capability?
- [ ] share files in threads
   - [ ] needs also improvements unrelated to threads beforehand: must be possible to reply with message types other than text
- [ ] uncomment `.openHelperFactory(factory)`
- [ ] revert to show messageIds in debug mode
- [ ] offline support for threads overview
- [ ] further testing, esp of chatBlocks

### 🏁 Checklist

- [ ] 📅 Milestone is set


nickvergessen:
Has some conflicts, but otherwise this can be merged to get it into the testing alpha versions

PR #5078, opened 2025-06-24
show meaningful message when offline messages are not saved

### LINKED ISSUE #5038: Remove/improve modal when server is down
1. Have two talk account configured
2. Select one account
3. Turn off the server

---

Expectation:
- I can browser the list of conversations
- I can offline read chat
- I can change to another profile
- I can change back to the profile while the server is still down

sowjanyakch:
Resolve #5038 

While functionality is already implemented - just a UI modification. When no offline messages are present whatever case server is down/maintenance mode show a meaningful warning.

Show server is in maintenance mode warning in chat activity.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
@@ -162,7 +162,7 @@ class OfflineFirstChatRepository @Inject constructor(
                 if (!weAlreadyHaveSomeOfflineMessages) {
                     Log.d(TAG, "An online request for newest 100 messages is made because offline chat is empty")
                     val maintenanceMode = network.getServerStatus(currentUser.baseUrl!!).maintenance
-                    if (networkMonitor.isOnline.value.not() || maintenanceMode){
+                    if (networkMonitor.isOnline.value.not() || maintenanceMode) {
                         _generalUIFlow.emit(ChatActivity.NO_OFFLINE_MESSAGES_FOUND)
                     }
                 } else {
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -161,9 +161,7 @@ class OfflineFirstChatRepository @Inject constructor(
             } else {
                 if (!weAlreadyHaveSomeOfflineMessages) {
                     Log.d(TAG, "An online request for newest 100 messages is made because offline chat is empty")
-                    if (networkMonitor.isOnline.value.not()) {
```
mahibi, line 164: This check is to show the "No offline messages saved" info in a chat when the smartphone has no internet connection. It's not related to when the server is down. Removing it does not solve #5038

#5038 is about the modal in the Conversation list

app/src/main/res/values/strings.xml
```diff
@@ -667,6 +667,7 @@ How to translate with transifex:
     <string name="nc_dialog_outdated_client_option_update">Update</string>
     <string name="nc_switch_account">Switch account</string>
     <string name="nc_dialog_maintenance_mode_description">Server is currently in maintenance mode.</string>
```
mahibi, line 669: nc_dialog_maintenance_mode_description is used in the conversation list. We should keep it the same for all screens.

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -161,7 +161,13 @@ class OfflineFirstChatRepository @Inject constructor(
...
+                    val offline = networkMonitor.isOnline.value.not()
+                    val maintenance = if (!offline) {
+                        network.getServerStatus(currentUser.baseUrl!!).maintenance
```
mahibi, line 166: This will trigger a request every time which is not necessary.
Also, we should not begin to introduce logic for this for every screen separately.  
If you have a look at how it's solved in ConversationList screen you see that this can be handled via the http response errors or headers. Also see https://docs.nextcloud.com/server/latest/developer_manual/client_apis/general.html#maintenance-mode

A clean solution would be to introduce an overall handling of the http responses so all screens could use it. That's out of scope here and should be done in a different PR.
In any way, this is not related to https://github.com/nextcloud/talk-android/issues/5038

PR #5085, opened 2025-06-25
fix duplicate messages

### LINKED ISSUE #5068: Duplicate messages in missing microphone permissions scenario
STR:
1. Don't grant "Talk" Microphone permissions.
2. Create a new conversation and send text message.
3. Send a voice message, you will see duplicate messages.
4. Click on attachment icon and select "Take photo", now camera opens - try to navigate back, you will see duplicate messages.

duplicate messages occur when uploading messages from gallery, also sharing a contact without permissions

It's just UI, so nothing is duplicated persistently..

Seen on v21.1.0

sowjanyakch:
Resolve #5068

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -1404,9 +1404,7 @@ class ChatActivity :
 
         cancelNotificationsForCurrentConversation()
 
-
         chatViewModel.getRoom(roomToken)
-
 
         actionBar?.show()
 
@@ -2665,6 +2663,7 @@ class ChatActivity :
         if (mentionAutocomplete != null && mentionAutocomplete!!.isPopupShowing) {
             mentionAutocomplete?.dismissPopup()
         }
+        adapter = null
     }
 
     private fun isActivityNotChangingConfigurations(): Boolean = !isChangingConfigurations


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -1334,7 +1334,9 @@ class ChatActivity :
 
         cancelNotificationsForCurrentConversation()
 
-        chatViewModel.getRoom(roomToken)
+        if (adapter == null) {
```
mahibi: This feels like it might introduce a race condition..

In onCreate, initObservers() is called which in turn initializes the adapter in
```
            chatViewModel.getConversationFlow
                .onEach ....
```

At the same time, onResume is executed with the current check. This might be a race condition as we can not be sure if adapter is initialized or not. So getRoom may not be triggered when it should?

mahibi (changes requested):

sowjanyakch:
When the user opens the camera to take a picture or share a contact, the app goes into onPause(). The adapter is set to null.

When the user comes back to the app (onResume()), the messages are reloaded again preventing duplicate messages.

rapterjet2004 (approved):

PR #5099, opened 2025-06-30
Show delete all messages in 1:1 only when canDeleteConversation is true

sowjanyakch:
- show delete all messages in one to one conversation only when canDeleteConversation is true

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
~~For a followup PR:
please also add "Delete all messages" to the context menu of a conversation in conversation list (if setting is enabled). There should be the same entries as in ConversationInfo screen, with the same order.~~

update: @sowjanyakch  i checked on sever and there it's also not in the context menu, so just forget about about it

PR #5108, opened 2025-07-01
🎨 Migrate to outlined icon style

AndyScherzinger:
### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
QA build for testing see above

mahibi:
grafik

In `ConversationCreationActivity` in line 292 lets change  
`painter = painterResource(id = R.drawable.ic_mimetype_folder),`
to
`painter = painterResource(id = R.drawable.ic_folder),`


.. regardless of this:
the mimetype icons should not be outlined, right?

AndyScherzinger:
Nice catch @mahibi 

Yes for a non-mimetype folder icon we should go for the outline variant, for the "normal" file-listings the icons shall be filled.
Let me fix that, I already caught it for app/src/main/res/layout/activity_profile.xml but seem to have missed the one you posted. Great catch 👍

AndyScherzinger:
Fixed  @mahibi - also fixed the landscape profile layout one, since that one also still had the filled folder instead of the outline variant 👍

mahibi (approved):

PR #5145, opened 2025-07-10
Login via QR Code

### LINKED ISSUE #1071: Enter Login URL via QR code
enter Login URL via QR code like android files app already does

rapterjet2004:
- fixes #1071 

Uses this dependency `com.github.blikoon:QRCodeScanner:0.1.2` to align with files, which is pretty old.

### 🚧 TODO

- [x] Got the logic down
- [x] Test it

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi (changes requested):
works good :+1: 

Some things that should be improved:

- The contents of the QR code scanner are overlapped by the system bar (On my device, "Scan" is also overlapped by the  notch). It's for sure because the dependency is old. From my side this is not a blocker for merge because it's the same situation on android files.

- Choose "Picture" feature in the upper right corner is not available for files app. We should also not show it.

grafik

- When going back without to scan a QR code, there should not be the "Please continue the login process in the browser" screen but the screen to enter a URL or choose QR icon.

- Please align the design with the files app. They have a bigger icon and no text.

mahibi:
:+1: 

@rapterjet2004 can do these in a followup PR?:
- [ ] use the same icon and size as android files
- [ ] on android files, the content of the page is aligned when the keyboard opens (so the QR code icon is not overlapped by the keyboard). Would be nice for the talk app as well.

PR #5172, opened 2025-07-21
Bugfix/5038/improve modal when server down

### LINKED ISSUE #5038: Remove/improve modal when server is down

sowjanyakch:
Resolve #5038 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
@rapterjet2004 - Please test the PR with multiple server accounts and different user profiles.

app/src/main/java/com/nextcloud/talk/data/network/NetworkMonitor.kt
```diff
@@ -24,4 +24,8 @@ interface NetworkMonitor {
      * Returns the device's current connectivity status as LiveData for better interop with Java code.
      */
     val isOnlineLiveData: LiveData<Boolean>
+
+    val isServerReachable: StateFlow<Boolean>
+
+    fun setServerReachable(reachable: Boolean)
```
rapterjet2004: the `isServerReachable` flow and `setServerReachable` function should not be included in the NetworkMonitor, as NetworkMonitor exists as an immutable source of truth regarding network connection. Also I think the naming is a misrepresentation, if the server is online, then it is also reachable. This code really just checks if an error occurred.

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -166,6 +166,9 @@ class OfflineFirstChatRepository @Inject constructor(
                     if (networkMonitor.isOnline.value.not()) {
                         _generalUIFlow.emit(ChatActivity.NO_OFFLINE_MESSAGES_FOUND)
                     }
+                    if (!networkMonitor.isServerReachable.value) {
```
rapterjet2004: Is there a way you can implement a check here in the repository? One that detects from the source of truth (network/local storage) that the server is online, but not loading chats?

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -1213,6 +1215,7 @@ class ConversationsListActivity :
                         openConversationItems.add(conversationItem)
                     }
                     searchableConversationItems.addAll(openConversationItems)
+                    networkMonitor.setServerReachable(true)
```
rapterjet2004: Setting/retrieving the `isServerReachable.value `in the view layer is [bad practice](https://developer.android.com/topic/architecture/ui-layer), as the view is a visual representation of the application state as retrieved from the data layer.  The philosophy behind this is [separation of concerns](https://developer.android.com/topic/architecture#separation-of-concerns), as a view that is isolated is easier to test, and easier to debug.

For example, on lifecycle changes, what happens to the state of network monitor? It should be lifecycle agnostic, not caring about the implementation details of the view or the view's current state.

app/src/main/java/com/nextcloud/talk/serverstatus/ServerStatusRepositoryImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+    private var _isServerReachable: MutableStateFlow<Boolean> = MutableStateFlow(false)
+    override val isServerReachable: StateFlow<Boolean>
+        get() = _isServerReachable.asStateFlow()
+
+    override suspend fun getServerStatus() {
+        serverStatus()
+    }
+
+    @Suppress("Detekt.TooGenericExceptionCaught")
+    private suspend fun serverStatus() {
```
rapterjet2004: `serverStatus` should be renamed `getServerStatus` should be using `withContext(Dispatchers.IO) { ... }` as it's involves a network call, and should be on the IO thread pool, not the default pool, which is used for CPU intensive tasks.

app/src/main/java/com/nextcloud/talk/serverstatus/ServerStatusRepositoryImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.nextcloud.talk.api.NcApiCoroutines
+import com.nextcloud.talk.data.user.model.User
+import com.nextcloud.talk.utils.ApiUtils
+import com.nextcloud.talk.utils.database.user.CurrentUserProviderNew
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import javax.inject.Inject
+
+class ServerStatusRepositoryImpl @Inject constructor(
```
rapterjet2004, line 22: I believe this should be annotated as a `@Singleton` for dagger to know it should only be instantiated once. [See Docs](https://docs.oracle.com/javaee/7/api/javax/inject/Singleton.html). This way, in case multiple repositories need access to server status information, you can share the instance of the implementation class, and importantly, it's state flow

app/src/main/java/com/nextcloud/talk/serverstatus/ServerStatusRepositoryImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Suppress("Detekt.TooGenericExceptionCaught")
+    private suspend fun serverStatus() {
+        val baseUrl = currentUser.baseUrl
+        val url = baseUrl + ApiUtils.getUrlPostfixForStatus()
+        _isServerReachable.value = try {
+            ncApiCoroutines.getServerStatus(url)
+            true
+        } catch (e: Exception) {
```
rapterjet2004: check if e is a `CancellationException`, if so throw it. [The CancellationException](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.coroutines.cancellation/-cancellation-exception/) is a special kind of exception that is needed for proper coroutine behavior. Will not throwing it cause your PR to fail? Maybe not for now, but it's considered [good practice.](https://stackoverflow.com/questions/76259793/is-it-necceary-to-rethrow-the-cancellationexception-in-kotlin)

app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
```diff
@@ -255,12 +260,21 @@ class ConversationsListActivity :
...
     }
 
+    @Suppress("Detekt.TooGenericExceptionCaught")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
 
         currentUser = currentUserProvider.currentUser.blockingGet()
 
+        lifecycleScope.launch {
```
rapterjet2004: https://github.com/nextcloud/talk-android/pull/5172/files#r2285759661

rapterjet2004 (changes requested):
Also please rebase and squash your commits. Other than that, looks good 👍🏾

rapterjet2004:
You have a lot of commits, make sure to rebase and squash them before pushing.

mahibi:
To be honest, i would have not approved this PR if i would have seen it.

1)
it does still not solve https://github.com/nextcloud/talk-android/issues/5038

2) 
As I already stated in 
https://github.com/nextcloud/talk-android/pull/5078/files#r2191942331
an additional request just to check a server state should not be done.

For a clean solution which can serve for all screens and no additional requests are done,
this should be handled on a lower level.
So instead to have one request that explicitly needs to be triggered to check the state, any http request should update the state. Adding a ResponseInterceptor to RestModule.java would be the way to go.

3) 
In this PR request to check the server state is done in ConversationListActivity in onCreate.
So the state would not be reset during usage of the app. 
It would be a bit better in onResume, but also then the request is still only done when going back to ConversationListActivity.

4) 
Handling of the unableToLoadMessages screen is buggy:

https://github.com/nextcloud/talk-android/pull/5172/files#diff-a3567d71e5bb91cd3a5b0f732565ce89ebb3edcaa51f45dfe36922cb1b4e7259R1037-R1047

`binding.offline.root.visibility` and `binding.unableToLoadMessages.root.visibility` can be shown both overlayed as they are not removed in the respective other condition.

When the device is offline, the server can't be reached of course. That's why both states are emitted in

https://github.com/nextcloud/talk-android/pull/5172/files#diff-613b4702fd5108bfa720e7a128cd700663fd46f7e5b8c5207c04088899cfb0b4R168-R173

5)
As the logic with ServerStatus check is buggy, the long polling is destroyed by this check when the device was offline:
https://github.com/nextcloud/talk-android/pull/5172/files#diff-613b4702fd5108bfa720e7a128cd700663fd46f7e5b8c5207c04088899cfb0b4R583-R585
As a result there is an endless loop in the sync method and messages won't load.

6)
Please solve only one topic in a PR, at least it should be related

I will try to fix some things for 22.0.0 RC1 now, but i suggest there should be followup PRs afterwards.

PR #3952, opened 2024-06-04
Offline support

mahibi:
resolve #217

Authors: @rapterjet2004 @mahibi 

[Now in Android Documentation for the App's Architecture](https://github.com/android/nowinandroid/blob/main/docs/ArchitectureLearningJourney.md)

NC Offline DB

### 🚧 TODO

- [x] Implement Network Monitor
- [x] Declare Synchronizer (it's implemented in each repository, using the various DAO's)
- [x] Implement ChatMessageEntity
- [x] Implement ConversationEntity
- [x] Abstract away ChatMessage from ChatMessageEntity and ChatMessageJSON
    - [x] Implement Mappers
    - [x] Implement data classes 
- [x] Abstract away Conversation from ConversationEntity and ConversationJSON
    - [x] Implement Mappers
    - [x] Implement data classes
- [x] Implement ChatDao
- [x] Implement ConversationDao
- [x] Implement local version storage using dataStore
- [x] Implement OfflineFirstChatRepository
    - [x] Model Flows
    - [x] Sync function
    - [x] Get Chat Messages
    - [x] implement on demand synchronization functions 
- [x] Implement OfflineFirstConversationRepository
    - [x] Model Flows
    - [x] Sync function
- [x] Research Synchronization strategies for settings as they don't require on demand Synchronization as Chatting does. 
- [x] chat block handling
- [x] "handleUpdateMessages" method must handle all it's system message types immediately
- [ ] Queued messages fail to be sent after connection gained
   - [ ] show them also while being offline
- [x] enter ConversationList in offline mode: 
   - [x] replace "Please check your internet connection" with red info bar as in chat
   - [x] avoid to show "Failed to fetch pending invitations"
- [x] enter chat in offline mode:
   - [x] avoid "Sorry something went wrong"
- [x] avoid "unread messages" bubble when some system message is received that is not shown in chat 
- [x] X-Chat-Last-Common-Read handling
- [x] rearrange entity attributes (to fit documentation order)?
- [x] double check/ fix data types (esp null allowed) for entities
- [x] final steps:
   - [x] comment in ".openHelperFactory(factory)" in TalkDatabase.kt
   - [x] set database migrations
   - [x] bump database version

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not needed
- [x] 🔖 Capability is checked or not needed 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set
[...]

PR #5182, opened 2025-07-22
Adding Migration Tests

### LINKED ISSUE #5168: Test Migrations
- Overview at #5165

[use this for reference](https://medium.com/androiddevelopers/testing-room-migrations-be93cdb0d975)

### Files
Migrations.kt

rapterjet2004:
- fixes #5168

Includes explicit tests for all manual migrations, and a implicit test for all migrations from version 10 (last version before offline support).

> Schema 10.json was corrupted, which resulted in test
failing. It was from the offline support days, where the
database was updated, but not the version, which updated
10.json instead of 11.json. Replaced with proper schema
from the 9-10 auto migration.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -82,6 +82,11 @@ android {
             }
         }
 
+        sourceSets {
+            // Adds exported schema location as test app assets.
+            getByName("androidTest").assets.srcDir("$projectDir/schemas")
+        }
+
         testInstrumentationRunnerArgument "TEST_SERVER_URL", "${NC_TEST_SERVER_BASEURL}"
         testInstrumentationRunnerArgument "TEST_SERVER_USERNAME", "${NC_TEST_SERVER_USERNAME}"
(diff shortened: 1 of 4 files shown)


mahibi:
Indeed the 10.json was mistakenly modified in https://github.com/nextcloud/talk-android/pull/3952/files, nice catch :+1: 

As a side note:
Normally, modifying json schema files in retrospect must be avoided. If a wrong schema was released, there are users out there where it was applied to their data. By modifying the schema after releasing them, we don't have a chance to understand the bugs out there. Think about schema JSONs as immutable history logs..

For this special case i think we a not rewriting history, but we are actually restoring the "truth". So i think it's the best thing to do..

Going forward...
One idea how we could avoid modifying schema jsons again: We could add checks on github that prevent PRs to be merged when they modify existing json schemas. 
Makes sense also for the files app if not already solved otherwise?
What do you think @alperozturk96 @tobiasKaminsky @rapterjet2004 ?

mahibi (approved):

Issue #5007, opened 2025-05-28
Conversation list flickers


skjnldsv:
### Steps to reproduce

Use Talk 21.1.0RC2
Have a list of conversations 
Switch back and forth to messages list and conversation list

### Expected behaviour

No flickers

### Actual behaviour

https://github.com/user-attachments/assets/e53761b8-94d3-4e4f-a6a9-0267cd6eb435

### Device brand and model

Samsung A33

### Android version

14

### Nextcloud Talk app version

21.1.0 RC2

### Nextcloud server version

31

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

mahibi:
This is a known behavior (although it seems to be more intense than on my devices).

For the conversation list we will have a look especially when migrating this screen to compose.

The avatars in chat may flicker because of a different reason.

skjnldsv:
> This is a known behavior (although it seems to be more intense than on my devices).

It was better a few weeks ago.
Last update I did (don't recall which version I was on), dramatically increased the issue :)

ariedov:
I have taken some time to investigate this one; it is a bit tricky.

There are a few issues in place:

#### ConversationListActivity issues

1. The `ConversationListActivity` is using the `FlexibleAdapter` to display the items, which does provide an `updateDataSet` method to update the list without animations. Unfortunately, passing `animate: false` doesn't really do much.
2. If to remove the item animator from the `recylerView`, the flickering goes away (as do all the other animations, but there is a workaround)

#### Image loading issues:

3. However, in that case, another flickering happens: the image loading makes the avatar flicker, and it's not the fault of coil, but the fact that the views don't seem to preserve, and the images load multiple times for seemingly the same item
4. I saw it on the "Note to self" chat, and tried just setting the drawable without using coil, but that didn't help, the image was overwritten.


```kotlin
fun ImageView.loadNoteToSelfAvatar() {
    val layers = arrayOfNulls<Drawable>(2)
    layers[0] = ContextCompat.getDrawable(context, R.drawable.ic_launcher_background)
    layers[1] = ContextCompat.getDrawable(context, R.drawable.ic_note_to_self)
    val layerDrawable = LayerDrawable(layers)
    val data: Any = layerDrawable

   // This does not help
    setImageDrawable(data)
}
```

I'll keep digging into it, since the flickering annoys me very much. I'll try to make sure the views don't get rerendered if not changed.

[...]

PR #5205, opened 2025-07-30
Fixes flickering in the conversation list

ariedov:
Resolve https://github.com/nextcloud/talk-android/issues/5007

### 🚧 TODO

- [x] Creates explicit equals and hashCode for the ConversationModel
- [x] Removes usage of coil for setting drawable image for the Note to Self conversation
- [x] Adds CircularDrawable helper Drawable to make the image circular without coil load and dispatch

I am a newbie here with Nextcloud, tried my best to make this pretty and keep unnecessary changes out of the PR.

Would appreciate the review of the newly created equals and hashCode to actually include the correct values. I am only guessing that the token and expiration values change too often, which makes the UI rerender on every call, hence the flickering. Removing that also improved the image loading performance, and the app feels more pleasant to open.

Let me know what needs to be changed. I'd appreciate having this fixed in the upcoming releases of the application.

Keep up the great work!


mahibi:
Thank you to contribute @ariedov We will review soon

ariedov:
Attempting to be a good citizen and fix the linter issues.

mahibi:
Hi @ariedov 
This is really great!! :1st_place_medal: 

Your solution works great, however as already stated by you it might be a bit invasive.

I had a closer look at the data class solution as well.
So this only flickers for the changed items because because a data class includes all properties in its internal `equals()` and `hashCode()`, even the ones not relevant to UI updates. 
It's not perfect, but already improves the situation a lot!

While your solution with the overwritten methods works perfectly fine and no flickering is seen at all, i'm concerned there will be side effects as the methods may be used (now or in the future) also in backend code and not only for the UI logic.

One solution for this would be to use `DiffUtil.ItemCallback` instead.
But this is not offered in the FlexibleAdapater lib but the lib relies on the equals method.
And migrating to normal recylcerview for this is senseless as we directly migrate to Jetpack Compose.

So my suggestion is:
Even though it is very tempting to take the solution with overwritten `equals`, i would go for the data class solution. Then the situation can hopefully be further improved when we migrate to Jetpack Compose and get rid of FlexibleAdapter.
I'm open on other opinions.

For the loading of the Note to self avatar: It works fine for me and the solution can be taken.

Are you okay with changing it to the data class solution?
It's fine to do it as additional commit, so the `equals` solution can be looked up if it's ever needed again.


Thanks again @ariedov :+1: :+1: 


PS: i'm on vacation next 2 weeks, my colleagues @sowjanyakch or @rapterjet2004 could merge..

ariedov:
DiffUtils was my first thought. The thing is, FlexibleAdapter actually implements it under the hood, and that is basically the root of the problem. That's why the equals/hashCode matter so much.

I am very happy for my contribution to be positively taken. Changing the `ConversationModel` now to be a data class, and removing the equals/hashCode.

Thank you for letting me be part of the project :)

mahibi:
We appreciate your efforts to contribute!
If you are interested in further contributions to this app, let us know and we can give you advanced access to the repo so you can create branches inside it.

For your last commit, DCO fails. Could you sign off the commit? See
https://github.com/nextcloud/talk-android/pull/5205/checks?check_run_id=47203134718

Also a rebase on latest master would be good

ariedov:
Oh, no, I forgot to do it. Do I create a new one? Amend the old one? Not sure what the protocol is here.
I guess if I rebase, I can just amend and force push. Would that be okay?

I use the app regularly along with my family, so I will keep working on it. Having access is nice I guess, but I don't think it matters that much.

mahibi:
I see you figured out the sign off. amend and force push is fine.
Don't worry about failing analysis check, this is okay for forks.
Would be one of the advantages that all checks run when you can have branches inside the repo..
I will let my colleagues know to add you, might take some time as some are on vacation.

ariedov:
Thank you very much. This is actually incredible.

Have a great vacation 🏖️

AndyScherzinger:
One thing I am experiencing now it that the conversation avatar isn't always correct.

If you have lenghty list of conversations, put the notes to yourself to the top via "favorite" status, then scroll down far enough for the item to be out of the rendered view port, than scroll up again, the notes to yourself item will show a (random) icon on another conversation, and will refresh at some point or forced after doing a pull to refresh on the recycler view

ariedov:
Oh no. I do not have a conversation list long enough for it to be outside the viewport, so I didn't notice. Will find some time to handle that

mahibi:
sounds like a recycler view issue.
ConversationItem#bindViewHolder should be a good place to start investigating.
I created https://github.com/nextcloud/talk-android/issues/5272 to keep track of it.
@ariedov just let us know if you plan to have a look or if we should take over

ariedov:
@mahibi I'd actually appreciate some help here, thank you for offering

Issue #5165, opened 2025-07-18
Improve test coverage


rapterjet2004:
Currently our test coverage isn't that great compared to the files app. While full coverage isn't necessary, the point should be to reduce regressions by catching bugs before they occur.

Basically, if we encounter a bug , quirk, or edge case involving our business logic, write a test case for it, so we can make sure it doesn't break again in the future.

This is particularly useful for anything involving room, datastore, date/time, hardware apis (`MediaPlayer`, `AudioRecord`, etc), and Networks (Especially anything involving synchronization or offline functionality).

In addition, existing tests in the androidTest directory should be moved to the unit test directory, as with proper separation of concerns, business logic should be in it's own layer, separate from the UI logic, so it can be tested in isolation.

 [According to best practices](https://developer.android.com/training/testing/fundamentals/what-to-test) we should cover. 

### Essentials

- Unit tests for ViewModels, or presenters.
- Unit tests for the data layer, especially repositories. Most of the data layer should be platform-independent. Doing so enables test doubles to replace database modules and remote data sources in tests. Maybe try using the `ComposePreviewUtils` and `ComposePreviewUtilsDao` I made earlier.
- Unit tests for other platform-independent layers such as the Domain layer, as with use cases and interactors.
- Unit tests for utility classes such as string manipulation and math.

### Edge cases

- Math operations using negative numbers, zero, and [boundary conditions](https://en.wikipedia.org/wiki/Off-by-one_error).
- All the possible network connection errors.
- Corrupted data, such as malformed JSON.
- Simulating full storage when saving to a file.
- Object recreated in the middle of a process (such as an activity when the device is rotated).

### Avoid

- Tests that verify the correct operation of the framework or a library, not your code.
- Framework entry points such as activities, fragments, or services should not have business logic so unit testing shouldn't be a priority

Issue #4711, opened 2025-02-14
Login Flow v2


mahibi:
Implement Login Flow v2 

Doc:
https://docs.nextcloud.com/server/19/developer_manual/client_apis/LoginFlow/index.html#login-flow-v2

See android files app:
- https://github.com/nextcloud/android/pull/12922
- https://github.com/nextcloud/android/pull/14461

This will also avoid issues like https://github.com/nextcloud/talk-android/issues/3386

The best approach would be to extract Login&Authentication Screens to https://github.com/nextcloud/android-common and use them for all Nextcloud android apps.

mario-spitze:
We still have problems while we login by username and password. Our nextcloud server working well behind an apache reverse proxy with subdirectory. If we try to login with some apps like the talk app for iOS and android we get requests like the following at the last login-step:
```
xx.xx.xx.174 - - [21/Feb/2025:10:02:24 +0100] "POST /bhhcloud/login/flow HTTP/1.1" 303 1069 "-" "Volla Volla Phone 22 (Nextcloud Talk)"
xx.xx.xx.174 - - [21/Feb/2025:10:02:25 +0100] "GET /server:https://myweb.de/cloud&user:username&password:1euf6i7sWv4VFoJ2TUFsb1ITnbO99lDdxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxA16gSLLL HTTP/1.1" 404 577 "-" "Volla Volla Phone 22 (Nextcloud Talk)"
```
I have no idea where this request have to be redirected to by the proxy. The last try was with the android talk app 20.1.1 installed by F-Droid an Nextcloud Server 30.0.6. 

The login flow documentation talks about URLs like this. (https://docs.nextcloud.com/server/31/developer_manual/client_apis/LoginFlow/index.html#obtaining-the-login-credentials) Now my question is, could this problem be solved by implementing the login flow v 2? I don't now witch version is used by other working apps. If this could solve my problem it would save some off my time. Otherwise i will walk on and open a new ticket.

mahibi:
I'm not sure why it fails for you but login flow v2 would solve this. However we might not find the time to switch to this in the near future

wegylexy:
The in-app browser doesn't support SSO that requires WebAuthn or opening another native app to complete the OIDC login flow.

mahibi:
@rapterjet2004 can you work on this?

It's okay to have an initial version by copy&paste from the files app.
[...]

PR #5215, opened 2025-08-01
Improving Login V2 - MVVM

rapterjet2004:
- #5165 
- https://github.com/nextcloud/talk-android/issues/4711

The Login pipeline is a critical component of the app. Without it, the app can't function. Therefore, It's crucial to have it be robust and testable. Edge cases from account management deletes, to duplicate users, to invalid tokens can then be tested for and debugged easier. This prevents code regressions and makes the code more reliable.

In addition, this makes it easier to eventually share the network code with the files app using the android commons library, which makes things easier on our end.

This PR does two things
1. Refactor `BrowserLoginActivity` to MVVM
2. Unit testing the newly created classes in isolation

A lot of this code can also be copy & pasted for use in other classes/developing other features. The new files are:

### `app`
- `BrowserLoginActivityViewModel` models the view state of the browser login activity. This sets the flow of the login
- `LoginRepository` main class, holds the login logic and other business logic
- `NetworkLoginDataSource` actually performs the network requests
- `LocalLoginDataSource` holds the account storage logic, abstraction over userManager and accountWorker 
- `LoginResponseModels` common data classes used to model Login response
### `/test`
- `LoginRepositoryTest`
- `NetworkLoginDataSourceTest`

The edited files are
- `BrowserLoginActivity` heavily refactored, now observes view model state and bubbles up user input
- `RepositoryModule`/`RestModule`/ `ViewModelModule` for dependency injection
- `build.gradle` added two new dependencies, Square's [MockWebServer](https://mvnrepository.com/artifact/com.squareup.okhttp3/mockwebserver) and Google's [Robolectric](https://chromium.googlesource.com/external/robolectric/) for running tests without the overhead of the emulator

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -356,10 +356,10 @@ dependencies {
     testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutines_version"
 
-    testImplementation("com.google.dagger:hilt-android-testing:2.57")
+    testImplementation 'org.junit.vintage:junit-vintage-engine:5.13.4' // DO NOT REMOVE
     testImplementation "androidx.room:room-testing:${roomVersion}"
-    testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2' // DO NOT REMOVE
     testImplementation("com.squareup.okhttp3:mockwebserver:$okhttpVersion")
+    testImplementation("com.google.dagger:hilt-android-testing:2.57")
     testImplementation("org.robolectric:robolectric:4.15.1")
 }
 


sowjanyakch:
@rapterjet2004 - I tested this and the login is not working as expected. After entering credentials on the login screen and clicking Login, I grant access, but nothing happens—the app simply redirects back to the login screen. This issue occurs when an account is already configured and I try to log in with another account.

When no accounts are configured in Nextcloud, adding a new account only works after multiple attempts (around 4–5 tries). Sometimes I get a state token missing error, sometimes I get connected to Talk Web, and other times I’m redirected back to the login screen again.

mahibi:
We have some special situation here that login v2 does not work most of the time (or at all?). With and without this PR.
I am not able to login anymore, so using only QR code currently! 

When loginV2 was initially merged for android talk it at least worked most of the time for me, but maybe it was just luck..

As the PR introduces a better architecture and will make it easier to find bugs, i will merge it despite i was not able to login.

We may have to switch back to loginv1 in the next days and analyze what's wrong with loginv2 on a different branch..

mahibi (approved):

Issue #5207, opened 2025-07-31
Text size is too small relative to other apps


rrthomas:
### Steps to reproduce

1. Observe text size in messages.
2. Compare to other apps, e.g. Signal, Snikket, Google Messages, WhatsApp. All of these have larger text.

### Expected behaviour

The text should be approximately the same size as other apps.

### Actual behaviour

The text is smaller than other apps.

### Device brand and model

Samsung A32 5G

### Android version

13

### Nextcloud Talk app version

21.1.2

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

### Additional information

The text size can be changed via the Android system settings, no problem. However, because the size for message text in Talk is smaller than in other apps, it's not possible to adjust the system setting to get a satisfactory text size for all apps.

It looks like the text size for messages in Talk needs to be a few points bigger in order to match the other apps.

ariedov:
I do experience this also. Some other messaging apps allow changing the font dynamically from accessibility settings, however that's quite a huge effort.

Another solution would be to just increase the font size slightly in the chat.

rrthomas:
> I do experience this also. Some other messaging apps allow changing the font dynamically from accessibility settings, however that's quite a huge effort.

I think that would be a bad idea. Accessibility settings should not on the whole be per-app, the user should only have to set them in the system.
 
> Another solution would be to just increase the font size slightly in the chat.

Exactly what I suggested, and the obvious solution here. The other messaging apps I've tried all seem to have the same size, more or less, for their message texts.

One other thing I've noticed is that the size that Talk currently uses for chats is the same size that both it *and other apps* use in the preview of the most recent chat in the chats list. Here, Talk matches the other apps fine. It's simply that the text size in the individual chat screen is too small.

Issue #4984, opened 2025-05-23
Add option to change font size in the Talk Android app


juro200:
Hi,
I’d like to suggest a feature to adjust the font size in the Talk Android app.
This would improve readability, especially for users with visual difficulties.
Currently, the app relies only on system font settings, but an in-app option (like small / medium / large) would be very useful.
Thanks!

sylvaticus:
I support this, or at least increase the default font.
For some reason, the text in this app is smaller than the "normal" font in similar apps (Signal, WhatsApp...) making the chat conversation difficult to read..

![Screenshot_2025-06-23-22-52-00-764_org.thoughtcrime.securesms.jpg](https://github.com/user-attachments/assets/bfe02aa1-59ec-46dc-9959-a5de5a449ef9)

![Screenshot_2025-06-23-22-52-32-260_com.nextcloud.talk2.jpg](https://github.com/user-attachments/assets/ab6e4082-7b8d-4535-9ef4-02b2dbb24371)

PR #5218, opened 2025-08-04
Increase default chat message size

ariedov:
Resolves https://github.com/nextcloud/talk-android/issues/5207 and https://github.com/nextcloud/talk-android/issues/4984

This does not implement the ability to change the font size, however it does makes the default font size same as other messaging apps.

Telegram has 16sp as default as shown here
image

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image  |  image

- [x] Changed default chat message size from 14sp to 16sp


rrthomas:
Many thanks for the quick fix!

PR #5227, opened 2025-08-05
Improve AccountDialogFragment

### LINKED ISSUE #5095: Improve ChooseAccountDialogFragment
The account chooser `ChooseAccountDialogFragment` has the problem that accounts are shown with a delay because it's waiting for

`invitationsRepository.fetchInvitations(userItem)`

to finish, just to add a badge..

It should be improved to immediately show all available accounts and just update the badge whenever `fetchInvitations` is finished.

Maybe for some reason related to this, accounts are not shown at all (see https://github.com/nextcloud/talk-android/issues/5093)

If necessary / possible without too much effort, this should also be done:
- convert to kotlin
- get rid of FlexibleAdapter
- migrate to Compose

Also, keep an eye on the files app (as the code was copied from there back then) if there are some fixes.
If it makes sense and it's possible, it could also be extracted to android-common lib

sowjanyakch:
fix #5095

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
settings option is shown in the accounts dialog when the device is offline. Handling settings screen will be done in a different PR. I created a seperate issue for it #5440

mahibi:
Can you squash into one commit?

nickvergessen:
> Can you squash into one commit?

Let's save the time here if it's too complicated.
If we can deduplicate/merge commits with the same name or purpose, it of course makes sense.

rapterjet2004 (approved):

PR #4828, opened 2025-03-27
Modify some strings

mahibi:
- change translations
- change rules to work with translations

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #5242, opened 2025-08-12
Improved Message Actions UX

### LINKED ISSUE #5231: Confirm deletion of a post
Deletion of a post happens immediately. So if you press the delete button by accident the post is gone.

Please build a simple confirmation before execute the deletion process.

rapterjet2004:
- fixes #5231 

Adding a deletion dialog would be bad UX, and would be a pain to maintain. Here I decided to move the edit button up top, since it's most likely to be used alongside reply, and added a barrier in between the reversible actions and the non reversible deletion action, which should prevent accidental taps (unless you have thumbs like [this guy](https://www.google.com/url?sa=i&url=https%3A%2F%2Ftenor.com%2Fsearch%2Fspy-kids-thumb-gifs&psig=AOvVaw2vfhJk3TmXlk6xHmR9EC1G&ust=1755104059751000&source=images&cd=vfe&opi=89978449&ved=0CBkQjhxqFwoTCOD13aHehY8DFQAAAAAdAAAAABAK))

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot 2025-08-12 at 11 50 06 AM | Screenshot 2025-08-12 at 11 47 47 AM

### 🏁 Checklist

- [ ] 📅 Milestone is set


sowjanyakch:
@rapterjet2004 - I think it is always better to show an alert dialog before deleting a message in order to prevent accidental data loss just like other messaging apps does.

rapterjet2004:
[Screen_recording_20250813_101513.webm](https://github.com/user-attachments/assets/78098915-7b53-46b0-909f-190a1c5f6ef0)


How's this

nimishavijay:
> Adding a deletion dialog would be bad UX, and would be a pain to maintain

Checked other apps (WhatsApp, telegram, signal) and they all have a confirmation dialog so I'd agree with @sowjanyakch and push for this.

> Here I decided to move the edit button up top,

- [x] Great call! :) I'd still suggest we keep reply/start a thread on top though, it would definitely be the feature used way more than edit. So the order can be:
` Reply, Start a thread, Forward, Edit, Copy, Mark as unread, Translate, Add to Notes, Share, Delete ` 

Regarding the design itself, I have some concerns about accessibility:
- [x] Not sure if red text has enough contrast, so we can keep the text default color with a red icon
- [x] The red icon itself would be enough of an indicator so then we don't need the red line anymore :) 
- [x] The "Are you sure?" appearing in the same place as the delete action can cause issue for screen readers, so I'd push for a confirmation dialog

Other small visual changes:
- [x] I see that the share and forward button have the same icon. We can use the updated [material symbol for forward](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23222222&icon.query=forward&selected=Material+Symbols+Outlined:forward:FILL@1;wght@400;GRAD@0;opsz@24)
- [x] There is a bit too much space between the icons of all the items and the text. It can be reduced by about 8px I'd say

sowjanyakch:
> * [x]  Great call! :) I'd still suggest we keep reply/start a thread on top though, it would definitely be the feature used way more than edit. So the order can be:
>   `Reply, Start a thread, Forward, Edit, Copy, Mark as unread, Translate, Add to Notes, Share, Delete`
> 

@nimishavijay - Where do you want to place the option 'Remind me later'?

@rapterjet2004 

Screenshot 2025-08-19 at 13 48 49

Please modify the dialog header to 'Delete message' and the text to 'Are you sure you want to delete this message'?

nimishavijay:
@sowjanyakch @rapterjet2004 the "Remind me later" can go after "Mark as unread" :)

sowjanyakch (changes requested):
Image

- [x] The option has to be Delete.

Image

- [x] Rename dialog header to 'Delete message'  (see: https://github.com/nextcloud/talk-android/pull/4828)

- [x] We could theme buttons and dialog background just like the dialog used for deleting a conversation.

PR #5266, opened 2025-08-19
Update to latest status icons and status dialog design from server

AndyScherzinger:
### 🖼️ Screenshots

🏚️ light | 🏡 dark
---|---
Screenshot_20250819_104848|Screenshot_20250819_104824

### 🚧 TODO

- [ ] discuss invisible color for dark mode

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi:
Looks good!
and finally no cut off status descriptions any more :+1: 

Is there some overview issue? 
- [x] I see that on sermo there is an additional status "busy" now? Is it missing here? Not sure which icon should be used..
- [x] For the invisible color for dark mode i would prefer a light circle as it's only shown there..

AndyScherzinger:
> Is there some overview issue?

yes, https://github.com/nextcloud/files-clients/issues/103

> I see that on sermo there is an additional status "busy" now? Is it missing here? Not sure which icon should be used..

Not in scope of this PR but to be added via an extra one which has the server check for the capability / version

> For the invisible color for dark mode i would prefer a light circle as it's only shown there..

Same, which is why I added it as a todo to this PR to clarify with designers

kra-mo:
> discuss invisible color for dark mode

It matching the foreground color makes sense to me.

Issue #5272, opened 2025-08-20
Note to self avatar shows random icon from another conversation


mahibi:
see https://github.com/nextcloud/talk-android/pull/5205#issuecomment-3187866026

ariedov:
I cannot reproduce the issue because I do not have a long enough conversation list, however, let's consider a different scenario when this could happen:

1. I have set the drawable explicitly without coil (like I do for Note to Self)
2. I scroll long enough and have to load the image with coil now
3. Loading fails
4. The default for coil seems to be just not doing anything, so the previously set image would still be present.

This scenario is very similar to what is being reported and seems reproducible with some code/network tampering.

Suggested fixes:

1. Provide a fallback for coil to display in case of failed loading
2. Set a default image in the `ConversationItem.showAvatar` before loading anything

The `loadFirstLetterAvatar` is actually a drawable that is set without using coil, so it could be a sensible default to apply in the `ConversationItem.showAvatar`. Thoughts @mahibi?

ariedov:
I spent another minute checking the image loading code, and it does have a fallback, which is not what is getting reported. This seems to leave us with the option to reset the image before loading a new one in the `showAvatar` method, or I am missing something crucial about how the image is being loaded into the ImageView.
 
The best way to handle it is to reproduce and debug, all I am doing is guessing...

ariedov:
I believe this should be closed as fixed

PR #5294, opened 2025-08-25
Handles wrong image shown for some conversations

ariedov:
Aftermath of this https://github.com/nextcloud/talk-android/pull/5205

An unexpected side effect was reported: the "Note to Self" image randomly appears for different chats
https://github.com/nextcloud/talk-android/pull/5205#issuecomment-3187866026

Attempting to fix by explicitly removing the image from the viewholder before loading the new one.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.kt
@@ -22,6 +22,7 @@ import android.view.View
 import android.widget.RelativeLayout
 import androidx.core.content.ContextCompat
 import androidx.core.content.res.ResourcesCompat
+import coil.dispose
 import com.nextcloud.talk.R
 import com.nextcloud.talk.adapters.items.ConversationItem.ConversationItemViewHolder
 import com.nextcloud.talk.application.NextcloudTalkApplication.Companion.sharedApplication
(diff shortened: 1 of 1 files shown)


ariedov:
@AndyScherzinger can you please take a look at the test build here, I'm trying to fix https://github.com/nextcloud/talk-android/issues/5272, the issue with Note to Self image popping up randomly in the conversation list. Thank you!

mahibi:
i did not try yet, but this solution looks very promising ..classic recyclerview things.

AndyScherzinger:
Tested and unfortunately still broken @mahibi @ariedov

ariedov:
Bummer. This might actually mean that the Note to Self icon is requested on the wrong conversation.

I'll dig a bit more to understand why that would happen, and will have more builds to test out to get to the root of the issue.

ariedov:
@AndyScherzinger @mahibi I might have played it a bit too safe, clearing the image at the top of the method now. Please take a look.

AndyScherzinger:
@ariedov @mahibi still loads the wrong conversation avatar for me 😢

ariedov:
Right idea, wrong implementation. This one is gonna fix it. Just using the `dispose` from coil to achieve the right images being loaded to the right ImageViews.

@AndyScherzinger @mahibi please one more round :)

AndyScherzinger:
Woohoo 🎉 

Tested successfully - can't reproduce, icon is consistently correctly loaded 👍

AndyScherzinger (approved):

AndyScherzinger:
Thanks a lot for the fix @ariedov 🎉

ariedov:
Thanks for the patience, happy to contribute :)

PR #5297, opened 2025-08-26
don't overlap screen content with nav bar

### LINKED ISSUE #5269: nav bar for SDK 35 and up overlaps
nav bar for SDK 35 and up overlaps (at least for incoming call screen)

Image

- [ ] check all screens again

sowjanyakch:
fix #5269

- [X] fix padding for gesture and three bar navigation.
- [X] fix display cutout for gesture and three bar navigation. (content will not be clipped in landscape mode).
- [ ]  set navigation bar colors for gesture (transparent) and three bar navigation (translucent).

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -1574,6 +1574,7 @@ class ChatActivity :
         } while (true && pos >= 0)
     }
 
+    @Suppress("LongMethod")
     private fun initMessageHolders(): MessageHolders {
         val messageHolders = MessageHolders()
         val profileBottomSheet = ProfileBottomSheet(ncApi, conversationUser!!, viewThemeUtils)
--- a/app/src/main/java/com/nextcloud/talk/contacts/ContactsScreen.kt
+++ b/app/src/main/java/com/nextcloud/talk/contacts/ContactsScreen.kt
@@ -18,6 +18,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.nextcloud.talk.R
 import com.nextcloud.talk.contacts.components.ContactsAppBar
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -3242,7 +3242,7 @@ class CallActivity : CallBaseActivity() {
     }
 
     override fun suppressFitsSystemWindows() {
-        binding!!.callLayout.fitsSystemWindows = false
+        binding!!.callLayout.fitsSystemWindows = true
```
mahibi: The function name "suppressFitsSystemWindows" now is the opposite of what it does.
If true is the default, it can also be deleted.

Any clue why false was needed beforehand?

mahibi (changes requested):
The PR destroys the edge to edge experience (well it's only about the navigation bar to be exact as we always keep the searchbar on top).

with this PR (not wanted):

grafik

without this PR (wanted):

grafik

(If overlap for navigationbar is too confusing, it could be semi transparent but content should reach the bottom.)

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
sowjanyakch: When in pip mode, it's value is false, there the entire content should occupy the window. Doesn't make sure to modify this value, so changing it back to false

sowjanyakch:
@mahibi - By default, xml and compose screens should have translucent navigation bar background for three bar navigation. Composable screens get translucent background by default, whereas xml screens get transparent navigation bar. I tried changing the navigation bar color dynamically for xml screens but it did not work. I did some work on padding for xml and compose screens in this PR. I added bottom padding also for composable screens so that all screens have a similar background (common background instead of different navigation bar colors for xml and compose but it destroys edge to edge experience). I can work on fixing navigation bar colors in a different PR or we could wait until all screens are modified to compose, so that  it can easily handled by just removing bottom padding for composable screens. What do you think?

mahibi:
i won't find the time to have a close look. I will merge this PR for RC2 because having overlapping contents like in https://github.com/nextcloud/talk-android/issues/5269 fixed is more important than the edge to edge support in my opinion.

Could you give it another try in a followup PR to fix the navigation bar? If edge to edge is not possible then at least coloring would be great.

PR #5300, opened 2025-08-27
Fix translations for the Delete message dialog

rapterjet2004:
>💥 Some translation strings have a different English source text between branches:
name="nc_delete_message"

I think, this should trigger the translation bot to create new translations for `delete_message` which replaces the `nc_delete_message`. Same text, just a different key.

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
i will just change the translated text also on stable branches. This will solve the transifex warning.

Whenever changing string values, the according PR should also be backported. 
In this case we did not do it for https://github.com/nextcloud/talk-android/pull/5242
If a backport is not wanted, there should be new strings introduced instead.

mahibi:
transifex run was successful

PR #5339, opened 2025-09-08
Add a rounded background to reply message part

AndyScherzinger:
### 🖼️ Screenshots

* Also fixing the wrong coloring (see first 2 replies) to always use primary if the original message and the current user match.
* Also slightly lowered the border radius which I would further reduce in a follow up to be closer to MS Teams and WhatsApp and less like Signal/Telegram

🏚️ Before | 🏡 After
---|---
before|after
Screenshot_2025-09-08-17-19-47-41|840-4096-max

### 🚧 TODO

- [x] get design feedback

### 🏁 Checklist

- [x] 📅 Milestone is set


nimishavijay (approved):
Looks really nice and sleek!

mahibi:
this is great :+1:

AndyScherzinger:
@nimishavijay I updated the screenshots, in light mode it would look nicer if I go for a light tone as a background, no? Teams does a light background on light and dark on dark mode while Whatsapp always does a dark mode but with a lower opacity / higher transparency.

I'd go with dark/light. WDYT?

EDIT: Go with light one

Issue #5346, opened 2025-09-09
Investigating the Chat Bubbles


ariedov:
I am trying to figure out how to make the chat screen more aligned with the industry standard. So currently the `master` branch has the font size increased, which is definitely welcome and looks better, but it hides quite a few messages from the screen. Also, the bubble padding and line height seem exploded in that case.

So I removed the line spacing multiplier and reduced the vertical padding from 12dp to 8dp, and now, even with the bigger font, I see an extra message compared to the "current" version.

The Telegram font size still seems different from the "increased font size" version of the app, but that is probably because they are using a different font.

I see 17 message bubbles on Telegram, and I also see 17 on Bigger Font Smaller Padding. Only 16 though currently, and only 14 with the bigger font.

I think small padding and big font is the way to go.

Telegram | Current
---|---
<img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/74716aa3-7cd8-4096-b975-ff6f4bb1c91b" /> | <img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/ad66ee54-a8e5-4bdf-bdb6-43af13413b8a" />

Telegram | Bigger Font
---|---
<img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/74716aa3-7cd8-4096-b975-ff6f4bb1c91b" /> | <img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/6be09a3c-7f26-4f58-91ea-3574718771d7" /> 

Telegram | Bigger Font Smaller Padding
---|---
<img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/74716aa3-7cd8-4096-b975-ff6f4bb1c91b" /> | <img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/68632ba5-2a57-43f3-96e7-85beca87ec87" />

Telegram | Small Padding | Bigger Font | Current
---|---|---|---
[...]

PR #5347, opened 2025-09-09
feat: Improves consistency with popular messaging apps

ariedov:
My personal opinion on the chat bubbles padding and line height. Regarding this https://github.com/nextcloud/talk-android/issues/5346

Telegram on the screenshot here uses a different font, so it's not quite apples to apples, but still the change is visible.

### 🖼️ Screenshots

🏚️ Previous | 🏡 Current
---|---
Image | Image

Looks more concise and fits more messages on the screen, while keeping the font larger to make messages easier to read. I like how it turned out, appreciate a review.


mahibi:
Nice improvement @ariedov :+1: 
if you want, feel free to play around with the right-padding as suggested in https://github.com/nextcloud/talk-android/issues/5346#issuecomment-3271601985

One side note just in case you are tempted to make further optimizations in the future:
The viewholders esp. for the chat are quite messy and use duplicated code. We plan to migrate to Compose which will replace them and also we plan to remove Chatkit ( https://github.com/nextcloud/talk-android/issues/1552).
So further XML design optimizations may have a limited lifespan until Compose replaces everything.
Nevertheless it's really helpful to already have improvements to the design which can then be rebuilt with Compose.

ariedov:
Thanks! I am very used to my code having a limited lifespan, just happy to contribute :) I would also like to play with the padding a bit more, as suggested.

ariedov:
Okay, I tried the horizontal padding, and it requires more tinkering than I am ready to invest at the moment. Let's leave that for a future PR. Waiting for a review :)

nimishavijay:
Looking great! The reduced padding is really nice! As for the line height, it is intentionally set to 1.2 to allow better readability and in fact on web it's even more (1.5 which is a standard recommended line height), so I would recommend we stick to 1.2. If you're interested you can read up more in the [Material Design 3 typography guide](https://m3.material.io/styles/typography/applying-type#bce6f69a-1222-43fe-82b6-ce6c116ee52a) :)

AndyScherzinger:
I extracted the vertical padding change into #5356

ariedov:
Appreciate it, thank you!

AndyScherzinger:
With the designers' decision to keep the current line height multiplier I'd close the PR if that is fine with you @ariedov - trying to do some housekeeping

ariedov:
Yeah, sure. A little busy now with school starting, feel free to take over whatever.

PR #5348, opened 2025-09-09
Improve link-preview visuals

AndyScherzinger:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi:
nice

nimishavijay:
Assuming the screenshots are the same as the ones posted on the chat, looks good! I would only suggest to add a 4dp space between the preview and the timestamp row as rn they look very close.

AndyScherzinger:
>  I would only suggest to add a 4dp space between the preview and the timestamp row as rn they look very close.

I actually added 8dp 😆 
So the screenshots aren't up-to-date

AndyScherzinger:
Updated to 4dp as discussed

PR #5387, opened 2025-09-16
Background blurring 👤✨

### LINKED ISSUE #5349: Allow background blur during video calls
Currently, the Web and iOS clients allow background blurring during video calls, but the Android app doesn't. iOS has this feature built into the SDK, but there is no equivalent for android.

One solution could be to use [MediaPipe](https://github.com/google-ai-edge/mediapipe), an open source framework AI/ML with first class support for Android.

The light weight (250 KB), on device image segmentation model is available for commercial use (Apache License V2) here is [more info](https://storage.googleapis.com/mediapipe-assets/Model%20Card%20MediaPipe%20Selfie%20Segmentation.pdf) about the model.

It would require capturing the camera frames using the MediaPipe in order to segment the image, blurring the background mask, and then passing the edited frames as a stream through WebRTC.

Due to hardware concerns, this feature should be off by default, and enabled in the more call actions dialog.

rapterjet2004:
- fixes #5349 

This PR implements Background Blurring, an optional (by default off) effect that, like web and iOS, uses a low-memory on-device AI model to create a mask between the foreground/background, before using openCV to perform gaussian blur, along with several other file format conversions, necessary for image processing Y'UV frames to RGBA and vice-versa. In the future this PR also makes it easier to implement additional background effects, like background images, background videos/gifs, party hats, company logo overlays, and more.

Adds new files
- `BackgroundBlurFrameProcessor` an implementation of `VideoProcessor` which is an interface that WebRTC provides to perform operations on camera frames, and `ImageSegmentationHelper.SegmenterListener`, which is a small interface I wrote to handle asynchronous passing of camera frames from MediaPipe (You can actually handle it synchronously too, since the VideoProcessor operates on it's own thread, but this is cleaner)
- `ImageSegmentationHelper` an edited version of [google's](https://github.com/google-ai-edge/mediapipe-samples/blob/d3671ea3ef4387b28b5bd22712ce73d911a6270f/examples/image_segmentation/android/category_mask/app/src/main/java/com/google/mediapipe/examples/imagesegmenter/ImageSegmenterHelper.kt). Basically a wrapper over MediaPipe's `ImageSegmenter` which uses the on device AI Model to segment images between the foreground and background.
- `BlurBackgroundViewModel` handles the state of the blur background functionality, which is by default off
- `download_model.gradle` a small gradle file to download the on device AI Model on compile time, if not already found. The model is pretty light, just a couple hundred KBs. Also it's licensed for Commercial use under Apache v2.
- `background_replace_24px` - Background Blur icon

### 🖼️ Screenshots

Mask | Post Process
---|---
Screenshot 2025-09-16 at 11 17 07 AM | Screenshot 2025-09-16 at 11 19 03 AM

### 🚧 TODO

- [x] Fix bug where turning off background blur stops frames from being sent over WebRTC. ( it was a timestamp issue ) 
- [x] Optimize object creation, passing, and processing ( still opportunities for optimization, but trivial time savings )
- [x] Research GPU acceleration ( not maintainable, openGL ES has too high learning curve for future devs )
- [x] Bug on orientation change, the camera is flipped upside down, change the flip code on detection of landscape in `onFrameCaptured`
- [x] The blur background processor crops the Video Frame to square, look into preserving aspect ratio ( Not possible the Selfie Segmentation Model only outputs square or landscape )

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ freeline_project_description.json
 **/__pycache__/
 
 /gradle/verification-keyring.gpg
+/app/src/main/assets/selfie_segmenter.tflite
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -238,7 +238,7 @@ class CallActivity : CallBaseActivity() {
     private val peerConnectionWrapperList: MutableList<PeerConnectionWrapper> = ArrayList()
     private var videoOn = false
     private var microphoneOn = false
-    private var isVoiceOnlyCall = false
+    var isVoiceOnlyCall = false
     private var isCallWithoutNotification = false
     private var isIncomingCallFromNotification = false
     private val callControlHandler = Handler()
@@ -1214,7 +1214,11 @@ class CallActivity : CallBaseActivity() {
     }
 
     private fun isCameraFrontFacing(enumerator: CameraEnumerator?): Boolean? {
-        val deviceNames = enumerator!!.deviceNames
+        if (enumerator == null) {
+            return false
+        }
+
+        val deviceNames = enumerator.deviceNames
 
         // First, try to find front facing camera
         for (deviceName in deviceNames) {
@@ -1319,6 +1323,7 @@ class CallActivity : CallBaseActivity() {
             } else {
                 binding!!.cameraButton.setImageResource(R.drawable.ic_videocam_off_white_24px)
                 binding!!.switchSelfVideoButton.visibility = View.GONE
+                blurBackgroundViewModel.turnOffBlur()
             }
             toggleMedia(videoOn, true)
         } else if (shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)) {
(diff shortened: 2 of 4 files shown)


sowjanyakch:
Tested this PR.

1. When background blur is enabled, the front camera is slightly zoomed compared to when blur is off.
2. For audio-only calls, the background blur button should be hidden to avoid unnecessary resource usage. Wdyt?
3. If you answer a video call, turn on background blur, and then stop your camera and then turn on camera, the person on the other end sees a blank screen. The video is displayed only a certain delay. (here delay is noticeable).
4. If you answer a video call, turn off background blur, and then stop your camera and then turn on camera, the person on the other end sees a blank screen. The video is displayed only a certain delay. (here delay is small).

rapterjet2004:
> 1. When background blur is enabled, the front camera is slightly zoomed compared to when blur is off.

Not possible to fix, as the Selfie Segmentation Model only outputs square or landscape. Besides that, I removed the option for audio only calls, and made sure to turn off background blur when the camera is off, to avoid errors with state.

sowjanyakch:
Tested this on Pixel 9 and it works. I earlier tested this on old pixel device, and it shows some distortion when camera is turned off/ turned on even without this PR.

PR #5419, opened 2025-09-29
Scroll position maintains position on lifecycle changes

### LINKED ISSUE #5398: Scroll position resets to bottom when resuming app from background
### Steps to reproduce

1. Open the Nextcloud Talk Android app.
2. Enter a chat with a long history of messages.
3. Scroll up to view earlier messages.
4. Switch to another app or press the Home button.
5. Return to Nextcloud Talk.

### Expected behaviour

The chat should remain at the same scroll position where I left off, so I can continue reading older messages without interruption.

### Actual behaviour

Upon reopening the app, the chat view automatically scrolls to the most recent messages at the bottom, discarding my previous scroll position.

### Device brand and model

Google Pixel 6a

### Android version

Android 16

22.0.0 RC3

### Nextcloud server version

31.0.9

### Talk version

21.1.5

None

Yes

Yes

### Android logs

_No response_

### Server log

```bash

```

### Additional information

_No response_

rapterjet2004:
- fixes #5398 
- See #5049

image

image

image

[Screen_recording_20251002_121623.webm](https://github.com/user-attachments/assets/76c032e4-5c89-46a5-a370-05c65b5f1270)

It works by keeping track of the top most bounds and position of first fully visible message in the adapter onPause(). After scrolling to the position, it also scrolls by a tiny difference between the old and new bounds, to be the exact position where the user left off. I need to test this out with extra large messages, and other large files, but this should work as is.

> Don't worry about the large diff, most of that was auto-generated in 22.json from updating the DB, actual diff is +128

### 🚧 TODO

- [x] Create ScrollPositionState
- [x] Update the Entity
- [x] Update the DB version
- [x] Update the converters
- [x] Retrieve scroll position state from conversation object on resume
- [x] Store scroll position state to conversation object on pause
- [x] Restore with proper offset
- [x] test it out a bit more, it doesn't always seem to be reliably saving. Apparently, might have just been an emulator issue, seems that saving to storage is slow at times.
- [ ] Need to figure out unread messages and last reading messages scrolling, should I remove these or make them a popup instead?

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/schemas/com.nextcloud.talk.data.source.local.TalkDatabase/22.json
+++ b/app/schemas/com.nextcloud.talk.data.source.local.TalkDatabase/22.json
@@ -1,0 +1,766 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 22,
+    "identityHash": "c212d18052ee2cdabdb16ebb1b0c004f",
+    "entities": [
+      {
+        "tableName": "User",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `userId` TEXT, `username` TEXT, `baseUrl` TEXT, `token` TEXT, `displayName` TEXT, `pushConfigurationState` TEXT, `capabilities` TEXT, `serverVersion` TEXT DEFAULT '', `clientCertificate` TEXT, `externalSignalingServer` TEXT, `current` INTEGER NOT NULL, `scheduledForDeletion` INTEGER NOT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "userId",
+            "columnName": "userId",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "username",
+            "columnName": "username",
+            "affinity": "TEXT"
+          },
+          {
+            "fieldPath": "baseUrl",
+            "columnName": "baseUrl",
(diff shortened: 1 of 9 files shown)


korea821:
I tested the build and the behavior is improved but still inconsistent on my device.

**Steps to reproduce**
1. Open a chat with a long message history.
2. Scroll up several screens to read older messages.
3. Press Home (or switch to another app).
4. Return to Nextcloud Talk.

**Expected**
The chat remains exactly at the previous scroll position.

**Actual**
Intermittently, the list jumps straight to the newest messages at the bottom.
Other times it restores near the previous spot, but the position is shifted downward (offset by a few messages / part of a screen).

**Video**
Reproduction clip (jump-to-bottom and slight offset):  https://youtube.com/shorts/86SbaREIeBA

**Notes**
Reproduces most reliably with very long histories.

mahibi:
Q from Julius in chat:

> It conflicts with the earlier approach of auto-scrolling to the last read message. Should I remove that ability, or perhaps have it be toggled as an option in Conversation Settings?

A: i don't think that an option is the way to go.
I would expect:
when entering the chat from the conversation list, always show last unread message.
when the chat was in background an coming back, always show the last scroll state..

Also: i don't think storing the scroll position per conversation makes sense. Other chat apps also don't behave like this. It would be more confusing than helpful.

When testing, it seemed the scroll position was only saved for some conversations and only quite inaccurate.

I suggest to focus more on moving to Compose. Then we could do these things with more modern solutions (remembering the state..)

rapterjet2004:
> I suggest to focus more on moving to Compose. Then we could do these things with more modern solutions (remembering the state..)

Makes sense, closing

PR #5448, opened 2025-10-08
fix http access?

mahibi:
just a guess why http access might be failing for some users

### 🏁 Checklist

- [ ] 📅 Milestone is set


nickvergessen:
Fixes HTTP for me, but self-signed HTTPS cert is sending me to a loop through the "Enter Server URL" screen now.

Issue #869, opened 2020-07-04
App mutes microphone in lock screen


Jookia:
I've experienced this on two phones in slightly different ways, but Nextcloud likes to mute the microphone during the lock screen or something like that.

On my Galaxy S2 running LineageOS when I lock the phone my microphone stops working and after unlocking I hate to mute then unmute the microphone for it to work again.
On my friend's Huawei P30 Lite it mutes whenever the phone is locked and unmutes when unlocked.

Exoyds:
I too have this issue on Xiaomi Mi Mix 3 5G, and can provide logs as needed.

Jookia:
Sure, could you post a log of you starting a call, then locking the
screen and continuing the call, then unlocking, then unmuting to get the
call working again?

Exoyds:
Strukur AG HPB Backend:
[hpblogs.txt](https://github.com/nextcloud/talk-android/files/5278611/hpblogs.txt)

Janus:
[januslogs.txt](https://github.com/nextcloud/talk-android/files/5278613/januslogs.txt)

NATS:
[natslogs.txt](https://github.com/nextcloud/talk-android/files/5278614/natslogs.txt)

Coturn:
[coturnlogs.txt](https://github.com/nextcloud/talk-android/files/5278616/coturnlogs.txt)

Partial logcat log where I think it happened:
[partiallogcat.txt](https://github.com/nextcloud/talk-android/files/5278615/partiallogcat.txt)

Full logcat log:
[fulllogcat.txt](https://github.com/nextcloud/talk-android/files/5278612/fulllogcat.txt)

Jookia:
I can't make anything interesting out of the logs, though I'm not an Android programmer.

Could you provide more information about your system? What Android version is it? Which Talk version are you on? Are you on the latest Nextcloud Talk version, 8.0.9?

The only things I really understand is that Talk's code that accesses the microphone is MagicAudioManager, which uses android's AudioManager and WebRTC. Maybe the bugs are somewhere there?

Have you made sure battery optimization and everything else is disabled for the app?

Exoyds:
I am not an Android developer either, so I won't be too much help with that side of things.

The phone is Xiaomi Mi Mix 3 5G. Android version is 9 PKQ1.190321.001 .
I'm on the Play Store Talk so that would be v8.0.9.

I have not disabled battery optimisation for the app itself, but that should not be an issue, as Signal,Telegram and WhatsApp handle screen turning off just fine.

[...]

Issue #2042, opened 2022-05-13
call ends when display is turned off


mahibi:
## Steps to reproduce
1. start call
2. switch off display

### Expected behaviour
call continues

### Actual behaviour
call hangup

we need a service for this. also see 
https://github.com/nextcloud/talk-android/issues/1688

https://developer.android.com/guide/components/services#Foreground

mahibi:
this is a feature that is asked for by some google play reviews...

AMajor-C:
Srsly ? Finally app will get feature that must have for a calling app ?

mahibi:
similar and maybe related: https://github.com/nextcloud/talk-android/issues/869

PH03N1X999:
Same happens to me also sometimes it gets muted sometimes get disconnected

mahibi:
Android 14 changes: https://developer.android.com/about/versions/14/changes/fgs-types-required

mahibi:
fixed by https://github.com/nextcloud/talk-android/pull/5465

Issue #1688, opened 2021-11-15
continue call when minimizing app and PIP is not available/enabled


mahibi:
currently it's only possible to continue a call in background when Picture in Picture is enabled.

There must be a solution to continue the call when PIP is not available/enabled.

Bryantdl7:
Telegram-FOSS has a call implementation that doesn't use google services, perhaps you guys could use that API: https://github.com/Telegram-FOSS-Team/Telegram-FOSS

AMajor-C:
PiP is a crutch not a solution and on my device it doesnt work (last alpha)

Byter3:
Actually it is solved for me now. Th call continues in the background properly when I open an other app. Its a bit finicky to get back in to the call. (Basically you have to open Talk then press on the call butzon, then the call will re-appear) anyways it's working for me now, thank you for the devs!
Device : Galaxy S21 Ultra
Talk app: 13.0.0

tareko:
Unfortunately this bug remains. The code in question is below and essentially kills the call if PiP is not available:

https://github.com/nextcloud/talk-android/blob/49991d70103f153bd10b25ac89169dc434ec7213/app/src/main/java/com/nextcloud/talk/activities/CallBaseActivity.java#L121-L130

I've been investigation how to fix this, as it is the last remaining piece needed along with Issue #869 and #5467 to bring Talk to equivalence with other talk apps on Android.

In essence, a few things have to happen here:

1. **Call lifecycle**
   - Prevent `CallActivity` from tearing down the ongoing call when it is backgrounded. Calls must continue without forcing a hangup, and the foreground notification should stay active.
   - Ensure WebRTC renderers (self view and PiP) are only initialized once and are safely reused when the UI is restored.
   - Avoid triggering `onCreate` when the user re-enters the call while it is already running. Resume the existing activity instance instead of launching a new one.

2. **Bring existing call to the front**
   - Add a helper (e.g., `CallActivity.show(...)`) that first searches for an existing `CallActivity` and moves it to the front.
   - Update all in-app entry points (`ChatActivity`, foreground service notification actions, incoming call notification taps, etc.) to use the helper so they reuse the running activity.
   - Make sure notification actions continue to support the "Leave meeting" broadcast flow.

4. **Call UI work and notification**
   - Guard `CallActivity`'s teardown logic with a flag so lifecycle-driven `onDestroy` calls do not hang up active calls. Only dispose local streams during an explicit hangup.
[...]

PR #5465, opened 2025-10-14
Place ongoing calls into foreground service

tareko:
As per issue #869 , there is a bug where bringing a call away from focus results in deactivating the microphone. This is the result of the call not being started with `FOREGROUND_SERVICE_TYPE_PHONE_CALL` . This fix was initially implemented by the [Jami client](https://git.jami.net/savoirfairelinux/jami-client-android/-/issues/1612), which gave me a good starting point.

This pull request does the following solves this bug by doing the following:

* Declare all call foreground service types
* Handle call foreground service permissions without dialer role
* Use microphone foreground service type for calls
* Start calls in a phone-call foreground service
* Add foreground service for ongoing calls

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -42,6 +42,8 @@
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CAMERA" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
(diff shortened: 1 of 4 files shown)


mahibi:
related: 
- https://github.com/nextcloud/talk-android/issues/2042
- https://github.com/nextcloud/talk-android/issues/1688

mahibi:
Thank you @tareko this is really great! :rocket: 
I always wanted to do this but never found the time and it's much easier than i imagined :+1: 

Just one tiny thing: Could you signoff the commits? See https://github.com/nextcloud/talk-android/pull/5465/checks?check_run_id=52712472066

tareko:
> Thank you @tareko this is really great! 🚀 I always wanted to do this but never found the time and it's much easier than i imagined 👍
> 
> Just one tiny thing: Could you signoff the commits? See https://github.com/nextcloud/talk-android/pull/5465/checks?check_run_id=52712472066

Thank you! Signed.

Issue #5466, opened 2025-10-14
Application fails when Picture-in-picture is activated during a call and then the lock screen is activated or the screen goes to sleep


tareko:
### Steps to reproduce

1. Start a talk call
2. Minimize, resulting in picture-in-picture
3. Hit power button or allow to go to sleep

### Expected behaviour

The call should continue

### Actual behaviour

The call dies

### Device brand and model

Pixel 6 pro / emulator

### Android version

16

### Nextcloud Talk app version

22.0.2 / 23.0.0 alpha 11

### Nextcloud server version

32.0.0

### Talk version

22.0.0

### Custom Signaling server configured

Yes (specify version in Additional Information)

### Custom TURN server configured

Yes

### Custom STUN server configured

Yes

### Android logs

```
2025-10-14 07:01:03.495 17876-19387 org.webrtc.Logging      com.nextcloud.talk2                  I  WebRtcAudioTrackExternal: Calling AudioTrack.stop...
2025-10-14 07:01:03.495 17876-19387 AudioTrack              com.nextcloud.talk2                  D  stop(69): called with 328800 frames delivered
2025-10-14 07:01:03.496   787-932   AppOps                  system_server                        E  attributionTag  not declared in manifest of com.nextcloud.talk2
2025-10-14 07:01:03.496   787-932   AppOps                  system_server                        E  attributionTag  not declared in manifest of com.nextcloud.talk2

2025-10-14 06:58:37.707 17876-18934 WebSocketInstance       com.nextcloud.talk2                  E  Error : WebSocket 66217947
                                                                                                    java.net.SocketException: Software caused connection abort
                                                                                                    	at java.net.SocketInputStream.socketRead0(Native Method)
                                                                                                    	at java.net.SocketInputStream.socketRead(SocketInputStream.java:118)
                                                                                                    	at java.net.SocketInputStream.read(SocketInputStream.java:173)
                                                                                                    	at java.net.SocketInputStream.read(SocketInputStream.java:143)
                                                                                                    	at org.conscrypt.ConscryptEngineSocket$SSLInputStream.readFromSocket(ConscryptEngineSocket.java:920)
                                                                                                    	at org.conscrypt.ConscryptEngineSocket$SSLInputStream.processDataFromSocket(ConscryptEngineSocket.java:884)
                                                                                                    	at org.conscrypt.ConscryptEngineSocket$SSLInputStream.readUntilDataAvailable(ConscryptEngineSocket.java:799)
[...]

PR #5467, opened 2025-10-14
Prevent Picture-in-Picture calls from ending when screen turns off

tareko:
This PR solves the bug in issue #5466 , and is related to bug #869 and PR #5465 . With these two PRs, calls work normally when in background.


mahibi:
really great :+1: thank you @tareko !

just like in the other PR: Could you signoff the commits?

PR #5517, opened 2025-10-31
[issue 5299] fix text color of additional hint explaining why the 'typing status' setting is disabled when no high-performance backend is used

### LINKED ISSUE #5299: wrong color for description of typing status (when using internal signaling)
description is hardly readable..:

Image

arkascha:
- fix #5299

Fixes the color of the additional hint that explains why the 'typing status' settings toggle is disabled in case no high-performance backend is used. 

Previously the existing hint's alpha channel was set to 'disabled'. However since the whole container already has a changed alpha channel due to being disabled this lead to a color barely readable against the background. 

This PR changes the hint into a text element and removes the change of the alpha channel. So the hint has the same appearance as the disabled text of the toggle switch. 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_talk-issue-5299_before | Screenshot_talk-issue-5299-after 

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
Thank you @arkascha

PR #5522, opened 2025-10-31
improving the background of reactions with added contrast

rapterjet2004:
### 🖼️ Screenshots Before

Light  | Dark | |
---|---|---
Screenshot 2025-12-02 at 7 57 13 AM | Screenshot 2025-12-02 at 7 57 59 AM | Outgoing
Screenshot 2025-12-02 at 8 00 19 AM | Screenshot 2025-12-02 at 7 59 25 AM | Incoming

### 🖼️ Screenshots After

Light  | Dark | |
---|---|---
Screenshot 2025-11-03 at 10 22 04 AM | Screenshot 2025-11-03 at 10 22 46 AM | Outgoing
Screenshot 2025-11-03 at 10 24 05 AM | Screenshot 2025-11-03 at 10 23 09 AM | Incoming

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/messages/Reaction.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/messages/Reaction.kt
@@ -11,8 +11,8 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import android.widget.TextView
-import com.nextcloud.talk.databinding.ReactionsInsideMessageBinding
 import com.nextcloud.talk.chat.data.model.ChatMessage
+import com.nextcloud.talk.databinding.ReactionsInsideMessageBinding
 import com.nextcloud.talk.ui.theme.ViewThemeUtils
 import com.nextcloud.talk.utils.DisplayUtils
 import com.vanniktech.emoji.EmojiTextView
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
@rapterjet2004 how about always having a border on the reaction items? That way we don't waste extra space for margins and paddings?

AndyScherzinger:
Would also be closer to the the web UI, no?

mahibi:
> @rapterjet2004 how about always having a border on the reaction items? That way we don't waste extra space for margins and paddings?

i agree always having a border on the items will be nicer (instead for the whole LinearLayout).

This is how it looks on web and we should do it similar on clients:

image

nimishavijay:
Good call on separating the reactions! Agreed that a border for all together looks a bit off though. We could go for a light background for each reaction, and for the ones the user has reacted with we can do primary colored border with a stronger background :)

AndyScherzinger:
@rapterjet2004 can you also add the "before" state, to have a better comparison if the color schema is "correct"? Thanks in advance

mahibi:
> @rapterjet2004 can you also add the "before" state, to have a better comparison if the color schema is "correct"? Thanks in advance

^ @rapterjet2004 can you update the screenshots & rebase?

AndyScherzinger:
Fine by me as an iteration to improve the display of reactions.

Mid-term for mobile I think we should pin them vertically centered on the bottom of the bubble like other messengers. Also but debatable of course only show the number if it is "more than 1". What do you think @nimishavijay

PR #5535, opened 2025-11-04
Message Pinning 📌

### LINKED ISSUE #5534: Message Pinning  📌
* Ref https://github.com/nextcloud/spreed/issues/3390
* Ref https://github.com/nextcloud/spreed/pull/16239
* Ref https://github.com/nextcloud/talk-ios/pull/2262

rapterjet2004:
### Note: The CI/CD check seems to be broken

- fixes #5534 

| Header | Header | Header |
|--------|--------|--------|
| Screenshot 2026-01-15 at 11 19 17 AM | Screenshot 2026-01-15 at 11 19 43 AM | Screenshot 2026-01-15 at 11 25 49 AM | 

### Features 
- tapping on a shared pinned item opens up the context chat window for that message
- tapping on the unpin icon unpins the message
- Message pinning/unpinning from message actions
- Message unpinning/dismissal/viewing from pinned menu actions
- Custom pinned message time limit
- Large pinned messages are scrollable
- Pinned messages are reactive to server side changes, state changes on other clients are reflected in the opened application in real time

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -34,6 +34,7 @@ import android.provider.MediaStore
 import android.provider.Settings
 import android.text.SpannableStringBuilder
 import android.text.TextUtils
+import android.text.format.DateFormat
 import android.util.Log
 import android.view.Gravity
 import android.view.Menu
@@ -59,11 +60,39 @@ import androidx.activity.result.contract.ActivityResultContracts.PickVisualMedia
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.view.ContextThemeWrapper
 import androidx.cardview.widget.CardView
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Menu
+import androidx.compose.material3.Divider
+import androidx.compose.material3.DropdownMenu
+import androidx.compose.material3.DropdownMenuItem
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.shadow
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.content.ContextCompat
 import androidx.core.content.FileProvider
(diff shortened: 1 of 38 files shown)


mahibi:
@rapterjet2004 as there is a chance that checks succeed now, can you rebase? Might be there will some merge conflicts..

mahibi (changes requested):
just rejecting for now to not merge before a potential v22.1.0 branchoff...

mahibi (changes requested):
- [x] Showing the details (the "Pinned until..." bubble) should be changed in my opinion.

- It has the same bubble color as the own message have. It's a bit confusing in my opinion.
- It does not need to be shown permanently.
- The bubble doesn't seem to catch tap events. When clicking on it, the underlying chat view is clicked (Noticed it when collapse/uncollapse was triggered in background)
- I think the buttons to hide and unpin are very tiny
- I would expect a menu when long clicking the pinned message (see iOS implementation: https://github.com/nextcloud/talk-ios/pull/2262)

- [x] Go to message is not implemented? The chat should scroll the the message when the pinned message is clicked.

- [x] Pinned messages should be added as a new type to "shared items"

- [x] In 1:1 conversations only possible to pin own message. But i can unpin a message from other participant. Please align how the behavior should be in 1:1 chats with other clients

- [x] If not being a moderator i see pinned messages and the "hide" button, but i can not hide. Please align with other clients

- [x] Somehow i got "Error in pinMessage: retrofit2.HttpException: HTTP 403"
but i can't reproduce...
I think i was not a moderator but saw the pin message option somehow.

sowjanyakch (changes requested):
- [ ] I tried to pin multiple messages, I see only the first message that is pinned. I would expect to see all the pinned messages in a scrollable view.
- [ ] The messages from the moderators cannot be pinned. I would expect to pin the messages from the moderators also.
- [ ] When a message is pinned, I see two buttons hide and unpin - the UI is little confusing and clicking on both hide and unpin buttons shows the same system message "You unpinned a message". Looks like hide button is also performing unpin action. 
- [ ] Pinned until .... is showing wrong time. Pin a message and set custom time pin until 6:00 PM and see the message and it shows wrong time.

mahibi:
looks much better now :+1: 

- [x] go to message should show the message centered. Else it is on top and overlayed by the pinned message box itself.

- [x] additionally it also makes sense to scroll to the message when just tapping the pinned message.

- [x] scroll to only works if the have the message load in the adapter. Maybe (until we have a better solution in the future..) we should show the message search results window if it was not found in adapter?

- [x] on server the string is "until unpin". on android "Pin indefinitely". please align.

- [x] the custom date dialog borders are not transparent but colored wrong:
grafik

- [x] in the attachments overview, clicking on a pin should not delete it but open the chat and scroll to it. At best would be an option to delete it (like in web)
grafik

mahibi (changes requested):
see comments

app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -1320,6 +1376,156 @@ class ChatActivity :
         }
     }
 
+    @Composable
+    private fun PinnedMessageView(message: ChatMessage) {
+        message.incoming = true
+
+        val pinnedBy = stringResource(R.string.pinned_by)
+
+        // FIXME this causes some problems with duplicate some times
```
mahibi: - [ ] todo before merging?..

rapterjet2004:
> - [ ] if possible, make it reactive to directly show updated pinned messages whenever they change (when system message is received). right now, when a change is done on web, it's just updated when reloading the chat or waiting for the next pull.
> 
> - [ ] on web, there is a pin shown inside the chat for the pinned messages. would be nice to also have this for mobiles.
> grafik

I don't think the first is possible without a significant refactoring, the second would require changing a lot of viewHolders, and would be better off implemented in another PR

sowjanyakch (changes requested):
- [x] Pin a very long message and the pinned message banner at top occupies more space leaving very less space for the chat messages. The message should be 1 or 2 lines long leaving more space for chat messages.

- [ ] Pin multiple messages and the pinned message banner shows the first pinned message. Only after a refresh latest pin message shows. Web UI is reactive and shows the latest pin message.

mahibi:
> > * [ ]  if possible, make it reactive to directly show updated pinned messages whenever they change (when system message is received). right now, when a change is done on web, it's just updated when reloading the chat or waiting for the next pull.
> > * [ ]  on web, there is a pin shown inside the chat for the pinned messages. would be nice to also have this for mobiles.
> > 
> > grafik
> 
> I don't think the first is possible without a significant refactoring, the second would require changing a lot of viewHolders, and would be better off implemented in another PR

okay lets postpone this until we migrated to Compose

mahibi:
i will merge it and further bugs can be solved with follow ups.

- [ ] issue with authors (for a follow up PR):

1. pin as user 1
2. in the same conversation, pin an other message as user 2

bug: message 2 is shown, but user1 is shown as author

PR #5567, opened 2025-11-21
filter out user only if userId and baseUrl are the same

tobiasKaminsky:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountDialogCompose.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountDialogCompose.kt
@@ -257,7 +257,9 @@ class ChooseAccountDialogCompose {
                     .size(40.dp)
                     .clip(CircleShape)
             )
-            Column(modifier = Modifier.padding(start = 12.dp).weight(1f)) {
+            Column(modifier = Modifier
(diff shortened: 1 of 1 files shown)


mahibi:
Thanks Tobias!
I did not test but this should solve it.

However do we need

`val userId: String?`

inside AccountItem at all?

It's confusing that the userId it saved twice so it can be accessed via:

`user.userId`
and
`user.user.userId`

?

The PR should fix the issue caused by
https://github.com/nextcloud/talk-android/blob/fd3a4aae26af7a1a1aa4865f3086f2515af62246/app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountDialogCompose.kt#L184
but it feels more like a workaround while the main issue is not solved: why can userId be null?

sowjanyakch:
tested and it works. Most of the checks fail here - so created a new PR #5588 which replaces this PR and also removes the userId from AccountItem. userId can only be null for guest accounts.

mahibi:
thanks again tobias!
will close in favor of https://github.com/nextcloud/talk-android/pull/5588

PR #5587, opened 2025-11-25
fix CI memory crashes

mahibi:
___

Update: 
The PR might not fix it completely as the crash still happens sometimes but it happens less frequently.
See comment https://github.com/nextcloud/talk-android/pull/5587#issuecomment-3584850506
___

Trying to fix memory crashes on CI.

See https://github.com/nextcloud/android-config/pull/317 that will do the changes that come from config repo.

This tries to fix errors like:

`Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)`

`FAILURE: Build failed with an exception`

---

Android Lint is part of the Android Gradle Plugin. Lint runs in its own worker JVM, not inside the main Gradle daemon, so `org.gradle.jvmargs` does not control Lint’s memory. Lint has its own heap settings.

`--no-daemon` prevents Gradle from keeping background JVMs alive, which ensures Lint’s worker processes are cleaned up correctly. This avoids memory leaks on CI runners.

Lint and SpotBugs are not fully compatible with Gradle's configuration cache. Therefore `--no-configuration-cache` must be used to avoid long running jobs and random cancellations.

### Good to know if it happens again in the future:

If Lint still runs out of memory, AGP exposes an internal setting `android.experimental.lint.heapSize`, which allows increasing the heap of the Lint worker JVM

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -36,7 +36,7 @@ jobs:
                 uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
 
             -   name: Run unit tests with coverage
-                run: ./gradlew testGplayDebugUnit
+                run: ./gradlew --no-daemon testGplayDebugUnit
 
             -   name: Upload test artifacts
                 if: ${{ failure() }}


mahibi:
Todo: check github gradle actions if this can improve cache and memory problems...

mahibi:
Okay so this PR might not fix the root cause, there were still fails with this. However it feels like it improved the situation somehow, need to have an eye on it..

The issue was reported for github runner actions!: https://github.com/actions/runner-images/issues/13263 

So the approach in https://github.com/nextcloud/talk-android/pull/5571 to change the concurrency settings was maybe not the worst idea..

PR #5588, opened 2025-11-25
Fix to show all accounts in AccountSwitcher

sowjanyakch:
refactor ChooseAccountDialogCompose

The userId is null only for guest accounts so userId parameter can be removed from AccountItem.

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
the more important change in this PR is that the baseUrl is also checked to avoid that accounts with the same name are not added to the account switcher list.
Will rename the title of the issue.

That's the change Tobias introduced in
https://github.com/nextcloud/talk-android/pull/5567

mahibi:
I think https://github.com/nextcloud/talk-android/pull/5227 introduced the bug with disappearing accounts recently (that why i won't list it as a bugfix in v22.1.0RC1).
So it may be confusing as the bug was just re-introduced with https://github.com/nextcloud/talk-android/pull/5227
It occurred before https://github.com/nextcloud/talk-android/pull/5227 and was already fixed.

PR #5606, opened 2025-12-02
Fixing NPE in startPlayback

### LINKED ISSUE #5601: NPE in ChatActivity.startPlayback
from gplay:

from 20.0.2 (and maybe older) up to 22.0.3

```
Exception java.lang.NullPointerException:
  at com.nextcloud.talk.chat.ChatActivity.startPlayback (ChatActivity.kt:1567)
  at com.nextcloud.talk.chat.ChatActivity.access$startPlayback (ChatActivity.kt:253)
  at com.nextcloud.talk.chat.ChatActivity$setUpWaveform$1$1.invokeSuspend (ChatActivity.kt:1552)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:34)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:104)
  at android.os.Handler.handleCallback (Handler.java:959)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at android.os.Looper.loopOnce (Looper.java:257)
  at android.os.Looper.loop (Looper.java:342)
  at android.app.ActivityThread.main (ActivityThread.java:9634)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:619)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:929)
```

rapterjet2004:
- fixes #5601 

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -1564,7 +1564,7 @@ class ChatActivity :
         message.isPlayingVoiceMessage = true
         adapter?.update(message)
 
-        var pos = adapter?.getMessagePositionById(message.id) ?: -1
+        var pos = adapter?.getMessagePositionById(message.id)?.minus(1) ?: -1
         do {
             if (pos < 0) break
             val nextItem = (adapter?.items?.get(pos)?.item) ?: break


app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
```diff
@@ -1564,20 +1564,21 @@ class ChatActivity :
         message.isPlayingVoiceMessage = true
         adapter?.update(message)
 
-        var pos = adapter?.getMessagePositionById(message.id)!! - 1
+        var pos = adapter?.getMessagePositionById(message.id) ?: -1
```
mahibi: I did not test, but this will change the behavior?

Before it was the message position -1. 
Now it is -1 if the message position is null, but if it's not null pos points to the current message so nextItem will also be the initial message..
This will result in the same message being queued again?

To have the same behavior it but null safe it should be

`adapter?.getMessagePositionById(message.id)?.minus(1) ?: -1`

mahibi:
@rapterjet2004 btw. i think commit messages like "dumb mistake" are not ideal because:
1. you are not dumb!!
2. in the future we or other devs won't understand the change without reading the comments on the github PR

For the future let's either phrase the commit message a bit more clearly or even better:
if there are no reasons against it, squash a commit like this with prior related commits.

Issue #1658, opened 2021-11-01
photo/video view: swipe left/right to show next/previous photo/video


jakobroehrl:

Smarshal21:
Would like to work on this 
@AndyScherzinger @mahibi are there any specific ways to work on this ?

AndyScherzinger:
None that I am aware of @Smarshal21 
Maybe @mahibi you have some opinion/requirements on it?

mahibi:
Thanks for your interest to work on this @Smarshal21 
Some first thoughts on this:

A good first step will be to familiarize with the package "shareditems" (SharedItemsRepositoryImpl, ...) to learn where the file list come from. The endpoint to use is https://nextcloud-talk.readthedocs.io/en/latest/chat/#list-items-of-type-shared-in-a-chat
As it will be about photos and videos only, the SharedItemType to use is "MEDIA".

Then FullScreenImageActivity/FullScreenMediaActivity have to be modified/replaced. Seems ViewPager2 is recommended to do the swiping. Not sure if it would make sense to combine rendering stuff from FullScreenImageActivity/FullScreenMediaActivity to have only one UI-view which can handle both photo/video.

Anyway it would make sense to create MVVM structure to have the data logic separated from rendering. 
Also, the viewModel could take care about downloading the next files in advance when the user swipes. See FileViewerUtils#openOrDownloadFile for download/cached handling.

Just let me know if you have any questions.

Smarshal21:
> Thanks for your interest to work on this @Smarshal21 Some first thoughts on this:
> 
> A good first step will be to familiarize with the package "shareditems" (SharedItemsRepositoryImpl, ...) to learn where the file list come from. The endpoint to use is https://nextcloud-talk.readthedocs.io/en/latest/chat/#list-items-of-type-shared-in-a-chat As it will be about photos and videos only, the SharedItemType to use is "MEDIA".
> 
> Then FullScreenImageActivity/FullScreenMediaActivity have to be modified/replaced. Seems ViewPager2 is recommended to do the swiping. Not sure if it would make sense to combine rendering stuff from FullScreenImageActivity/FullScreenMediaActivity to have only one UI-view which can handle both photo/video.
> 
> Anyway it would make sense to create MVVM structure to have the data logic separated from rendering. Also, the viewModel could take care about downloading the next files in advance when the user swipes. See FileViewerUtils#openOrDownloadFile for download/cached handling.
> 
> Just let me know if you have any questions.

[...]

PR #5622, opened 2025-12-05
Swipe Down support on Media files

### LINKED ISSUE #5609: Be able to close photos and videos by swipe gesture
When you want to close a photo or a video, you have to reach to the top of the app.

Swiping from top to bottom on the picture would close it, same as messager has it.

Going to the android opened apps, closing the window with the picture and opening the chat, work, but isn't convinient.

**Additional context**
None

rapterjet2004:
- fixes #5609 

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/fullscreenfile/FullScreenGestureListener.kt
+++ b/app/src/main/java/com/nextcloud/talk/fullscreenfile/FullScreenGestureListener.kt
@@ -1,0 +1,42 @@
+/*
+ * Nextcloud Talk - Android Client
+ *
+ * SPDX-FileCopyrightText: 2025 Julius Linus <juliuslinus1@gmail.com>
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+package com.nextcloud.talk.fullscreenfile
+
+import android.content.Context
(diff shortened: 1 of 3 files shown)


rapterjet2004:
After thinking about it, I think it might be better to refactor these full screen viewers to use fragments instead in a follow up PR. That way they can just use a Bottom Sheet, which IMO has a better UX. @AndyScherzinger thoughts?

AndyScherzinger:
Sounds good to me for a follow up but don't think that would be a priority right now. For example getting the sdk bumped to 36 to unblock library updated from being merged would be more important.

I am also fine with the current state since it is already an improvement but leave the review to @mahibi

mahibi (changes requested):
This PR breaks zooming and the toggle to show/hide appbar/navbar.

Yes it would be much nicer with an animation.
Also, both swiping up and down should close it?

Regarding the refactoring i suggest to take a step back and take into account more requirements, 
e.g.:
- https://github.com/nextcloud/talk-android/issues/1658

But i agree it's not a high priority and other things are more important.

mahibi:
closing in favor of https://github.com/nextcloud/talk-android/pull/5756

PR #5630, opened 2025-12-09
Refactoring search for correctness and efficiency and clarity

### LINKED ISSUE #5629: fetch open conversations only when searching
Right now, open conversations are fetched on every opening of the conversation list!
This is way too much and is stressing the server.

https://github.com/nextcloud/talk-android/blob/49991d70103f153bd10b25ac89169dc434ec7213/app/src/main/java/com/nextcloud/talk/conversationlist/viewmodels/ConversationsListViewModel.kt#L210

Then filtering is done locally when searching.

This must be changed:
- Open conversations should only be fetched when searching. 
- And only query when at least one search character is typed. 
   - (The conversation list should not list all open conversations, also not for an empty search. To get an overview of all open conversations, the "Join open conversations" should be used instead. This is  done and doesn't need changes)

rapterjet2004:
- fixes #5629 

The `getOpenConversations` API endpoint was being called every time the user gets a list of rooms, which was very inefficient, considering how often that occurs. This PR, moves that functionality to the querying logic, while also refactoring the querying logic for clarity and correctness. 

Now every time you search, it launches a Coroutine that gets the open conversations/users with structured concurrency, and processes their data in a thread safe way, before retrieving the messages async and adding them to the adapter when ready.

### 🚧 TODO

- [x] Figure out why users appear when tapping the search bar initially
- [x] Clicking load more doesn't remove the refresh indicator
- [x] Test this out with filtering button

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/contacts/ContactsViewModel.kt
+++ b/app/src/main/java/com/nextcloud/talk/contacts/ContactsViewModel.kt
@@ -12,8 +12,8 @@ import androidx.lifecycle.viewModelScope
 import com.nextcloud.talk.data.user.model.User
 import com.nextcloud.talk.models.json.autocomplete.AutocompleteUser
 import com.nextcloud.talk.models.json.conversations.Conversation
-import kotlinx.coroutines.CancellationException
 import com.nextcloud.talk.utils.database.user.CurrentUserProviderOld
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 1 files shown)


mahibi (approved):
I approve as it solves the issue. 👍 

But we should solve that in a different way in a follow up PR:

A better solution (but not for this PR) will be to move the fetching into the viewModel and there we should make use of `combine` to combine different data sources.
This should be done in general throughout the app and not only for this scenario.
Using async + awaitAll is not necessary then and also mutex handling is not needed.

So in the viewModel there should be code like:

```
viewModelScope.launch {
    combine(
        repository.fetchUsersFlow(filter),
        repository.fetchOpenConversationsFlow(filter)
    ) { users, conversations ->
        users + conversations
    }.collect { combined ->
        _searchResults.value = combined
    }
}
```

So in general, move more logic to view model while keeping the UI "dumb". The UI should only get the ready to use data for displaying..
We should make use of combine in many places which will also reduce boilerplate and complex code as we use too many flows that emit to the UI's.
Using combine will avoid to have too many flows.
Having too many flows makes code hard to read and it's hard to keep track where something is catched and what happens.

PR #5662, opened 2025-12-22
Fix - Bug in reactions

### LINKED ISSUE #5192: When a user reacts to a quoted message and then removes the reaction, the quote no longer appears in the message

sowjanyakch:
fix #5192 

- [X] Fix issue where removing a reaction from a quoted message causes the quote to disappear.
- [X] Also Prevent emoji count from showing as zero/negative value in offline mode (this case occurs when you add and delete a particular reaction multiple times when online).

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
@@ -687,7 +687,8 @@ class OfflineFirstChatRepository @Inject constructor(
             when (messageJson.systemMessageType) {
                 ChatMessage.SystemMessageType.REACTION,
                 ChatMessage.SystemMessageType.REACTION_REVOKED,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -687,7 +687,8 @@ class OfflineFirstChatRepository @Inject constructor(
             when (messageJson.systemMessageType) {
                 ChatMessage.SystemMessageType.REACTION,
                 ChatMessage.SystemMessageType.REACTION_REVOKED,
-                ChatMessage.SystemMessageType.REACTION_DELETED,
+                ChatMessage.SystemMessageType.REACTION_DELETED -> {
+                }
```
mahibi, line 691: this will completely ignore processing of reactions that were done on other clients

mahibi (changes requested):

PR #5716, opened 2026-01-16
Send data as POST param

p1gp1g:
I found out while debugging #5715 that the parameters where send via GET params, and they should be sent as a POST


mahibi (approved):
thank you.
both seem to work but @FieldMap with @FormUrlEncoded is definitely preferred 👍

PR #5762, opened 2026-01-27
Pinned Message Improvements

### LINKED ISSUE #5750: Pinned message improvements
- [x] Improve scolling in pinned message container

> - Text and icons change according to dark/light
    - vertical padding matches the input field (likely 16dp) for the container
    - padding in the container seems to be too large as well, should also be 16dp
    - furthermore I wouldn't use a fab, but rather have a 3-dot menu in the container at the top right/end

- [x] ref https://github.com/nextcloud/talk-android/issues/5717

> i will merge it and further bugs can be solved with follow ups.
> 
> * [ ]  issue with authors (for a follow up PR):
> 
> 1. pin as user 1
> 2. in the same conversation, pin an other message as user 2
> 
> bug: message 2 is shown, but user1 is shown as author

- [x] ref https://github.com/nextcloud/talk-android/pull/5535#issuecomment-3758823534

> For further polishing we might want to to make the menu headline bold else it currenrly looks like a menu item without an icon followed by a divider but the time info is not clickable. Else I think it could also be an option to put the time into the pinned message field also.

- [x] ref internal
> Thanks. I'm on v23 RC1. Setting custom dismissal time for pinned messages doesn't work. Once time is selected, tapping Ok does nothing.

rapterjet2004:
- fixes #5750

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/PinnedMessage.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/PinnedMessage.kt
@@ -2,6 +2,7 @@
  * Nextcloud Talk - Android Client
  *
  * SPDX-FileCopyrightText: 2026 Julius Linus <juliuslinus1@gmail.com>
+ * SPDX-FileCopyrightText: 2026 Andy Scherzinger <info@andy-scherzinger.de>
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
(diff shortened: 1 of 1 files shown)


mahibi:
i think the scrolling behavior in https://github.com/nextcloud/talk-android/pull/5807 is fine so this PR can be closed?

mahibi:
just tested this PR another time. the scroll behavior is buggy for me (moving too fast and reaches end too early).
I will close this PR as i think the scrolling without the scrollbar is fine. Hope that's okay for you @rapterjet2004

PR #5766, opened 2026-01-28
Improvise scheduled messages UI and solve edit bug

### LINKED ISSUE #5765: UI improvements to scheduled messages
see comment  https://github.com/nextcloud/talk-android/pull/5690#issuecomment-3811163717

sowjanyakch:
fix #5765 

some other followups:

- There is a bug with rescheduling/editing scheduled thread replies.
- Removed native android emoji library as it does not have inbuilt search and delete emoji options. I used vanniktech library for this PR and this will be improved in future.
- More testing should be done on rescheduling, editing all messages (thread, replies, text scheduled messages)

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2543,7 +2543,7 @@ class ChatActivity :
             (layoutManager as LinearLayoutManager).scrollToPositionWithOffset(position, 500)
         } else {
             Log.d(TAG, "message $messageId that should be scrolled to was not found (scrollToMessageWithId)")
-            startContextChatWindowForMessage(messageId, conversationThreadId.toString())
+            startContextChatWindowForMessage(messageId, conversationThreadId?.toString())
         }
     }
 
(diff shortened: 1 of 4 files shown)


mahibi:
> There is a bug with rescheduling/editing scheduled thread replies.

Can you describe this in more detail? What was the problem and what is fixed by the PR?

> Removed native android emoji library as it does not have inbuilt search and delete emoji options. I used vanniktech library for this PR and this will be improved in future.

I suggest to use the android emoji library as this should be the default in the future in my opinion. Also if there is no search.

sowjanyakch:
> Can you describe this in more detail? What was the problem and what is fixed by the PR?

STR:

1. Create a thread.
2. Schedule a reply to the thread by long press on the send button. You can also schedule a reply to the existing thread.
3. Try to edit/reschedule the scheduled thread reply. There is an error in both the cases.

Screenshot 2026-01-29 at 18 11 55

The PR will fix this issue.

You can also schedule a reply to a message. Also can edit, delete and reschedule them.

Screenshot 2026-01-29 at 18 20 14

All the cases are fully supported:  scheduled messages - reply to a thread, reply to a message, normal text message.

Screenshot 2026-01-29 at 18 25 02


> I suggest to use the android emoji library as this should be the default in the future in my opinion. Also if there is no search.

I will take a look again into it and see how we can improve this behavior.

PR #5769, opened 2026-01-29
Suppress ForegroundServiceType warning (because lint fails to recognize them?)

mahibi:
lint gives:

> "Missing foregroundServiceType attribute in manifest
To call Service.startForeground(), the <service> element of manifest file must have the foregroundServiceType attribute specified"

for the lines
CallForegroundService.kt:40
CallForegroundService.kt:42

However the AndroidManifest contains the foregroundServiceType for this class:

        <service
            android:name=".services.CallForegroundService"
            android:exported="false"
            android:foregroundServiceType="microphone|camera" />

For
`startForeground(NOTIFICATION_ID, notification, foregroundServiceType)`
i guess it's just a lint problem to not recognize that the service types are passed as variable.

For
`startForeground(NOTIFICATION_ID, notification)`
i guess lint is not able to take into account the android version check where it's not necessary to pass a service type.

So i will suppress the lint warnings.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/services/CallForegroundService.kt
+++ b/app/src/main/java/com/nextcloud/talk/services/CallForegroundService.kt
@@ -6,6 +6,7 @@
  */
 package com.nextcloud.talk.services
 
+import android.annotation.SuppressLint
 import android.app.Notification
 import android.app.PendingIntent
 import android.app.Service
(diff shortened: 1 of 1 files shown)


mahibi:
errors are "gone"

Issue #5551, opened 2025-11-13
speed up builds


mahibi:
Besides memory optimizations, we should finally aim to set

```
android.enableJetifier=false
android.nonTransitiveRClass=true
```

to speed up builds.

Todos:
get rid of old dependencies, especially chatkit, see https://github.com/nextcloud/talk-android/issues/1552

mahibi:
<img width="3072" height="556" alt="Image" src="https://github.com/user-attachments/assets/9059d92f-ba65-4446-8913-8032d863560e" />

On top of https://github.com/nextcloud/talk-android/pull/5831 , when i set 
`android.enableJetifier=false`
and also remove
`androidTestImplementation('com.android.support.test.espresso:espresso-intents:3.0.2')`

i get

```
AndroidStudioProjects/talk-android/app/src/main/java/com/nextcloud/talk/presenters/MentionAutocompletePresenter.java:48: Fehler: Kein Zugriff auf RecyclerView
public class MentionAutocompletePresenter extends RecyclerViewPresenter<Mention> implements FlexibleAdapter.OnItemClickListener {
       ^
  Klassendatei für android.support.v7.widget.RecyclerView nicht gefunden
```

Seems FlexibleAdapter is also relying on jetifier despite the check tool did not report it?
But FlexibleAdapter needs to be removed anyway..

PR #5831, opened 2026-02-10
remove spotlight lib for push to talk

mahibi:
This is a quite outdated lib and i don't think this feature is worth it to show the intro. Yes, it will become even a more hidden feature but i don't think many people want to use it at all.

Main reason why i want to remove is because it requires 
`Jetifier=true`
which might slow down builds.
ref https://github.com/nextcloud/talk-android/issues/5551

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -17,10 +17,10 @@ How to translate with transifex:
 -->
 
 <resources>
-    <!--
+<!--
   ~ Nextcloud Talk - Android Client
   ~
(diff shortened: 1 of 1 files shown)


mahibi:
thanks for removing the strings Andy. Just did the same at the very same time so i ended up with a "Invalid revision range" warning :D

AndyScherzinger:
> thanks for removing the strings Andy. Just did the same at the very same time so i ended up with a "Invalid revision range" warning :D

Sorry @mahibi bad/lacking communication on my end. I thought I save you the troubles and take care of getting it green and merged

AndyScherzinger:
^ so we speed up maintenance and final release builds as well

PR #5864, opened 2026-02-14
⚒️ Migrate geo location search to compose

AndyScherzinger:
* Migrated to composables incl. previews for dark/light/rtl
* Beware the migrated implementation respects the server colors (see blue vs. yellow shade (server color) in the screenshots below)

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before|after

### 🚧 TODO

- [x] review & test

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi (changes requested):
Searching in the results screen does not work at all. The whole searchterm remains visible when trying to delete chars with backspace.

Only the initial passed searchterm from the map view will show results.

AndyScherzinger:
Didn't test that ... I'll have a look @mahibi  👍

mahibi:
How i tested:
1. in map view, type Paris and click on search
2. in result view, try to delete paris
3. type hamburg

paris cant be deleted

https://github.com/user-attachments/assets/d90740d6-700a-4f85-ba3f-eca2bc525a9d

AndyScherzinger:
@mahibi rebased, fix amended and pushed:
search_location

The fix is
https://github.com/nextcloud/talk-android/pull/5864/changes#diff-b0cdc8e932b7d44dd3b140c3563097f3e840cdc99c0406e18a836eada2403ad6R63-R64

Other than that I replaced the query handling of just being a String to be a text field value to also ensure the cursor is initially put at the end, not at the beginning of the input field when opening the geo screen.

mahibi (approved):
works fine now 👍 

just now i see that LiveData is used. In general we prefer to stick with Flows.
It's fine for this PR though.

Thank you Andy!

AndyScherzinger:
> just now i see that LiveData is used. In general we prefer to stick with Flows.

Ah, I thought we didn't use Flow but LiveData or might have confused it with Files 🙈

AndyScherzinger:
@mahibi migrated via https://github.com/nextcloud/talk-android/pull/5864/commits/0c960ce5a4d320c81265e9a177290640e7594201

feel free to re-review and merged if fine (I tested and all worked fine for me)

app/src/main/java/com/nextcloud/talk/location/components/GeocodingScreen.kt
```diff
@@ -42,15 +42,15 @@ import fr.dudie.nominatim.model.Address
 
 @Composable
 fun GeocodingScreen(viewModel: GeoCodingViewModel, onBack: () -> Unit, onAddressSelected: (Address) -> Unit) {
-    val results by viewModel.getGeocodingResultsLiveData().observeAsState(viewModel.getGeocodingResults())
+    val results by viewModel.geocodingResults.collectAsState()
```
mahibi: Lets use collectAsStateWithLifecycle()

https://medium.com/@saykat-mir/differences-between-collectasstate-and-collectasstatewithlifecycle-in-jetpack-compose-c0a6ff282d03

mahibi (approved):

app/src/main/java/com/nextcloud/talk/location/components/GeocodingScreen.kt
mahibi: fine for me to merge afterwards

gradle/verification-metadata.xml
```diff
@@ -2652,6 +2652,14 @@
             <sha256 value="5cb52631cdc1e5f7632e0e4e3ca0bd602df17eda096636e77bb0b805bd5b8afa" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="androidx.compose.runtime" name="runtime-livedata" version="1.10.3">
```
Elsensee, line 2655: Probably leftover from before migrating livedata to flow?

AndyScherzinger: Yes, I can clean that up as well

AndyScherzinger: No, seems it needs to stay. I removed it but ran the meta data generation again which adds it again. So there must be some other place / arbitrary dependency adding it to the tree.

app/src/main/java/com/nextcloud/talk/location/components/GeocodingScreen.kt
AndyScherzinger: Done and merged all into a single commit 👍

AndyScherzinger: see https://github.com/nextcloud/talk-android/pull/5864/changes/561d6c7b2a61dc112d4c79a7f1e338027d045da5#diff-b0cdc8e932b7d44dd3b140c3563097f3e840cdc99c0406e18a836eada2403ad6R45

AndyScherzinger: Ah, found it, of course it is still added as long as I keep having the dependency in the gradle file... Sorry. Cleanup via #5882

gradle/verification-metadata.xml
AndyScherzinger: Ah, found it, of course it is still added as long as I keep having the dependency in the gradle file... Sorry. Cleanup via #5882

PR #5807, opened 2026-02-04
Simplifying pinned messages

rapterjet2004:
- ref #5774 

This PR will
- replace/move the button to open options
- modify pinned message design
- fix hardcoded strings
- add logic to differentiate in headline who pinned it (match web behavior)
- fix ANR crashes

### 🖼️ Screenshots

message from other...

| pinned by hisself | pinned by me |
|---|---|
|grafik|grafik|

message from myself

| pinned by hisself | pinned by myself |
|---|---|
|grafik|grafik|

### 🏁 Checklist

- [x] ⛑️ Tests (unit and/or integration) are included or not needed
- [x] 🔖 Capability is checked or not needed 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set
- [x] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #5865, opened 2026-02-14
📆 Show "Next meeting" in the chat

### LINKED ISSUE #5661: Show "Next meeting" in the chat
The Talk desktop client and the web interface show nicely the next meeting scheduled in the room:

Image

When I'm traveling it would be really nice to also see this information in the mobile client.

AndyScherzinger:
* **Please see https://github.com/nextcloud/talk-android/issues/5867 for discussion first**

---

* Resolves #5661
* Follows the implementation approach of the out-of-office feature
* Reuses the scheduled-call feature date-formatting logic
* Uses a generic copyright header, like we do in other non-mobile projects (can also switch to personal ones if you like)
* Add persistence to store info when dismissed with key o id+summary+time, so when any of it changes it will be displayed again

### 🖼️ Screenshots

🕸️ Web | 🏚️ Before | 🏡 After
---|---|---
web|before|Screenshot_20260220_112919

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/data/database/model/ConversationEntity.kt
+++ b/app/src/main/java/com/nextcloud/talk/data/database/model/ConversationEntity.kt
@@ -100,7 +100,9 @@ data class ConversationEntity(
     @ColumnInfo(name = "hasImportant") var hasImportant: Boolean = false,
     @ColumnInfo(name = "hiddenPinnedId") var hiddenPinnedId: Long? = null,
     @ColumnInfo(name = "lastPinnedId") var lastPinnedId: Long? = null,
+    // local-only field
     @ColumnInfo(name = "messageDraft") var messageDraft: MessageDraft? = MessageDraft(),
+    // local-only field
     @ColumnInfo(name = "hiddenUpcomingEvent") var hiddenUpcomingEvent: String? = null
     // missing/not needed: attendeeId
     // missing/not needed: attendeePin


jancborchardt (dismissed):
It looks different from the message pinning style as far as I can see? Or does it work well in conjunction with that?
https://github.com/nextcloud/talk-android/pull/5807

2 floating boxes would be fine. Ideal could be one floating box with a horizontal divider line, to save on space and visual complexity.

app/src/main/java/com/nextcloud/talk/models/json/upcomingEvents/UpcomingEvent.kt
```diff
@@ -0,0 +1,35 @@
...
+    @JsonField(name = ["calendarUri"])
+    var calendarUri: String,
+    @JsonField(name = ["start"])
+    var start: Long,
+    @JsonField(name = ["summary"])
+    var summary: String,
+    @JsonField(name = ["location"])
+    var location: String,
+    @JsonField(name = ["calendarAppUrl"])
+    var calendarAppUrl: String
```
mahibi: For now it fails with

```
2026-02-18 16:05:34.364 29876-29876 ChatActivity            com.nextcloud.talk2                  E  Error fetching upcoming events (Fix with AI)
                                                                                                    java.lang.NullPointerException: Parameter specified as non-null is null: method com.nextcloud.talk.models.json.upcomingEvents.UpcomingEvent.setCalendarAppUrl, parameter <set-?>
                                                                                                    	at com.nextcloud.talk.models.json.upcomingEvents.UpcomingEvent.setCalendarAppUrl(Unknown Source:2)
                                                                                                    	at com.nextcloud.talk.models.json.upcomingEvents.UpcomingEvent$$JsonObjectMapper.parseField(UpcomingEvent$$JsonObjectMapper.java:37)
                                                                                                    	at com.nextcloud.talk.models.json.upcomingEvents.UpcomingEvent$$JsonObjectMapper.parse(UpcomingEvent$$JsonObjectMapper.java:27)
```

please check which fields should be nullable.
see https://github.com/nextcloud/server/blob/a6d1fa1fa6f622f4036bdb9d5d9e2e83f99443f8/apps/dav/lib/CalDAV/UpcomingEvent.php

mahibi (changes requested):
Would be nice to already follow the suggestion from https://github.com/nextcloud/talk-android/issues/5867#issuecomment-3909147410 to make it collapsible?
Otherwise fine for me to do this in followup PRs.

app/src/main/java/com/nextcloud/talk/models/json/upcomingEvents/UpcomingEvent.kt
AndyScherzinger: Ouch, will do. I am seeing the definition while asking myself how they could be null no matter the spec. But oh well, seems they can 👍

All my events worked. Can you describe the event that fails for you, so I can test that case myself?

AndyScherzinger:
> Would be nice to already follow the suggestion from https://github.com/nextcloud/talk-android/issues/5867#issuecomment-3909147410 to make it collapsible?
Otherwise fine for me to do this in followup PRs.

@mahibi Yes, I will simply change it according to the discussion on the other item 👍 

Yet I am unsure if that means, hide it forever? hide until you open the chat again or hide it based on the date of the event you display?

WDYT @jancborchardt @nimishavijay ?

AndyScherzinger:
@mahibi implemented database layer change for dismissing and updated the design as discussed in the general design issue

app/src/main/java/com/nextcloud/talk/models/json/upcomingEvents/UpcomingEvent.kt
AndyScherzinger: Done, while I still find it puzzling that I could get an event that has no summary and no start date but is also an upcoming event 😆

jancborchardt (approved):
Looks good!

This should be added as follow-up :) https://github.com/nextcloud/talk-android/pull/5865#pullrequestreview-3824130663

app/src/main/java/com/nextcloud/talk/conversationlist/data/network/OfflineFirstConversationsRepository.kt
```diff
@@ -93,6 +93,8 @@ class OfflineFirstConversationsRepository @Inject constructor(
 
                     override fun onNext(model: ConversationModel) {
                         runBlocking {
+                            val existingEntity = dao.getConversationForUser(user.id!!, model.token).first()
+                            model.hiddenUpcomingEvent = existingEntity?.hiddenUpcomingEvent
```
mahibi, line 97: So this is the point where we begin to store local-only things in the same entity that is fed by the api.
That's okay of course. We would have to make sure this is handled everywhere where we use the endpoint which can easily be forgotten.

I'm just thinking if we should have a rule how we should do this in general.
What about to have this handled by mappers?
For example the mapper from ConversationModel to ConversationEntity could look like:


```
fun ConversationModel.asEntity(existing: ConversationEntity?): ConversationEntity {
    return ConversationEntity(
        internalId = internalId,
        accountId = accountId,
       ....

        // here, we could handle all "local-only" fields
        hiddenUpcomingEvent = existing?.hiddenUpcomingEvent ?: false
    )
}
```

With this, we won't accidentally forget to rescue specific columns from being overwritten. 

What do you think @AndyScherzinger ?

app/src/main/java/com/nextcloud/talk/data/database/model/ConversationEntity.kt
```diff
@@ -100,7 +100,8 @@ data class ConversationEntity(
     @ColumnInfo(name = "hasImportant") var hasImportant: Boolean = false,
     @ColumnInfo(name = "hiddenPinnedId") var hiddenPinnedId: Long? = null,
     @ColumnInfo(name = "lastPinnedId") var lastPinnedId: Long? = null,
-    @ColumnInfo(name = "messageDraft") var messageDraft: MessageDraft? = MessageDraft()
+    @ColumnInfo(name = "messageDraft") var messageDraft: MessageDraft? = MessageDraft(),
+    @ColumnInfo(name = "hiddenUpcomingEvent") var hiddenUpcomingEvent: String? = null
```
mahibi, line 106: makes sense to add a section as comment for all local-only fields?

app/src/main/java/com/nextcloud/talk/conversationlist/data/network/OfflineFirstConversationsRepository.kt
AndyScherzinger: In general it makes sense to discuss that, yes.

If I am not mistaken than `messageDraft` is local-only as well, no?

AndyScherzinger: See
https://github.com/nextcloud/talk-android/blob/feat/5661/nextMeeting/app/src/main/java/com/nextcloud/talk/models/domain/ConversationModel.kt#L70-L73

So it is field no3 😆

app/src/main/java/com/nextcloud/talk/data/database/model/ConversationEntity.kt
AndyScherzinger: DONE https://github.com/nextcloud/talk-android/pull/5865/commits/b42b7a7709e3979dfb22d33f024113b1537df41e

app/src/main/java/com/nextcloud/talk/conversationlist/data/network/OfflineFirstConversationsRepository.kt
mahibi: Okay i see, the idea wasn't quite fully developed yet :D
So fine for me to merge then

AndyScherzinger: Happy to discuss how to improve it 👍
The mapper does respect the fields, there just isn't a mapper with an entitled parameter input.

Issue #3248, opened 2023-08-14
Tighter integration on apps within Nextcloud ecosystem


tobiasKaminsky:
- have similar on Talk
- receive open intent with account to switch to passed account
- https://github.com/nextcloud/android/pull/11811/files#diff-1062717f7ca93b3caddbd46dae30394dcbc2f6aff9f81a5a1b8e113bbdcf9e7fR396

Ref: https://github.com/nextcloud/files-clients/issues/10

PR #5868, opened 2026-02-15
📚 Add ecosystem bar to account chooser dialog

AndyScherzinger:
* Resolves: #3248
* ...by adding a (composable) Ecosystem bar
* Displays the ecosystem as described in https://github.com/nextcloud/talk-android/issues/3248#issuecomment-2090687271
* Adds a branding switcher, like Notes and Files clients (see setup.xml at https://github.com/nextcloud/talk-android/compare/feat/3248/ecosystem?expand=1#diff-6f6e1776d754790ae82df99359aa6c05672719665fe748686624c4769cc094deR27)
* Adds a settings option to turn it off (same as Notes and Files clients)
* Hides the Ecosystem bar for non-portrait layouts modes due to lack of space and on branded clients in general
* RTL/dark/light preview for the composable added
* Resolves some detekt issues to turn CI green

### 🖼️ Screenshots

⚙️ Settings | 🏚️ Before/Off | 🏡 After
---|---|---
settings|before|after

### 🚧 TODO

- [x] review and merge
- [ ] ping @tobiasKaminsky afterwards for Brander implementation (equivalent to Notes client)

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -111,6 +111,11 @@
             </intent-filter>
 
             <intent-filter>
+                <action android:name="com.nextcloud.intent.OPEN_ECOSYSTEM_APP" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
+
+            <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
 


mahibi:
Design wise it's fine.
Just the actual opening of the other apps fail for me. 

I  get

"Cannot open app store. Please install manually."

`exception launching app [com.nextcloud.client, com.nextcloud.android.beta, com.nextcloud.android.qa]: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=com.nextcloud.intent.OPEN_ECOSYSTEM_APP flg=0x30000000 xflg=0x4 pkg=com.nextcloud.client (has extras) }`

but i guess that's something to be improved/fixed on android-commons..

AndyScherzinger:
@alperozturk96 to the rescue 😆 

> "Cannot open app store. Please install manually."

I guess cause it is a de-googlified phone?

```
exception launching app [com.nextcloud.client, com.nextcloud.android.beta, com.nextcloud.android.qa]: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=com.nextcloud.intent.OPEN_ECOSYSTEM_APP flg=0x30000000 xflg=0x4 pkg=com.nextcloud.client (has extras) }
```

Might be the need for the latest version of the app @alperozturk96 upcoming 3.36 ?

alperozturk96:
> Design wise it's fine. Just the actual opening of the other apps fail for me.
> 
> I get
> 
> "Cannot open app store. Please install manually."
> 
> `exception launching app [com.nextcloud.client, com.nextcloud.android.beta, com.nextcloud.android.qa]: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=com.nextcloud.intent.OPEN_ECOSYSTEM_APP flg=0x30000000 xflg=0x4 pkg=com.nextcloud.client (has extras) }`
> 
> but i guess that's something to be improved/fixed on android-commons..

What is the exact package name of the app you tried to open?

It should match one of the expected package names, as these should cover all variants across our build configurations.

```
enum class EcosystemApp(
    val packageNames: List<String>
) {
    FILES(
        listOf(
            "com.nextcloud.client", // generic, gplay, huawei
            "com.nextcloud.android.beta", // versionDev
            "com.nextcloud.android.qa" // qa
        )
    ),
    NOTES(
        listOf(
            "it.niedermann.owncloud.notes", // play, fdroid
            "it.niedermann.owncloud.notes.dev", // dev
            "it.niedermann.owncloud.notes.qa" // qa
        )
    ),
    TALK(
        listOf(
            "com.nextcloud.talk2", // generic, gplay
            "com.nextcloud.talk2.qa" // qa
        )
    )
}
```

----

Also, if they are using a branded client, the package name might be different and may need to be registered separately.

For example, the default Files client uses:
`com.nextcloud.client`

But a branded version could be something like:
`com.examplecloud.client`

@tobiasKaminsky @AndyScherzinger  right and we must support it?

AndyScherzinger:
> @tobiasKaminsky @AndyScherzinger right and we must support it?

Ecosystem bar is non-branded only on all our clients

> What is the exact package name of the app you tried to open?

It is fully using the common implementation calling https://github.com/nextcloud/android-common/blob/main/core/src/main/java/com/nextcloud/android/common/core/utils/ecosystem/EcosystemManager.kt#L62

So the issue @mahibi ran into is the Files client not supporting https://github.com/nextcloud/android-common/blob/main/core/src/main/java/com/nextcloud/android/common/core/utils/ecosystem/EcosystemManager.kt#L37 which I guess is due to the fact that the Files app is too old and than it breaks, right? So it will only work with 3.36+?

mahibi:
> I guess cause it is a de-googlified phone?

No, not de-googled

> So the issue @mahibi ran into is the Files client not supporting https://github.com/nextcloud/android-common/blob/main/core/src/main/java/com/nextcloud/android/common/core/utils/ecosystem/EcosystemManager.kt#L37 which I guess is due to the fact that the Files app is too old and than it breaks, right? So it will only work with 3.36+?

yes, tested with files 3.35.2

PR #5872, opened 2026-02-16
show diagnosis screen in english

### LINKED ISSUE #5823: Show diagnosis screen only in english
As the diagnosis screen is mostly for developers and the report is mostly used to be sent to Nextcloud devs, i suggest to not use translations for this screen, it should be always english.

sowjanyakch:
fix #5823 

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/res/layout/activity_settings.xml
+++ b/app/src/main/res/layout/activity_settings.xml
@@ -324,7 +324,7 @@
                     <com.google.android.material.textview.MaterialTextView
                         android:layout_width="wrap_content"
                         android:layout_height="wrap_content"
-                        android:text="@string/nc_diagnose_gplay_available_no" />
+                        android:text="@string/gplay_available_no" />
 
                 </LinearLayout>
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -188,7 +188,7 @@ How to translate with transifex:
 
     <string name="nc_settings_notifications_granted">Notifications are granted</string>
     <string name="nc_settings_notifications_declined">Notifications are declined</string>
-    <string name="play_available_no">Google Play services are not available.Notifications are not supported</string>
+    <string name="gplay_available_no">Google Play services are not available.Notifications are not supported</string>
 
     <!-- Diagnose -->
     <string name="nc_notifications_troubleshooting_dialog_title">Notification troubleshooting</string>


AndyScherzinger:
@sowjanyakch translations vanish automatically over time if no stable branch ships the value string anymore, until then transifex will always bring them back (!). So you can ignore string values in value-* folders

sowjanyakch:
> @sowjanyakch translations vanish automatically over time if no stable branch ships the value string anymore, until then transifex will always bring them back (!). So you can ignore string values in value-* folders

@AndyScherzinger - the problem is eventhough I mark string as not translatable (with  _translatable = "false"_), android still loads localised string because same string is already available in different languages. Without deleting strings, it is not possible to test the output. Happy to get some suggestions on how to improve the implementation.

Another reason to delete strings is increase in the warnings. 

_The resource string "nc_diagnose_signaling_mode_title" has been marked as translatable="false" elsewhere (usually in the values folder), but is translated to "sv" (Swedish) here_

AndyScherzinger:
>  Without deleting strings, it is not possible to test the output. Happy to get some suggestions on how to improve the implementation.

@sowjanyakch fine to leave them in in this case 👍 Given the files will get overwritten anyways after the first transifex sync, but that doesn't matter since it is merged.

app/src/main/java/com/nextcloud/talk/settings/SettingsActivity.kt
```diff
@@ -332,13 +332,13 @@ class SettingsActivity :
...
                 binding.batteryOptimizationIgnored.text =
-                    resources!!.getString(R.string.nc_diagnose_battery_optimization_ignored)
+                    resources!!.getString(R.string.battery_optimization_ignored)
                 binding.batteryOptimizationIgnored.setTextColor(
                     resources.getColor(R.color.high_emphasis_text, null)
                 )
             } else {
                 binding.batteryOptimizationIgnored.text =
-                    resources!!.getString(R.string.nc_diagnose_battery_optimization_not_ignored)
+                    resources!!.getString(R.string.battery_optimization_not_ignored)
```
mahibi, line 341: Please make sure that strings that are shown in other screens than the diagnosis screen are still translated.

If there is no fancy solution for this, just add them twice in strings.xml, one time with and one time without translation.

sowjanyakch: yes, I followed the same process. I created additional strings for settings screen and also modified the keys as the prefix nc_diagnose for settings screen creates confusion. The newly added strings for settings screen are not available in other languages yet(needs to be translated to other languages by transifex so we can test the settings screen).

PR #5891, opened 2026-02-21
Moving Background Blur to GPU

rapterjet2004:
### 🖼️ Screenshots

Before | After
---|---
Screenshot 2026-02-02 at 10 51 34 AM | Screenshot 2026-02-02 at 10 28 43 AM 

Demo (foreground original / background blurred)| Mask (foreground white / background black)
---|---
Screenshot 2026-02-20 at 7 57 46 PM | Screenshot 2026-02-20 at 8 04 26 PM

This PR addresses an issue with the previous implementation of Background Blurring. It used too much memory space, as openCV contains huge builds for every possible memory architecture supported on Android. This leads to duplicate files, large amounts of unnecessary code, and a larger download size. The MediaPipe library and the .tflite file used for foreground-background detection was trivial in comparison. 

This is fixed by moving the entire blurring operation to openGL ES, a graphics framework that is built into Android, cutting out 100 MB from the build size while also increasing performance. This is done by the implementing Gaussian Blur algorithm from scratch in GLSL, the openGL Shader language.  

This adds
- `BackgroundBlurGPUProcessor` a file that handles the openGL logic, and acts as an abstraction over the inner workings of the image processing. OpenGL's logic is ... very unclear. I tried my best to document as much as possible, including any linear algebra that I had to perform.
- `background_blur_vertex` the required vertex shader, needed to identify the position of each pixel on the screen, before the geometry, rasterization, and fragment shader steps of the openGL pipeline.
- `gaussian_blur_frag_shader` the Gaussian Blur implementation, this is what actually performs the logic on the image pixels. The code is also somewhat unclear if you're not familiar with shaders or gaussian blur. It helps to read the technical article by Intel that I commented.
- `seg_mask_frag_shader` the masking operation implementation, very simple, just checking if the pixel should be blurred or retain it's original unblurred value
- `libyuv_android` a light weight helper lib for converting between image formats without the overhead of OpenCV

Copied over from old PR
- `BackgroundBlurFrameProcessor` A WebRTC `VideoProcessor` object, this handles the video frame conversions and processing. Holds references to the `ImageSegmenterHelper` and the `BackgrounBlurGPUProcessor`. It now holds it's own dedicated Thread and Handler for running OpenGL's GPU operations. It also takes in a reference to the `surfaceTextureHelper` from `CallActivity` to send the `VideoFrame` objects to the WebRTC `VideoSink` on the appropriate capture thread. Failing to do both of these causes a crash.
- `BlurBackgroundViewModel` 
- `ImageSegmenterHelper` uses MediaPipe to perform the segmentation mask (foreground / background separation). Really just a wrapper over the MediaPipe API
- `background_replace_24px`
- 2 new strings

[See my blog post for more details](https://rapterjet2004.github.io/Blog/2-background-blur.html)

### 🚧 TODO

- [x] rebase + solve merge conflicts
- [x] analyze for memory leaks
- [x] linter

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -296,6 +296,10 @@ dependencies {
     implementation("io.noties.markwon:ext-tasklist:$markwonVersion")
     implementation("io.noties.markwon:ext-tables:$markwonVersion")
 
+    // Computer Vision - for background effects during video calls
+    implementation("com.google.mediapipe:tasks-vision:0.10.26")
+    implementation("io.github.crow-misia.libyuv:libyuv-android:0.43.2")
+
     // Avatar picker
     implementation("com.github.nextcloud-deps:ImagePicker:2.1.0.2")
     // Override avatar picker's internal crop lib for better edge-to-edge support


REUSE.toml
```diff
@@ -34,3 +34,9 @@ path = "app/src/main/res/raw/next_voice_message_doodle.ogg"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2024 Paciosoft"
 SPDX-License-Identifier = "CC-BY-4.0"
+
+[[annotations]]
+path = "app/src/main/assets/selfie_segmenter.tflite"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "Apache License Version 2.0, January 2004 http://www.apache.org/licenses/"
```
AndyScherzinger: This would be the copyright holder, but ships a lengthier license string, so I guess this value needs to be updated

mahibi:
Awesome work @rapterjet2004 :+1:

PR #5893, opened 2026-02-22
🗺️ Migrate location picker to composables

AndyScherzinger:
* with custom zoom controls and 
* dark mode support (discuss if it should be kept)
* visual revamp with search bar on top, current location moved to bottom and vertical zoom controls moved to the side on portrait and horizontal on landscape and also replaced the red pin using a different, less aggressive shade of red and a black outline

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before_light|Screenshot_20260223_175200
before_dark|Screenshot_20260223_175230
before_dark_landscape|Screenshot_20260223_175239

### 🚧 TODO

- [x] add screenshots
- [x] max search bar max-width-ed to 720dp as per M3 guide
- [x] review design
- [ ] review code

### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
The dark mode of the map might be debatable @jancborchardt since it interferes with the search bar. I could make that the same color as the bottom element (surface container highest instead or high) while that would not be according to M3 spec, but not an issue per se.

jancborchardt (approved):
Looks good.

In landscape mode there could be some space saved by not making the "Share location" box full width but keeping some space for the zoom and location buttons.

If it's according to spec as is, that's fine though.

AndyScherzinger:
> In landscape mode there could be some space saved by not making the "Share location" box full width but keeping some space for the zoom and location buttons.

@jancborchardt Added with the last push 👍

mahibi:
What about outlined buttons/cards to no interfere with the map?
Or/and add a shadow. 
The floating button currently has a shadow by default, the others not.

some optional gimmick for later follow ups maybe:
would be nice if the pin is lifted when the map is dragged, and dropped when dragging ends. And add a shadow where the pin sticks. This is how it's done on signal.

Also btw: it seems that geocoding search is broken, also on master. Maybe last minute changes from https://github.com/nextcloud/talk-android/pull/5864 ?

AndyScherzinger:
@mahibi outlined buttons for which ones? Zoom. controls or the fab?

The geo search results I also detected but didn't have the time yet to look into it again yet, let me find someone time for that 👍

PR #5895, opened 2026-02-23
Pinned messages are now opened properly in threads

### LINKED ISSUE #5860: pinned message in thread behavior
v23.0.0RC3

1. pin a message in a conversation
2. create a thread in the conversation
3. go to the thread (sidenote: pinned message still appears in the thread. same on web. tbd if this is what we want.)
4. click on the pinned message

## actual behavior:
search result window opens (for the thread) but message can't be found there

## expected:
at best, the normal conversation opens and is scrolled to message. alternatively the search result window for the normal conversation

Please also test the other way round: pin in thread & then click on it in normal conversation.

rapterjet2004:
- fixes #5860

This uses the correct context to launch the message viewer. Also since thread messages are just regular messages with a thread id, If you pin a message in a thread, and try to access it outside the thread, it just opens it up in the chat context view.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2600,11 +2600,10 @@ class ChatActivity :
         }
         if (position != null && position >= 0) {
             val layoutManager = binding.messagesListView.layoutManager
-            // FIXME Not a perfect offset, but works to clear the pinned message view, try and find a better solution
             (layoutManager as LinearLayoutManager).scrollToPositionWithOffset(position, 500)
         } else {
             Log.d(TAG, "message $messageId that should be scrolled to was not found (scrollToMessageWithId)")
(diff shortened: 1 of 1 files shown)


sowjanyakch:
@rapterjet2004 - tested this PR and works as expected. 

One small suggestion: when a pinned thread message is opened from the main conversation, the search results view points to the thread pinned message and from the UI it is hard to tell that it is a thread message (looks more like a reply to a message in this case). We could show a thread icon and title with the thread message, so it is obvious that the search result is a thread message. wdyt?

rapterjet2004:
> One small suggestion: when a pinned thread message is opened from the main conversation, the search results view points to the thread pinned message and from the UI it is hard to tell that it is a thread message (looks more like a reply to a message in this case). We could show a thread icon and title with the thread message, so it is obvious that the search result is a thread message. wdyt?

Might be better off implemented in another PR, as I think deliberating over the thread feature's UI is outside the scope of issue #5895

PR #5909, opened 2026-02-26
🖼️ Show images in message notifications

### LINKED ISSUE #2300: The ability to preview media such as images in the android notification tray.
The notifications are looking great with the latest updates however I think its missing one feature to polish it off.

The ability to preview media such as images in those notifications which is in the dev documentation:

https://developer.android.com/guide/topics/ui/notifiers/notifications.

AndyScherzinger:
* Resolves #2300
* Currently a simple check if the message ships an image info and if so, display it.

### 🖼️ Screenshots

Home | Collapsed | Expanded
---|---|---
home|collapsed|expanded

### 🚧 TODO

- [x] discuss hardening

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
@@ -484,7 +484,7 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
             pushMessage.subject = ncNotification.subject.orEmpty()
         }
 
-        checkAndExtractImagePreviewData (ncNotification)
+        checkAndExtractImagePreviewData(ncNotification)
     }
 
     private fun checkAndExtractImagePreviewData(
@@ -686,11 +686,13 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
     private fun styleImageNotification(notificationBuilder: NotificationCompat.Builder) {
         val bitmap = loadImageBitmapSync(imagePreviewUrl!!)
         if (bitmap != null) {
-            notificationBuilder.setStyle(
-                NotificationCompat.BigPictureStyle()
-                    .bigPicture(bitmap)
-                    .bigLargeIcon(null as Bitmap?)
-            )
+            notificationBuilder
+                .setLargeIcon(bitmap)
+                .setStyle(
+                    NotificationCompat.BigPictureStyle()
+                        .bigPicture(bitmap)
+                        .bigLargeIcon(null as Bitmap?)
+                )
         }
     }
 


app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
```diff
@@ -477,6 +483,36 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
...
+        imageMimeType = null
+        val msgParams = notification.messageRichParameters
+        if (msgParams != null) {
+            for ((_, param) in msgParams) {
+                if (processMessageParameter(param)) break
+            }
+        }
+    }
+
+    private fun processMessageParameter(param: HashMap<String?, String?>): Boolean {
```
mahibi: make this a nested function inside `checkAndProcessImagePreviewData`?

And maybe rename? `processMessageParameter` does not really unveil what it's doing. 
Maybe call the outer function `checkAndextractImagePreviewData` and the nested function `extractImagePreviewData`?

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
```diff
@@ -643,6 +683,32 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
...
+        val request = ImageRequest.Builder(context!!)
+            .data(imageUrl)
+            .allowHardware(false)
+            .addHeader("Authorization", credentials)
+            .target(
+                onSuccess = { result -> bitmap = (result as BitmapDrawable).bitmap },
+                onError = { Log.w(TAG, "Failed to load notification image: $imageUrl") }
+            )
+            .build()
+        context!!.imageLoader.executeBlocking(request)
```
mahibi, line 710: as we are in a worker, use of blocking should be fine.

Further improvements in followup PRs could be:

make it a suspend function instead to use blocking
-> for this, best is to migrate the Worker to CoroutineWorker

As updating notifications by using the notificationId is quite easy, we could think about to show a notification with text only first, and whenever image data is loaded, simply update the notification.

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
```diff
@@ -477,6 +483,36 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
...
+        }
+    }
+
+    private fun processMessageParameter(param: HashMap<String?, String?>): Boolean {
+        if (param["type"] == "file") {
+            val mimetype = param["mimetype"].orEmpty()
+            val fileId = param["id"]
+            if (mimetype.startsWith("image/") && fileId != null) {
+                val baseUrl = signatureVerification.user!!.baseUrl!!
+                val px = context!!.resources.getDimensionPixelSize(R.dimen.maximum_file_preview_size)
```
mahibi: > I remember thinking we only have 2 preview sizes, so using the optimal notification size might be nonesense (or I am wrong with this assumption).

The two different preview sizes were for the avatar endpoint, i guess you thought about this?

The `/index.php/core/preview` enpoint that is used here can act to any sizes that are requested. Though i think 
`<dimen name="maximum_file_preview_size">192dp</dimen>`
does not make sense.

I guess measuring the device width with

`context!!.resources.displayMetrics.widthPixels`

and then modify the

`getUrlForFilePreviewWithFileId`

function to handle only the width (height seems to be set accordingly by image ratio from server) might be a good solution.

Unrelated to this PR he current implementation of `getUrlForFilePreviewWithFileId` which also sets y=x is questionable in my opinion.

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
AndyScherzinger: We could think about that, yes, while the notification with the image is a different "type", so we would need to replace the whole notification, but yeah, could do so via the ID.

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
AndyScherzinger: > The two different preview sizes were for the avatar endpoint, i guess you thought about this?

Right, I might have confused it with that, while I thought this had been unsified to 2 sizes for previews at some point by @CarlSchwan but I might be wrong here again / as well.

> I guess measuring the device width with

Yeah, we could do that 👍 

> and then modify the `getUrlForFilePreviewWithFileId`

Sure 👍  

> Unrelated to this PR he current implementation of getUrlForFilePreviewWithFileId which also sets y=x is questionable in my opinion.

My guess for  sets y=x could only be that we did this for the square previews in the file browser when choosing files form Nextcloud to share in a room?

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
AndyScherzinger: Did the rename. I originally extracted it to not have a too deep hierarchy with detekt (or ktlint?) won't like...

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
AndyScherzinger: So currently did everything except modifying `getUrlForFilePreviewWithFileId` in any way (also used in chats and shared item views).

mahibi (approved):
works fine 👍

PR #5913, opened 2026-03-01
⚒️ Migrate open conversations list to composables

AndyScherzinger:
* Migrated to composables and from LiveData to Flow
* Added search term highlighting like the general conversation list on search does
* (semi)transparent bars system/navigation
* Avatar now with typical 48dp size
* Search field is now a M3 searchbar as defined in https://m3.material.io/components/search/overview

Automatically improves the overall theming (surface colored!). Also the sub-line text size is now in line with other screens (medium sized) that are composables already.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
list_old|list_new
search_none_old|Screenshot_20260302_164012
search_comm_old|Screenshot_20260302_164023

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/openconversations/OpenConversationsScreen.kt
+++ b/app/src/main/java/com/nextcloud/talk/openconversations/OpenConversationsScreen.kt
@@ -6,6 +6,8 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+@file:Suppress("TooManyFunctions")
+
 package com.nextcloud.talk.openconversations
 
 import android.content.res.Configuration
@@ -19,11 +21,11 @@ import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
 import androidx.compose.foundation.layout.asPaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.imePadding
 import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.only
-import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.safeDrawing
 import androidx.compose.foundation.layout.size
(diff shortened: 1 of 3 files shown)


sowjanyakch:
LGTM. One minor improvement - search bar should be themed.

AndyScherzinger:
> LGTM. One minor improvement - search bar should be themed.

Thanks for the review @sowjanyakch - updated the screenshots accordingly and replaced the textfield with an actual searchbar via https://github.com/nextcloud/talk-android/pull/5913/commits/2a99a30641c6779743163660d94cc28f8b307218

app/src/main/java/com/nextcloud/talk/openconversations/OpenConversationsScreen.kt
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+@Composable
+private fun EmptyConversationsView(modifier: Modifier = Modifier) {
+    Column(
+        modifier = modifier.padding(32.dp),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Icon(
+            imageVector = ImageVector.vectorResource(R.drawable.baseline_info_24),
```
mahibi, line 410: it was already in place before this PR, but ow that i'm seeing it an exclamation mark is not the right choice here as it feels like some error?
Maybe an info icon like 

https://fonts.google.com/icons?selected=Material+Symbols+Outlined:info:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=info&icon.size=24&icon.color=%23222222

is better here?

Also, the strings

    <string name="nc_no_open_conversations_headline">No open conversations</string>
    <string name="nc_no_open_conversations_text">No open conversations that you can join.\nEither there are no open conversations or you already joined all of them.</string>

are shown when the search does not have results. Maybe there should be a different info like "No results for your search"?

AndyScherzinger: Icon is fixed via https://github.com/nextcloud/talk-android/pull/5913/commits/95a69b67158aac49d84bb186ad99dc563d02440a which was also the semantically wrong one for the invitation activity.
The different text if client-side filtering leads to an empty result has been added via https://github.com/nextcloud/talk-android/pull/5913/commits/310074455efdf8f5192f54c461fdfc453769e322

PR #5914, opened 2026-03-01
⚒️ Migrate account chooser for sharing from flex adapter to composables

AndyScherzinger:
For an easier review I kept the commits split: migrate away from flex adapter, migrate to composables

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
dialog_old|dialog_new

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set


app/src/main/java/com/nextcloud/talk/ui/dialog/viewmodels/ChooseAccountShareToViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import javax.inject.Inject
+
+class ChooseAccountShareToViewModel @Inject constructor(private val userManager: UserManager) : ViewModel() {
+
+    val currentUser: User? = userManager.currentUser.blockingGet()
+
+    sealed interface ViewState
```
alperozturk96: I recommend using a more specific name, `ChooseAccountShareToViewState`, so it doesn’t interfere with other view states in the project, since `ViewState` is quite generic.

Also, ideally, `ChooseAccountShareToViewModel` should not belong to `com.nextcloud.talk.ui.dialog.viewmodels`; however, this is out of scope for this PR.

Ideal structure:

* `com.nextcloud.talk.ui.chooseAccount` → `ChooseAccountShareToViewModel` and `ChooseAccountShareToView`
* `com.nextcloud.talk.ui.chooseAccount.model` → `ChooseAccountShareToViewState`

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private val _viewState: MutableStateFlow<ViewState> = MutableStateFlow(LoadUsersStartState)
+    val viewState: StateFlow<ViewState> = _viewState.asStateFlow()
+
+    private val disposables = CompositeDisposable()
+
+    fun loadUsers() {
+        _viewState.value = LoadUsersStartState
+        disposables.add(
+            userManager.users
```
alperozturk96, line 41: Does it have to use `io.reactivex`? If migration to `suspend` function is not require too much additional time I would do that.

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(
```
alperozturk96, line 44: If you need handle specific case for `SwitchUserErrorState` you can use `.onError`

ref: https://github.com/ReactiveX/RxJava/wiki/Error-Handling-Operators

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+                        Log.e(TAG, "Error switching user", e)
+                        _viewState.value = SwitchUserErrorState
+                    }
+                )
+        )
+    }
+
+    override fun onCleared() {
+        super.onCleared()
+        disposables.dispose()
```
alperozturk96, line 76: Shouldn't be in different order?

```
disposables.dispose()
super.onCleared()
```

app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountShareToDialogCompose.kt
```diff
@@ -0,0 +1,209 @@
...
+    onCurrentUserClick: () -> Unit,
+    onOtherUserClick: (User) -> Unit
+) {
+    Surface(
+        color = MaterialTheme.colorScheme.surfaceContainerHigh,
+        modifier = Modifier.fillMaxWidth()
+    ) {
+        Column(modifier = Modifier.fillMaxWidth()) {
+            currentUser?.let {
+                CurrentAccountRow(user = it, onClick = onCurrentUserClick)
```
alperozturk96: As far as I can see the current account row must be always in the top if this is the case:

You can use stickyHeader inside of the `LazyColumn` and get rid of another `Column` inside `Surface`

```
stickyHeader { 
           
 }
```

Even maybe you can remove `Surface` and give color to the `LazyColumn` directly but I'm not sure how it will look.


You can test something like this:

```
@Composable
fun ChooseAccountShareToContent(
    currentUser: User?,
    otherUsers: List<User>,
    onCurrentUserClick: () -> Unit,
    onOtherUserClick: (User) -> Unit
) {
    LazyColumn(
        modifier = Modifier
            .fillMaxWidth()
            .background(MaterialTheme.colorScheme.surfaceContainerHigh)
    ) {
        currentUser?.let { 
            stickyHeader {
                CurrentAccountRow(user = it, onClick = onCurrentUserClick)
            }
        }
        items(otherUsers) { user ->
            OtherAccountRow(user = user, onClick = { onOtherUserClick(user) })
        }
    }
}
```

app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountShareToDialogFragment.kt
```diff
@@ -14,144 +14,98 @@ import android.os.Bundle
...
-                binding!!.currentAccount.userIcon.loadUserAvatar(user, user.userId!!, true, false)
-            } else {
-                binding!!.currentAccount.userIcon.visibility = View.INVISIBLE
+        viewModel = ViewModelProvider(this, viewModelFactory)[ChooseAccountShareToViewModel::class.java]
+
+        val otherUsers = mutableStateOf<List<User>>(emptyList())
+
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                viewModel.viewState.collect { state ->
```
alperozturk96: I think you need latest value if it's the case you can use `collectLatest` instead of `collect`.

ref: https://www.baeldung.com/kotlin/flow-collect-vs-collectlatest#understanding-collectlatest

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToDialogFragment.kt
```diff
@@ -14,144 +14,98 @@ import android.os.Bundle
...
-        if (userItems.size > position) {
-            val user = userItems[position].user
-            if (userManager!!.setUserAsActive(user!!).blockingGet()) {
-                cookieManager!!.cookieStore.removeAll()
-                activity?.recreate()
-                dismiss()
-            }
-        }
-        true
+        composeView = null
```
alperozturk96, line 108: Shouldn't be like this?

```
composeView = null
super.onDestroyView()
```

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
AndyScherzinger: Hmm, good question. I am used to "call super first" but fine to flip the order

app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountShareToDialogCompose.kt
AndyScherzinger: Yeah, let me try that. I was also a bit surprised I had to set the background color at all given I passed the material theme but the dialog's background wasn't the "Nextcloud dynamic one"

app/src/main/java/com/nextcloud/talk/ui/dialog/viewmodels/ChooseAccountShareToViewModel.kt
AndyScherzinger: > however, this is out of scope for this PR

Lemme check, I think it would also be fine to do that in this PR 👍

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
AndyScherzinger: I don't know, with the regular Account Switcher dialog I sometimes had the impression that loading can take quite a while, can't say why that has been the case. What do you think @mahibi ?

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToDialogFragment.kt
AndyScherzinger: Can't say if it makes much of a difference but I flipped the order 👍

alperozturk96: I think this order 

```
override fun onDestroyView() {
        super.onDestroyView()
        composeView = null
```

correct because the Fragment clean up its internal view state first then we can clear reference to new composable view but for other one first disposables needs to be disposed then we can clear the view model.

after testing it you can ignore this suggestion, sorry.

app/src/main/java/com/nextcloud/talk/ui/chooseaccount/ChooseAccountShareToViewModel.kt
AndyScherzinger: Fixed

app/src/main/java/com/nextcloud/talk/ui/dialog/ChooseAccountShareToDialogCompose.kt
AndyScherzinger: Done, shortened it 👍 
Kept/added a surface wrapper for general background and theme inheritance

PR #5921, opened 2026-03-02
⚒️ Migrate ConversationInfoActivity, Participants and MentionAutoComplete 

AndyScherzinger:
…away from flexible adapter

I tested @ mentions in message input as well as participant lists in the conversation info for users, guests and groups with status / emoji status and in a video call.

Also since this already touches some classes, kept it at that while still doing some renamings. Also didn't touch the packaging while this might need some cleanup in the future yet isn't worse than before (hinting to the mix of autocomplete classes and participant links)

Also some handling extraction in initViewStateObserver in the activity to make detekt happy again... 🙄 

This is a follow-up to #5914 

### 🚧 TODO

- [ ] get review feedback & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/MentionAutocompleteAdapter.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/MentionAutocompleteAdapter.kt
@@ -105,37 +105,25 @@ class MentionAutocompleteAdapter(
     }
 
     private fun setAvatar(holder: ViewHolder, item: MentionAutocompleteItem) {
+        val avatarView = holder.binding.avatarView
         when (item.source) {
             SOURCE_CALLS -> {
-                if (isPhoneNumber(item.displayName)) {
-                    holder.binding.avatarView.loadUserAvatar(
-                        viewThemeUtils.talk.themePlaceholderAvatar(
-                            holder.binding.avatarView,
-                            R.drawable.ic_phone_small
-                        )
-                    )
+                val placeholder = if (isPhoneNumber(item.displayName)) {
+                    R.drawable.ic_phone_small
                 } else {
-                    holder.binding.avatarView.loadUserAvatar(
-                        viewThemeUtils.talk.themePlaceholderAvatar(
-                            holder.binding.avatarView,
-                            R.drawable.ic_avatar_group_small
-                        )
-                    )
+                    R.drawable.ic_avatar_group_small
                 }
+                avatarView.loadUserAvatar(viewThemeUtils.talk.themePlaceholderAvatar(avatarView, placeholder))
             }
 
-            SOURCE_GROUPS -> {
-                holder.binding.avatarView.loadUserAvatar(
-                    viewThemeUtils.talk.themePlaceholderAvatar(
-                        holder.binding.avatarView,
-                        R.drawable.ic_avatar_group_small
-                    )
+            SOURCE_GROUPS ->
+                avatarView.loadUserAvatar(
+                    viewThemeUtils.talk.themePlaceholderAvatar(avatarView, R.drawable.ic_avatar_group_small)
                 )
-            }
 
             SOURCE_FEDERATION -> {
                 val darkTheme = if (DisplayUtils.isDarkModeOn(context)) 1 else 0
-                holder.binding.avatarView.loadFederatedUserAvatar(
+                avatarView.loadFederatedUserAvatar(
                     currentUser,
                     currentUser.baseUrl!!,
                     roomToken,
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/talk/chat/MentionAutocompleteAdapter.kt
```diff
@@ -0,0 +1,226 @@
...
+                    currentUser,
+                    item.objectId!!,
+                    requestBigSize = true,
+                    ignoreCache = false
+                )
+            }
+        }
+    }
+
+    private fun drawStatus(holder: ViewHolder, item: MentionAutocompleteItem) {
```
alperozturk96, line 154:
```
import android.content.Context
import com.nextcloud.talk.R

enum class StatusType(val string: String) {
    ONLINE("online"),
    OFFLINE("offline"),
    DND("dnd"),
    AWAY("away"),
    BUSY("busy"),
    INVISIBLE("invisible");

    companion object {
        fun getDescription(value: String?, context: Context): String {
            return when(value) {
                DND.string -> context.getString(R.string.dnd)
                BUSY.string -> context.getString(R.string.busy)
                AWAY.string -> context.getString(R.string.away)
                else -> ""
            }
        }
    }
}
```


```
private fun drawStatus(holder: ViewHolder, item: MentionAutocompleteItem) {
        val size = DisplayUtils.convertDpToPixel(STATUS_SIZE_IN_DP, context)

        holder.binding.userStatusImage.setImageDrawable(
            StatusDrawable(
                item.status,
                NO_ICON,
                size,
                ContextCompat.getColor(context, R.color.bg_default),
                context
            )
        )

        val statusMessage = StatusType.getDescription(item.statusMessage, context)
        holder.binding.conversationInfoStatusMessage.text = statusMessage
        alignUsernameVertical(holder, if (statusMessage.isEmpty()) NO_USER_STATUS_DP_FROM_TOP else 0f)

        holder.binding.participantStatusEmoji.run {
            if (!item.statusIcon.isNullOrEmpty()) {
                setText(item.statusIcon)
                visibility = View.VISIBLE
            } else {
                visibility = View.GONE
            }
        }
    }
```

app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
```diff
@@ -815,55 +794,53 @@ class ConversationInfoActivity :
...
+                participant.sessionId = "-1"
+                ownUiItem = ParticipantModel(participant, true)
             } else {
-                userItems.add(userItem)
+                uiItems.add(ParticipantModel(participant, isOnline))
             }
         }
 
-        Collections.sort(userItems, ParticipantItemComparator())
+        Collections.sort(uiItems, ParticipantItemComparator())
```
alperozturk96: You can remove `ParticipantItemComparator` but please double-check comparasion logic.

```
uiItems.sortWith(
            compareBy(
                { it.participant.actorType == GROUPS || it.participant.actorType == CIRCLES },
                { !it.isOnline },
                { it.participant.type !in listOf(
                    Participant.ParticipantType.MODERATOR,
                    Participant.ParticipantType.OWNER,
                    Participant.ParticipantType.GUEST_MODERATOR
                )},
                { it.participant.displayName!!.lowercase(Locale.ROOT) }
            )
        )
```

app/src/main/java/com/nextcloud/talk/conversationinfo/ParticipantItemAdapter.kt
```diff
@@ -37,77 +39,71 @@ import com.nextcloud.talk.ui.theme.ViewThemeUtils
...
-    var isOnline = true
-    override fun equals(o: Any?): Boolean =
-        if (o is ParticipantItem) {
-            model.calculatedActorType == o.model.calculatedActorType &&
-                model.calculatedActorId == o.model.calculatedActorId
-        } else {
-            false
+    private val conversation: ConversationModel,
+    private val onItemClick: (ParticipantModel) -> Unit
+) : ListAdapter<ParticipantModel, ParticipantItemAdapter.ViewHolder>(DiffCallback) {
```
alperozturk96, line 49: This is a really good upgrade. Using `ListAdapter` and `DiffCallback` definitely improve the performance for big lists however please double-check delete, delete all, add, add all etc otherwise you may end-up with crash.

alperozturk96 (changes requested):

app/src/main/java/com/nextcloud/talk/conversationinfo/ConversationInfoActivity.kt
AndyScherzinger: Code looks good 👍 
Added

app/src/main/java/com/nextcloud/talk/chat/MentionAutocompleteAdapter.kt
AndyScherzinger: Added

app/src/main/java/com/nextcloud/talk/conversationinfo/ParticipantItemAdapter.kt
AndyScherzinger: > please double-check delete, delete all, add, add all etc otherwise you may end-up with crash.

@alperozturk96 Could you elaborate on this? It currently is only called once via `participantAdapter!!.submitList(uiItems)` in the same way `InvitationsAdapter` has been implemented in another component which also only calls `submitList(items)`.

So I guess it is fine as is. Also all tests I did had been successful.

So any details would help, or I misunderstood and given the above "I did double check"?

AndyScherzinger: So if above is fine, can I get a re-review @alperozturk96 🙏

alperozturk96: It depends on the calling method. If there’s only one place calling it and `submitList` is triggered only from that method, it should be fine.

However, if there are multiple callers, and some call it from a background thread while others call it from the main thread, you may run into crashes due to invalid adapter positions.

For example, when I was trying to improve the performance of file listing in the Files client, we had multiple caller methods and a lot of unnecessary calls. Some of them needed to wait for an async task result, while others switched directly to the main thread. Moving to `ListAdapter` and using `DiffUtil` caused a lot of issues because the calls weren’t consolidated.

If there aren’t any architectural problems like the ones I described, then it should be okay.

AndyScherzinger: Then we should be fine, yes.
It is only called once and also only in that one place via submitList 👍

PR #5931, opened 2026-03-05
⚒️🎨 Migrate user info details to Composables

AndyScherzinger:
... style like latest expressive M3 contacts.

I'll be splitting the work on migrating to composables as a first modernization step into several UI-targeting PRs. This is the 1st.

## **Remark**
You might spot that disabled input field that is because that field has never been editable but the former UI did not make that transparent. For the sake of dev speed I kept the logic as is. For further reference - we could also think about a filter on the adapter that is currently in place for view mode hiding items with no values in them and could extend that one for edit mode hiding elements without editing permission.

## **Remark2**
On initial change you'll face http400 errors to to a server behavior change requesting re-authenticaiton via password confirmation which the clients doesn't do yet. This is also broken on master/older releases and *I would like this PR not to be blocked by that*. In order to test, first go to the web interface and change a value, after prompted for your password and submitted, you can start changing/adding fields on the client.

## **Remark3**
There is another pre-existing bug, invalid inputs will just be rejected showing a "saving failed" while looking at the frontend which implemented actual validation blocks on the frontend level. This is missing previous to this PR, so again, *I would like this PR not to be blocked by that*.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before_light_read|after_light_read
before_light_edit|after_light_edit
before_dark_read|after_dark_read
before_dark_edit|after_dark_edit

### 🚧 TODO

- [x] implement editing capability (currently values are hard-reset upon editing)
- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set


AndyScherzinger:
> Very nice, looks much better. Just one detail – the email address should ellipsize instead of breaking into multiple lines. Especially here with only the ".com" in the second line it looks off.

@jancborchardt fixed 👍 (did not update the screenshots though)

mahibi:
when editing, the keyboard overlaps the fields address, website and "X" aka twitter (btw i absolutely agree with https://github.com/nextcloud/server/issues/50841#issuecomment-2661523941)

in the old behavior, the fields that are being edited scrolled up so they were seen above the keyboard

PR #5936, opened 2026-03-06
🪪 Add further profile items including enablement switch

AndyScherzinger:
* Add all other profile fields + icons and enable/disable profile (new composable)
* fetches the profile data to display also in case editing permissions can be fetched or end up in an error

The mixed language is due to the lack of translations, which would happen after the merge via transifex.

### 🖼️ Screenshots

Description | 🏡 Screenshot
---|---
Top read | Screenshot_20260306_175858
Bottom read | Screenshot_20260306_175922
Top edit | Screenshot_20260306_175935

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/profile/ProfileActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/profile/ProfileActivity.kt
@@ -115,16 +115,21 @@ class ProfileActivity : BaseActivity() {
         }
     }
 
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        outState.putBoolean(KEY_EDIT_MODE, edit)
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         NextcloudTalkApplication.sharedApplication!!.componentApplication.inject(this)
         binding = ActivityProfileBinding.inflate(layoutInflater)
+        edit = savedInstanceState?.getBoolean(KEY_EDIT_MODE) ?: false
         setupActionBar()
         setContentView(binding.root)
         initSystemBars()
         val colorScheme = viewThemeUtils.getColorScheme(this)
         binding.profileSettingEnabledProfile.apply {
-            this!!.setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
             setContent {
                 MaterialTheme(colorScheme = colorScheme) {
                     ProfileEnabledCard(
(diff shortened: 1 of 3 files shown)


jancborchardt (changes requested):
Very nice! :)

Small things:
- [ ] We use American English, so "Organization" with z instead of "Organisation".
- [ ] The right side of the "Enable profile" box in edit mode seems a bit unaligned with everything else. Is it correct?

app/src/main/res/layout/activity_profile.xml
```diff
@@ -133,6 +133,16 @@
 
         </LinearLayout>
 
+        <androidx.compose.ui.platform.ComposeView
```
mahibi, line 136: this view is missing in landscape mode so it will crash with a NPE

app/src/main/java/com/nextcloud/talk/profile/ProfileActivity.kt
```diff
@@ -118,6 +122,18 @@ class ProfileActivity : BaseActivity() {
         setupActionBar()
         setContentView(binding.root)
         initSystemBars()
+        val colorScheme = viewThemeUtils.getColorScheme(this)
+        binding.profileSettingEnabledProfile.apply {
+            this!!.setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
```
mahibi: only apply if it's existing (would prevent NPE even if view was not found).

```
binding.profileSettingEnabledProfile?.apply {
```

`this!!` in `this!!.setViewCompositionStrategy` is not necessary.
But also i think OnViewTreeLifecycleDestroyed is not necessary here as we are just a normal activity. The default behavior will be just fine, so just delete

`this!!.setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)`

mahibi (dismissed):

app/src/main/res/layout/activity_profile.xml
AndyScherzinger: Argh, didn't even see the landscape .... Sorry about that, yeah, crashed obviously.... Fixed it 👍

app/src/main/java/com/nextcloud/talk/profile/ProfileActivity.kt
AndyScherzinger: I removed it. The reason it is in there is to keep the composition alive until the Activity is actually destroyed, regardless of window attachment state — the same strategy used for ComposeView in Fragments and the recommended approach for host-lifecycle-bound composition. But since we never reload profile data the scenario just isn't there.

AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/5936/commits/18f9986a9dc3ac426dacdc58cf4fe13f4c70945e

app/src/main/res/layout/activity_profile.xml
AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/5936/commits/f718416d05fe6d2880f19299a972e69b4c1f0486

AndyScherzinger:
@mahibi fixed your comments and also added edit state preservation upon screen rotation (also not implemented before but now we have it)

PR #5943, opened 2026-03-10
⚒️🎨 Migrate Profile to Composable

AndyScherzinger:
* Migrate the activity and scope dialog layouts to composable
* Move away from item adapter to be used for the scope dialog but use `MutableList` instead

Note: Ultimately but not done by this PR we should move to MVVM - When having/implementing a `ProfileViewModel`, the item list and `updateScope` belong there as a `StateFlow<List<UserInfoDetailsItem>>`. `ScopeDialog`'s `onScopeSelected` would call  `viewModel.updateScope(position, scope)` and the UI observes the flow. This would be even cleaner but is a larger refactor, making the review more complex. So better do it step by step.

### 🚧 TODO

- [x] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set


sowjanyakch:
Edit the profile and rotate the screen. Edited changes are lost - changes must be retained. This needs to be implemented along with MVVM.

AndyScherzinger:
@sowjanyakch regarding center alignment of https://github.com/nextcloud/talk-android/pull/5943#pullrequestreview-3955708204 for the buttons, if center aligned than they aren't aligned with the profile activation bo x_nor_ with the above user displayname or any other element, so it is very unaligned then, no?

Any feedback @nimishavijay @jancborchardt  ?

nimishavijay:
I would agree that center aligning looks off, I'm wondering if a layout like this is feasible? This removes the center/left aligned issue as the only option is center aligning everything in the left column. In the right column we should make sure to align the icons of the "Enable profile" item with the other icons below

image

Additionally, if possible we should make the following changes:
- [x] Use a smaller toggle icon for the "Enable profile" toggle
- [x] Reduce its height to the standard height (should be 44px?)
- [x] Remove the background for the "Enable profile" toggle 
What do you think? @AndyScherzinger @sowjanyakch

AndyScherzinger:
> Use a smaller toggle icon for the "Enable profile" toggle

Definitive no on this one. The switch is standard M3/Android.

> Reduce its height to the standard height (should be 44px?)

Let's see, list items would be min 56dp but... I did not add it to the list for a reason since all other items are input fields, so 
this complicates thing, but let's see. First step could still be to keep it on the left column for easier implementation life.

> Remove the background for the "Enable profile" toggle

Sure thing.

AndyScherzinger:
@nimishavijay @sowjanyakch 
Screenshot_20260318_114553

nimishavijay:
Nice! I checked the M3 specs and seems like there is no way to get around it. in that case the priority would be the spacing and alignment 

- [x] center align all the icons on the right with the toggle icon
- [x] add some spacing between the text fields and the icons such that there is equal spacing on the left and right of the icons 

Once that's done it should be good to go!

AndyScherzinger:
@nimishavijay final iteration :)

Screenshot_20260318_134058

app/src/main/java/com/nextcloud/talk/profile/AvatarSection.kt
```diff
@@ -0,0 +1,132 @@
...
+                    transitionName = "userAvatar.transitionTag"
+                    contentDescription = ctx.getString(R.string.avatar)
+                }.also { imageView ->
+                    DisplayUtils.loadAvatarImage(state.currentUser, imageView, state.avatarRefreshKey > 0)
+                }
+            },
+            modifier = Modifier.size(avatarSize).clip(CircleShape)
+        )
+    }
+}
```
mahibi, line 170: lets avoid to use AndroidView/ImageView and use AsyncImage. Like:

```
@Composable
private fun AvatarImage(state: ProfileUiState, avatarSize: Dp) {
    val user = state.currentUser
    val url = ApiUtils.getUrlForAvatar(user?.baseUrl,state.currentUser?.userId, false)

    AsyncImage(
        model = ImageRequest.Builder(LocalContext.current)
            .data(url)
            .crossfade(true)
            .build(),
        contentDescription = stringResource(R.string.avatar),
        modifier = Modifier
            .size(avatarSize)
            .clip(CircleShape)
    )
}
```

mahibi (changes requested):

app/src/main/java/com/nextcloud/talk/profile/AvatarSection.kt
AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/5943/commits/8a0849e18b22c1fc8b2b52f49d4baae48f0c5dfd

app/src/main/java/com/nextcloud/talk/profile/AvatarSection.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.nextcloud.talk.R
+import com.nextcloud.talk.utils.ApiUtils
+
+@Composable
+private fun AvatarImage(state: ProfileUiState, avatarSize: Dp) {
+    val user = state.currentUser
+    val url = ApiUtils.getUrlForAvatar(user?.baseUrl,state.currentUser?.userId, true)
+    val model = ImageRequest.Builder(LocalContext.current)
+        .data(url)
+        .crossfade(true)
```
mahibi: .memoryCacheKey(..)

could be used for the caching

AndyScherzinger: Fixed via https://github.com/nextcloud/talk-android/pull/5943/commits/bec78fb8934d4ae7c195283d1487f8693309a90f

PR #5962, opened 2026-03-12
⚒️🎨 Migrate to MapLibre

AndyScherzinger:
* replace OSMdroid with MapLibre and re-implement/preserve the features (current location, zoom, etc)
* ensure geoActivity result is preserved upon device rotation or any other non-location-changing action
* migrate from raster to vector tiles with nice color schemes for dark/light. You can use [maputnik](https://maplibre.org/maputnik/?layer=287555303%7E1&modal=codeEditor#15.09/52.51771/13.39745) to load the json themes for display and editing
* Implement attribution / info

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
old_light|new_light2
old_dark|new_dark2

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -303,7 +303,7 @@ dependencies {
 
     implementation("io.github.elye:loaderviewlibrary:3.0.0")
     implementation("org.maplibre.compose:maplibre-compose:0.12.1")
-	implementation("org.maplibre.compose:maplibre-compose-material3:0.12.1")
+    implementation("org.maplibre.compose:maplibre-compose-material3:0.12.1")
     implementation("fr.dudie:nominatim-api:3.4") {
         //noinspection DuplicatePlatformClasses
         exclude(group = "org.apache.httpcomponents", module = "httpclient")


AndyScherzinger:
@sowjanyakch your screenshot points to a code base _before all_ of my changes, because
* tiles are rastered nit vector
* shows a scalebar, yet hidden in my update + I show an info button

So you are lacking most of the commits and hence fixes and improvements beyond the pure lib migration.

I just rebased due to latest changes on main branch, so you might want to do a fresh checkout. In short, if the map doesn't look like the screenshots on the after-column color- and sharpness-wise. Than it is not the latest code of this branch and PR.

mahibi:
> Search for a location and then press enter. The map points to the selected location but under Share this location button at the bottom, location details are not updated (in your screenshot, I see the location details).

i could bet the map was moved after coming back from the geocoding screen? As soon as the map is moved the location details are gone until the new place is actually shared. This is done to not stress the server with too many geocoding queries as we are using it for free..

AndyScherzinger:
> This is done to not stress the server with too many geocoding queries as we are using it for free..

Yes, this. Technically we are only allowed 1 request per second overall (as in all clients together!)

PR #5981, opened 2026-03-19
feat: Show draft message in case the conversation has one attached to it

AndyScherzinger:
- Ref: https://github.com/nextcloud/spreed/issues/15542

also fixing the issue that before when fetching the conversations from the server, the server-state has been exposed to the UI, not the upserted state from DB including the client-side only data.

### 🖼️ Screenshots

yellow

### 🚧 TODO

- [ ] review & merge

### 🏁 Checklist

- [x] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.kt
+++ b/app/src/main/java/com/nextcloud/talk/adapters/items/ConversationItem.kt
@@ -246,6 +246,16 @@ class ConversationItem(
         }
 
     private fun setLastMessage(holder: ConversationItemViewHolder, appContext: Context) {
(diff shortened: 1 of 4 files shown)


mahibi:
Should we modify the time?
I see other messengers show the time when the draft was created. 
I'm not sure yet what time i expect there but i guess adapting to other messengers makes sense..?

AndyScherzinger:
> Should we modify the time?
> I see other messengers show the time when the draft was created.
> I'm not sure yet what time i expect there but i guess adapting to other messengers makes sense..?

Matches @nimishavijay's comment, but we currently lack a timestamp. Will add it in a follow-up and respect that timestamp in the sorting logic 👍

Issue #6011, opened 2026-03-30
WebP image upload support


anakin78z:
**Is your feature request related to a problem? Please describe.**
Currently webp images fail to upload. There is no specific error message, the upload just fails. With webp images being pretty common these days, this can be pretty frustrating.

**Describe the solution you'd like**
It looks like Nextcloud Talk may not support webp images at all, as they're also failing to upload int the web app. I propose that the android app converts it to a different format to allow the upload.

**Describe alternatives you've considered**
Ideally Nextcloud would just support webp images throughout, but that seems like a bigger task.
The user can convert images themselves, but that's pretty annoying and it's not always obvious how they can do that. The app handling the conversion seems like the easiest path forward.

**Additional context**

<img width="540" height="1170" alt="Image" src="https://github.com/user-attachments/assets/911b324f-e021-4d5d-b431-f3822ecb11a8" />

<img width="540" height="146" alt="Image" src="https://github.com/user-attachments/assets/4cc362ad-dc7a-48fb-a56e-70c0618cf29c" />

mahibi:
thanks for reporting

anakin78z:
This turned out to be a server issue which has been addressed: https://github.com/nextcloud/photos/issues/3425

PR #6013, opened 2026-03-31
Allow uploading of WebP images by converting to supported format.

anakin78z:
Addresses #6011

Detects WebP images on upload and converts them before upload. Transparent images are converted to PNG files, and opaque images are converted to JPG.

### 🖼️ Screenshots
x | x | x
---|---|---
Screenshot_20260331_152801_Talk | Screenshot_20260331_153349_Talk | Screenshot_20260331_153410_Talk

### 🚧 TODO

Optional: Display specific notification to user if the conversion fails. Currently it displays the same upload failed notification as before. 

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/UploadAndShareFilesWorker.kt
@@ -38,6 +38,7 @@ import com.nextcloud.talk.upload.normal.FileUploader
 import com.nextcloud.talk.users.UserManager
 import com.nextcloud.talk.utils.CapabilitiesUtil
 import com.nextcloud.talk.utils.FileUtils
+import com.nextcloud.talk.utils.Mimetype
 import com.nextcloud.talk.utils.NotificationUtils
 import com.nextcloud.talk.utils.RemoteFileUtils
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_INTERNAL_USER_ID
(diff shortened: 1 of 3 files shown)


nickvergessen:
I think this should be optional. Maybe you want to give the file to someone else so they can continue the work, or something a like. Since there is a question for upload anyway we could add an option to send as original

anakin78z:
@nickvergessen It doesn't allow uploading as original as the server rejects it (see the linked issue). They would have to share it a different way. 

We could maybe update the question text from 'Send this file' to 'Send this image'? Or, where it prints the file name, we could change it to 'file.webp → file.png'.

nickvergessen:
> It doesn't allow uploading as original as the server rejects it (see the linked issue). They would have to share it a different way.

That sounds like a bug and needs fixing then.
I went to wikipedia, downloaded the sample image: https://upload.wikimedia.org/wikipedia/commons/b/b2/Vulphere_WebP_OTAGROOVE_demonstration_2.webp

Went to Nextcloud Files and was able to upload it just fine:
grafik

I even get a preview. So webp uploading is and should be working normally be accepted in general.

anakin78z:
@nickvergessen in Files, yes, but not in talk. The web client also fails if you try to upload a webp in a conversation. I agree that nextcloud talk should just support it at the server level, but I don't know what all the implications are. If you know any of the server devs, can you tag them to start that discussion?

nickvergessen:
> The web client also fails if you try to upload a webp in a conversation.

That is confusing and then also a bug and something we should fix.
Actually Talk should be using the very same endpoints as the file upload on the files app, so I'm quite confused why it would not work in Talk.

nickvergessen:
grafik

Also works pretty fine here

anakin78z:
Odd, this is what I get: 
image

And here are the logs from the server: 
image

I used the AIO install. I'll see if there are any known issues there.

anakin78z:
OK, I found the related issue: https://github.com/nextcloud/photos/issues/3425

I'll close this PR.

Issue #5992, opened 2026-03-22
The ringing doesn't stop


Byter3:
### Steps to reproduce

Today I had the pleasure to notice it on multiple Android devices, that Talk does not stop to ring even if the user already answered the call and he/she ir actually in the call already. The ringing will still be audiable.

### Expected behaviour

Not to ring when answered.

### Actual behaviour

Today I had the pleasure to notice it on multiple Android devices, that Talk does not stop to ring even if the user already answered the call and he/she ir actually in the call already. The ringing will still be audiable.

### Device brand and model

Galaxy A70 GHalaxy S21 Galaxy S9 FE

### Android version

latest

### Nextcloud Talk app version

latest

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

pumuckly:
I have similar issue, and I can handle it with open UDP port for janus media service.
Edit the file: /etc/janus/janus.jcfg
At the `media:` section define `rtp_port_range` like:
```
media: {
    rtp_port_range = "10000-10100"
}
```
You may free to change the port values if you need more than 100 parallel connections.
Please do not forget to open these UDP ports in firewall or redirect these ports through the router.

Something else may be causing the problem, but it's worth testing this setting first.

mahibi:
Thanks for reporting! Indeed this sounds an annoying bug!

Thanks for sharing the insight @pumuckly , we will have a look

mahibi:
this is solved by https://github.com/nextcloud/talk-android/pull/6015
Will be available in 23.0.1 soon

Issue #5922, opened 2026-03-03
Call notification is not actually call


pfurovYnP:
### Steps to reproduce

Call an android client

### Expected behaviour

When the android client is recieving a call, the actuall call should pop up, like in desktop client/iOS client. Or like in any other messenger app (WhatsApp/telegram)

### Actual behaviour

Only a simple message-like push nortification pops up

### Device brand and model

Honor Magic 8 pro

### Android version

16

### Nextcloud Talk app version

23.0.0

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

sowjanyakch:
Could you please share the diagnosis settings? (Settings - Advanced - Diagnosis - share via upper right corner menu)

pfurovYnP:
> Could you please share the diagnosis settings? (Settings - Advanced - Diagnosis - share via upper right corner menu)

### Meta information

Generation of system report

Mar 4, 2026 10:30:48

### Phone

Device

HONOR BKQ-N49

Android version

36

Google Play services

Google Play services are available

### App

App name

Nextcloud Talk

App version

v23.0.0

Build flavor

gplay

Battery settings

Battery optimization is ignored, all fine

Notification permissions

Notifications are granted

Calls notification channel enabled?

Yes

Messages notification channel enabled?

Yes

Firebase push token

fxJMv...

Latest firebase push token generation

Mar 2, 2026 18:45:50

Latest firebase push token fetch

Mar 4, 2026 10:30:08

Registered users

1

### Current account

Server

https://****

User

Паша

User status enabled?

Yes

Server notification app installed?

Yes

Latest push registration at server

Mar 4, 2026 10:30:08

Latest push registration at push proxy

Mar 4, 2026 10:30:09

Server version

32.0.6

Server Talk version

22.0.9

Signaling Mode

External

mahibi:
What is the content of the "simple message-like push nortification" ?

Does it happen always or only sometimes?

pfurovYnP:
> What is the content of the "simple message-like push nortification" ?
> 
> Does it happen always or only sometimes?

@mahibi 

![Image](https://github.com/user-attachments/assets/b54f0b0b-0371-4dff-af3a-e8d05f580ec7)

it happens always


upd: here is what I expect it to look like (Telegram)
![Image](https://github.com/user-attachments/assets/8f446903-d10e-49c8-a38c-d2fdd5cc62d3)
[...]

Issue #1667, opened 2021-11-04
join call without unlock screen


mahibi:
## Steps to reproduce
1. Lock phone with lockscreen
2. call phone from web
3. (phone should wake up and show notification screen), accept call

### Expected behaviour
go to call without the need to unlock the phone after accepting.
Lockscreen should only be shown when user navigates to other screens (for privacy security reasons; also when entering picture-in-picture mode). 

### Actual behaviour
lockscreen is shown. Not able to join call without to unlock.

this behaviour occured when implementing https://github.com/nextcloud/talk-android/pull/1655
see comment https://github.com/nextcloud/talk-android/pull/1655/files#r738453308
see commit c0cca130ee84628510f7bc4e64ad3f93f8ab9d5f

jookk:
Issue still persist on android 14. When I look at Applications - Special settings permissions, Talk app doesn't ask for Showing notifications on full screen and Turn on screen  
Receiving the call behaves badly for user to pick call. First unlock then picking call.

mahibi:
> Talk app doesn't ask for Showing notifications on full screen and Turn on screen

But you can turn on these permissions, right?
I think the points mentioned in https://github.com/nextcloud/talk-android/issues/3957#issuecomment-2713481693 should be implemented to solve it

jookk:
No, i cant enable show on the top permission in Special permission. Toogle isn't there for talk app

mahibi:
yes, i think "show on the top permission in Special permission" won't be there as it's request is not implemented.
Showing notifications on full screen and Turn on screen could be enabled i think, but it's not the solution i guess.

jookk:
I don't have Talk in Turning on screen permissions.
Yes, it's confusing for people to pickup the incoming call. Thanks

Issue #708, opened 2020-01-20
Make it possible to switch from phone to video call


j-ed:
I've just played around with the talk app and tested different call scenarios. If I e.g. initiate a phone call from the desktop to a mobile phone I had to decide how to pickup the call. If I touch the phone call button, a normal call is initiate as expected.
On the desktop I can transparently switch to a video call, because the camera symbol is still shown at the bottom of the screen, but on Android I don't have this ability. I have to leave the call and return to the chat dialog and touch the video symbol on top of the window.

Therefore it would be nice if the following function would be added to the app:
1. It should be signaled to the called person, if the desktop caller has enabled his camera, because an enabled video isn't automatically shown on the mobile phone. But if the called person on his mobile device enables the video, this is instantly shown on the desktop.

2. It should be possible to switch transparently from a phone to a video call. For this an additional 
button should be added to the call screen. Currently only three buttons are shown: mute, loud speaker, hang-up call

### Environment
Nextcloud version: 18.0.0
Talk app version: 8.0.0
Coturn server set-up and used.

mario:
The first is not possible. But the second I can work on :)

Thanks for your requests.

EliterScripts:
Yes, I would very much appreciate this added as well!

@mario do you work for Nextcloud? Who is the person behind Nextcloud Talk Android?

elhananjair:
> it's not possible right now as it requires bigger changes. For now it's necessary to end the call and join with videocall.

At least it automatically rejoin on my IOS when I touch the video icon on the call, that's a good thing

mahibi:
This should be implemented via renegotiation like it was done in the iOS app:
- https://github.com/nextcloud/talk-ios/pull/2587

PR #6015, opened 2026-04-01
Improve answering calls

### LINKED ISSUE #5351: Notifications d’appels manquantes avec étui S-View (Galaxy S24 Ultra, Android 15 / One UI 7.0)
### Steps to reproduce

1. Installer Nextcloud Talk depuis le Play Store.  
2. Utiliser un **Samsung Galaxy S24 Ultra** sous **Android 15 / One UI 7.0**.  
3. Activer un **étui S-View officiel Samsung**.  
4. Recevoir un appel entrant avec l’étui fermé (téléphone en veille ou non).

### Expected behaviour

L’appel entrant devrait générer une notification d’appel (plein écran ou équivalent `CATEGORY_CALL`), même lorsque l’étui est fermé, comme c’est le cas avec Messenger ou WhatsApp.

### Actual behaviour

- Étui ouvert : les appels sont bien notifiés (veille et non veille).  
- Étui fermé : aucune notification d’appel, seule la date/heure s’affiche.  
- Les notifications de chat fonctionnent normalement.  

### Device brand and model

Samsung Galaxy S24 Ultra

### Android version

15

19.0.15

### Nextcloud server version

30

### Talk version

_No response_

None

None

None

### Android logs

_No response_

### Server log

```bash

```

### Additional information

_No response_

anakin78z:
Fixes some of the biggest annoyances with answering calls: 

1. Allow answering a call from a locked phone without having to unlock the phone after answering. #1667 (kind of?)
2. Stop calls from ringing after answering. #5992 
3. Update the incoming call notification to allow users to answer or decline. #5922
4. Show caller avatar on incoming call notification

Also fixes #5351 

### 🖼️ Screenshots

🏚️ Before 
Screenshot_20260401_120002_One UI Home 

🏡 After
 

S-View case notification: 

image
This matches other apps such as Facebook Messenger. I think we're getting this automatically by using `NotificationCompat.CallStyle`

### 🚧 TODO

There are a lot of other issues with calls. This is meant as some small changes that can have a big impact on usability.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -108,7 +108,6 @@ import com.nextcloud.talk.utils.ApiUtils
 import com.nextcloud.talk.utils.CapabilitiesUtil
 import com.nextcloud.talk.utils.CapabilitiesUtil.hasSpreedFeatureCapability
 import com.nextcloud.talk.utils.CapabilitiesUtil.isCallRecordingAvailable
-import com.nextcloud.talk.utils.NotificationUtils.cancelExistingNotificationsForRoom
 import com.nextcloud.talk.utils.NotificationUtils.getCallRingtoneUri
 import com.nextcloud.talk.utils.ReceiverFlag
 import com.nextcloud.talk.utils.SpreedFeatures
@@ -373,7 +372,7 @@ class CallActivity : CallBaseActivity() {
 
     private var isFrontCamera by mutableStateOf(true)
 
-    @SuppressLint("ClickableViewAccessibility")
+    @SuppressLint("ClickableViewAccessibility", "Detekt.LongMethod")
     override fun onCreate(savedInstanceState: Bundle?) {
         Log.d(TAG, "onCreate")
         super.onCreate(savedInstanceState)
@@ -1812,6 +1811,7 @@ class CallActivity : CallBaseActivity() {
         fetchSignalingSettings()
     }
 
+    @Suppress("Detekt.NestedBlockDepth")
     @Subscribe(threadMode = ThreadMode.BACKGROUND)
     fun onMessageEvent(webSocketCommunicationEvent: WebSocketCommunicationEvent) {
         if (currentCallStatus === CallStatus.LEAVING) {
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -30,6 +30,7 @@
 * hidrohase <github@honesz.net>
 * Jan-Christoph Borchardt <hey@jancborchardt.net>
 * jld3103 <jld3103yt@gmail.com>
+* Jens Zalzala <jens@shakingearthdigital.com>
 * Joas Schilling <coding@schilljs.com>
 * John Molakvoæ <skjnldsv@protonmail.com>
 * Jos Poortvliet <jos@opensuse.org>
--- a/app/src/main/java/com/nextcloud/talk/receivers/DeclineCallReceiver.kt
+++ b/app/src/main/java/com/nextcloud/talk/receivers/DeclineCallReceiver.kt
@@ -1,11 +1,8 @@
 /*
  * Nextcloud Talk - Android Client
  *
- *
- * SPDX-FileCopyrightText: 2026 Jens Zalzala <jens@shakingearthdigital.com>
  * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: GPL-3.0-or-later
- *
  */
 package com.nextcloud.talk.receivers
 


mahibi:
Thank you for contributing @anakin78z :+1: 

We will review this soon (will try today but i'm on vacation next week).
As a side note, depending on if https://github.com/nextcloud/talk-android/pull/5928 is merged first, there could be more merge conflicts.

anakin78z:
Thanks. I'm also waiting to hear back about https://github.com/thstyl2000/talk-android/issues/1 since they have a massive refactor plan. I'm not sure what stage they're at with implementation though.

mahibi:
and i oversaw it's "just" a draft for now. Just set the state when it's ready to review and let us know if you have questions.

app/src/main/java/com/nextcloud/talk/activities/CallBaseActivity.java
anakin78z, line 1: This is what was causing the lock screen to appear after answering a call.

anakin78z:
@mahibi It doesn't look like there would be conflicts with https://github.com/nextcloud/talk-android/pull/5928, as my code only changes call notifications, and the bubble code seems to only affect message notifications.

anakin78z:
@nimishavijay Hi, thanks for the feedback. 
I was actually trying to get it to do that, but it looks like the system handles the incoming call display completely. So it may look different depending on what phone you have. 
I'm using `NotificationCompat.CallStyle.forIncomingCall` here: https://github.com/nextcloud/talk-android/pull/6015/changes#diff-b51bc8baa586c30bbd846a44639cba62c95cde0f586adb76db389f9e74635914R349 which doesn't give us any real options other than `.setIsVideo()`. 

Docs here: https://developer.android.com/reference/androidx/core/app/NotificationCompat.CallStyle#setIsVideo(boolean)

I think we would have to use a custom style to change the text or add icons, but maybe that could be a separate task, since it would involve design work?

anakin78z:
Updated description to note that it also fixes https://github.com/nextcloud/talk-android/issues/5351

mahibi:
Thank you @anakin78z 
we will review this soon.

app/src/main/java/com/nextcloud/talk/receivers/CallNotificationActionReceiver.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.nextcloud.talk.receivers
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import androidx.core.app.NotificationManagerCompat
+import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_NOTIFICATION_TIMESTAMP
+
+class CallNotificationActionReceiver : BroadcastReceiver() {
```
mahibi: i suggest 
`DeclineCallReceiver`
as this is the only usecase

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
```diff
@@ -268,11 +269,66 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
...
+                    signatureVerification.user!!.baseUrl!!,
+                    conversation.name,
+                    false,
+                    darkMode = DisplayUtils.isDarkModeOn(applicationContext)
+                )
+                loadAvatarSync(avatarUrl, applicationContext)?.let { callerPersonBuilder.setIcon(it) }
+            }
+            val callerPerson = callerPersonBuilder.build()
+
+            val isVideoCall = (conversation.callFlag and Participant.InCallFlags.WITH_VIDEO) > 0
```
mahibi, line 328: Afaik call flags are still like this: 
It's not possible to know if the caller has video enabled. The video flag is set if there is the possibility to use video, regardless if it's enabled or not.
So this will most of the times result in showing the video button even when the caller has no video enabled.

Would be less annoying if https://github.com/nextcloud/talk-android/issues/708 would be solved.

@SystemKeeper @Ivansss how is this done on iOS?

anakin78z: In my testing it worked as intended:
If I initiate a video call, the video button shows, and if I initiate an audio call, it doesn't. 

That may not handle edge cases, like the caller turning off their video, or switching from audio to video, but I think the most common use case is handled. 

I'll wait to hear how it's handled on iOS though.

app/src/main/java/com/nextcloud/talk/receivers/CallNotificationActionReceiver.kt
anakin78z: Initially I had it set up to add an additional action to the call notification to always give the option to answer with video or audio, but the way the extra action displayed I found it a bit cramped and inconsistent, so I removed it. We could, however, decide to add additional actions in the future, so I left this receiver more generic. 

You can see how the action is being added on line 304 of NotificationWorker.kt: https://github.com/nextcloud/talk-android/pull/6015/changes/BASE..dcde0be993b7ee9733cdca5fa275c13b5b778caa#diff-b51bc8baa586c30bbd846a44639cba62c95cde0f586adb76db389f9e74635914R304

I can change the name if you don't think we'll add other actions. Or I could add some documentation to the Receiver explaining how it could be extended. 

Edit: Actually, answering voice/video wouldn't have used this receiver. So it would have to be an unrelated action, and the only thing I can think of right now is maybe sending a text response instead of answering. But it's probably unlikely that we'd use this for anything other than declining calls for now.

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
mahibi: For iOS it's the same behavior that a call is offered as videocall. 
However on iOS the camera is not activated automatically in this case but you have to enable it yourself. We should do the same.

In the `prepareCall()` method you find that `onCameraClick()` is called.
Could you please wrap this with the following check so it's only automatically enabled when the callscreen is not opened via a notification?:

```
            if (!isIncomingCallFromNotification) {
                onCameraClick()
            }
```

I think this is the expected behavior for most people.

app/src/main/java/com/nextcloud/talk/receivers/CallNotificationActionReceiver.kt
mahibi: I guessed it was a leftover from other approaches. 
I'd suggest to go with `DeclineCallReceiver`.
Then you could also delete the ACTION_DECLINE_CALL action.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
```diff
@@ -553,6 +555,11 @@ class CallActivity : CallBaseActivity() {
 
         if (extras.containsKey(KEY_FROM_NOTIFICATION_START_CALL)) {
             isIncomingCallFromNotification = extras.getBoolean(KEY_FROM_NOTIFICATION_START_CALL)
+            val notificationId = extras.getInt(KEY_NOTIFICATION_TIMESTAMP, 0)
+            if (notificationId != 0) {
+                // cancel the notification to stop the call ringing
+                NotificationManagerCompat.from(this).cancel(notificationId)
```
mahibi, line 560: this should work and fix the issue with the ongoing ringing.

however it should have been covered by the call of `cancelExistingNotificationsForRoom` method in `performCall`, but there seems to be a bug that this is not triggered sometimes.
I will continue debugging in the evening or tomorrow to sort things out..

app/src/main/java/com/nextcloud/talk/receivers/CallNotificationActionReceiver.kt
anakin78z: Fixed in 04606e3c045a085def3c896fea7f9c003df52733

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
anakin78z: Fixed in be0bd8d0e9ffc484e2fc80026dbe08e1997bf154

I will say, however, that I personally don't like this behavior. I'm not used to having to click an extra button to enable video for video calls. I think a setting would be nice for this. I can create an issue for it after this is merged.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
mahibi: i have some ideas what could go wrong, but it's not settled.

To simplify things, i suggest to keep your approach, but to remove the block

```

                            if (!TextUtils.isEmpty(roomToken)) {
                                cancelExistingNotificationsForRoom(
                                    applicationContext,
                                    conversationUser!!,
                                    roomToken!!
                                )
                            }
```
                            
from the function getRoomAndContinue completely. 
So even when other things go wrong, the ringing should definitely stop.

Just let me know if you want to do this or if i should take over from here.

app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
mahibi: I understand for some people it's not the expected behavior but as it's privacy related we should disable it initially.
I asked my team and they also voted to disable it.

Some background: the call flag just tells that there is a device existing that could transmit video. It does not say video is enabled. So most of the time the called person will see "incoming video call", but actually the other participant does not have video enabled. In this case, most people will be confused to unveil their own video after accepting the call but don't the the other participants video.

app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
anakin78z: @mahibi feel free to take over

mahibi:
Thanks again @anakin78z :+1: :+1:

Issue #6044, opened 2026-04-11
Call fails to join existing call on app launch due to stale roomJoined WebSocket event


tareko:
### Steps to reproduce

When launching the app and immediately joining a call that's already in progress, the call sometimes fails to connect with existing participants. The app appears stuck in a "connecting" state. The user must exit the call (and sometimes the channel) and re-enter multiple times before it works.

1. Have an active call in progress on another device
2. Open the Android app fresh (cold start or returning after being away)
3. Tap to join the existing call
4. Call appears to connect but never shows other participants — stuck in connecting state

### Expected behaviour

The app should join the existing call and connect to all current participants.

### Actual behaviour

Call fails to join and gets stuck in 'ringing'

### Device brand and model

Samsung Galaxy S23

### Android version

16

### Nextcloud Talk app version

Master

### Nextcloud server version

33

### Talk version

23

### Custom Signaling server configured

Yes (specify version in Additional Information)

### Custom TURN server configured

Yes

### Custom STUN server configured

Yes

### Android logs

00:15:03.899  onMessageEvent 'hello'
00:15:04.206  onMessageEvent 'roomJoined' currentCallStatus=CONNECTING  ← fires BEFORE joinRoomAndCall!
00:15:04.652  hasMCU is true
00:15:04.652  joinRoomAndCall                                              ← too late, stale roomJoined already consumed
The roomJoined event at 00:15:04.206 is stale — it was enqueued before joinRoomAndCall() was called at 00:15:04.652.

### Server log

```bash

```

### Additional information

**Root Cause**

The flow on app launch is:
1. CallActivity.onCreate() → CONNECTING state
2. WebSocket was already connected from a previous session (or ChatActivity had it open)
3. Server sends hello event → triggers initiateCall() → fetchSignalingSettings() → ...
4. Server also sends a stale roomJoined event from the previous WebSocket session
5. The stale roomJoined fires performCall() before joinRoomAndCall() has been called
6. Later, joinRoomAndCall() runs and calls WebSocketInstance.joinRoomWithRoomTokenAndSession()
7. The WebSocket sees it's already in the same room/session and does a local join without sending anything to the server
8. The server never re-sends the participant list because it believes the client is already in the room
9. handleCallParticipantsChanged() is never called with existing participants
10. State stays at JOINED, never transitions to IN_CONVERSATION

mahibi:
Please excuse the delayed answer, this one slipped through!

Thanks for the detailed description, we will have a closer look!

PR #6045, opened 2026-04-11
fix(call): ignore stale roomJoined events that fire before joinRoomndCall

tareko:
When the app launches and immediately joins an existing call, the WebSocket may already be connected from a previous session. The server sends a stale roomJoined event before joinRoomAndCall() is called, causing performCall() to fire prematurely. Later when joinRoomAndCall runs, the WebSocket does a local join and the server never re-sends the participant list. The call stays stuck in JOINED and never transitions to IN_CONVERSATION.

Add joinRoomInitiated flag to guard the roomJoined handler against stale events. Only process roomJoined after joinRoomAndCall has been called.

This fixes Issue #6044 . I used OpenCode and GLM-5.1 to debug and fix this bug.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -302,6 +302,7 @@ class CallActivity : CallBaseActivity() {
     private var externalSignalingServer: ExternalSignalingServer? = null
     private var webSocketClient: WebSocketInstance? = null
     private var webSocketConnectionHelper: WebSocketConnectionHelper? = null
+    private var joinRoomInitiated = false
     private var hasMCU = false
     private var hasExternalSignalingServer = false
     private var conversationPassword: String? = null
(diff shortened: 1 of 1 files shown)


mahibi:
Hi @tareko 
thanks for contributing! we will have a closer look soon!

mahibi:
@tareko are you interested to be invited to the nextcloud.com instance as a guest so we could also chat and you can further connect to the community? Should i invite you via tarek@tarek.org ?

Also, do you like to have advanced access to the talk repo? Then you could create PR's inside it instead in your fork, which will simplify the review process

mahibi:
Review is planned for this week

tareko:
> @tareko are you interested to be invited to the nextcloud.com instance as a guest so we could also chat and you can further connect to the community? Should i invite you via [tarek@tarek.org](mailto:tarek@tarek.org) ?
> 
> Also, do you like to have advanced access to the talk repo? Then you could create PR's inside it instead in your fork, which will simplify the review process

Thank you so much for the honour. I received the invitation and accepted.

mahibi:
The PR is on our list.

mahibi:
still on our list. Review is not forgotten

mahibi:

mahibi:
Apologies for the delay, i was out sick for a week and now try to catch up

mahibi (approved):

mahibi:
makes sense. thank you @tareko :+1:

Issue #6061, opened 2026-04-14
GIF playback broken


anakin78z:
### Steps to reproduce

1. Post a GIF

### Expected behaviour

GIF is animated in chat

### Actual behaviour

GIF is static

### Device brand and model

Samsung Galaxy S24 Ultra

### Android version

16

### Nextcloud Talk app version

master branch as of April 14 2026

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

### Additional information

It looks like this broke during the chatkit removal. 

Here is the commit that added it, for reference: #5969

anakin78z:
I should be able to fix this since I did the original implementation, if someone wants to assign me.

mahibi:
yes it was noted as TODO as followup in https://github.com/nextcloud/talk-android/pull/5635
feel free to implement it again 👍

PR #6047, opened 2026-04-12
Show file upload progress and placeholder media message in chat.

anakin78z:
Implements #2375 

I'm aware of the work in progress at #5317 but I couldn't get it to work when I tried it, and since there were a lot of changes with the ChatKit removal, I thought it best to start fresh.

### 🖼️ Screen captures:
image

The X button in the top right cancels the upload. 
Uploading images are blurred to indicate they are not yet fully sent (they remain blurred until the device receives the chat message from the server). The idea is to make it obvious that the messages can't be interacted with as a regular media message.

#### Video demo: 
https://youtube.com/shorts/FMw_yvhxXL4

### 🚧 ISSUES

I created issue #6081 for the following. It will be fixed once #6124 is merged: 

After a file has uploaded, but before the insurance request is made, tapping the image/video fails. While investigating this, I noticed that the return values from the WebSocket and the insurance request are different: 
image

The link and path parameters both have different values. I'm not sure if this is a server issue, or if the responses need to be handled differently, but it's odd that the link would be different.

My server version is 33.0.0

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [ ] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #6080, opened 2026-04-15
Auto play GIF media.

anakin78z:
Addresses #6061 
This PR reimplements GIF playback after the chatkit removal.

It also checks if the user has the Reduce/Remove animations accessibility setting enabled, and will not auto play GIF files if so.

### 🖼️ Demo

https://youtube.com/shorts/6SKiq43Svng

### 🚧 ISSUES

After posting or receiving a chat with a GIF, the file will not initially display. This is because of an issue with the WebSocket result not matching the REST insurance call result. I made a separate issue for this: #6081, which will be fixed once #6124 is merged

This will likely have a conflict with #6047 but that should be a quick and easy fix after one or the other gets merged.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ui/model/ChatMessageUi.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ui/model/ChatMessageUi.kt
@@ -7,6 +7,7 @@
 
 package com.nextcloud.talk.chat.ui.model
 
+import android.provider.Settings
 import androidx.compose.runtime.Stable
 import com.nextcloud.talk.R
 import com.nextcloud.talk.application.NextcloudTalkApplication.Companion.sharedApplication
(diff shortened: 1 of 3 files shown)


mahibi:
Great :+1: 
will review very soon

mahibi (approved):

anakin78z:
~~Based on the discussion in #6081 the issue with the GIF not displaying is due to it wanting to display the file, not the preview, for the animated GIF. When the WebSocket data comes in, we only have a valid preview url, but not a valid file url.~~

~~So either we need to somehow fall back to the preview URL or force the insurance call if the file request fails.~~ 

This will be fixed once #6124 is merged

anakin78z:
Fixed merge conflict

mahibi:
Thank you @anakin78z :+1:

Issue #6056, opened 2026-04-13
Hide image filename


anakin78z:
**Is your feature request related to a problem? Please describe.**
The app shows the image file name when no caption is added. This tends to look messy and (to me) ugly. The web app does not display file names for images, and neither do most chat apps.

**Describe the solution you'd like**
I'd like file names to not display. 

**Describe alternatives you've considered**
I understand that it could be useful for some work scenarios to see filename information such as version info. Perhaps there could be a setting to display or hide file names for media that has previews. 

**Additional context**
Android App | Web App
--- | ---
<img width="463" height="1002" alt="Image" src="https://github.com/user-attachments/assets/7cb89e8c-7082-40ba-a3b6-0b3c13d27682" /> | <img width="359" height="472" alt="Image" src="https://github.com/user-attachments/assets/cf727d0f-3b9e-4ffb-a61b-60bbf6e3b1e6" />

sowjanyakch:
Makes sense :+1: . Hiding the filename will make the UI look cleaner.

AndyScherzinger:
Yes, makes sense and is in line with @nimishavijay suggestion a while ago. 👍

PR #6082, opened 2026-04-15
Hide filenames for media in chat.

anakin78z:
Implements #6056 

Added setting to always show filenames.

### 🖼️ Screenshots

🏚️ Before | 🏡 After | ⚙️ Setting
---|---|---
Screenshot_20260415_142108 | Screenshot_20260415_142035 | Screenshot_20260415_143559

### 🚧 TODO

- [ ] Translate setting text, @string/nc_settings_show_media_filenames_title & @string/nc_settings_show_media_filenames_desc

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/chat/MediaMessage.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/chat/MediaMessage.kt
@@ -46,7 +46,13 @@ fun MediaMessage(
     conversationThreadId: Long? = null,
     onImageClick: (Int) -> Unit
 ) {
-    val captionText = message.message.takeUnless { it == FILE_PLACEHOLDER_MESSAGE }
+    val hasExplicitCaption = message.plainMessage != FILE_PLACEHOLDER_MESSAGE
+    val hasPreview = !typeContent.previewUrl.isNullOrEmpty()
+    val captionText = when {
+        hasExplicitCaption -> message.message
+        !hasPreview -> message.message
+        else -> null
+    }
     val hasCaption = captionText != null
     val mediaInset = 4.dp
     val mediaShape = remember(message.incoming) {


anakin78z:
I don't know how translations work for this app. Feel free to assign & resolve as needed.

AndyScherzinger:
Translations happen via transifex, so for development it is is main strings.xml only. Anything else gets overwritten daily.

Screenshots look nice. @nimishavijay for feedback especially regarding the config option, given we usually try to avoid them.

anakin78z:
My reason for adding a setting for it is that it can be useful for certain work scenarios. For example someone posting multiple versions of an image, with the version number in the filename. Yes, you can get it by clicking on the image, but it's sometimes truncated there.

jancborchardt (changes requested):
Very nice improvement @anakin78z !

As @AndyScherzinger mentioned, I would remove the setting as well and just hide the file names, as is done on other platforms.

For the truncation issue on multiple versions, we can solve it separately in the image viewer e.g. by middle ellipsizing the title.

mahibi:
+1 to avoid a setting for this. 

> For the truncation issue on multiple versions, we can solve it separately in the image viewer e.g. by middle ellipsizing the title.

Makes sense. Additionally we could also add a "info" item in the upper right corner menu of the image viewer to show the metadata (not in this PR).

anakin78z:
I removed the setting & pushed the changes.

mahibi (changes requested):
the behavior is not the same as on web (& other platforms).

As far as i see, for filetypes other than image and video, the filenames should always be shown.
For this, access to the mimetype would be necessary which is not the case in MediaMessage.kt.

But i would say it makes more sense to move the logic into getMediaContent in ChatMessageUi.kt so it's less logic in the UI. 
Wdyt @anakin78z ?

anakin78z:
Do other filetypes have previews? It only hides the filename if there is a preview. 

#6080 already passes the mimetype to MediaMessage, but for this usecase it could be done in ChatMessageUi.kt, or we could use `remember` in the MediaMessage.

mahibi:
The PR is on our list.

AndyScherzinger:
> Do other filetypes have previews?

To answer your question, yes they might. Which files have previews depends on the the preview providers configured on the server.

anakin78z:
Found an issue where you can't bring up the context menu for an image without having the text area to click. I plan on updating it so that a long press will bring up the context menu.

mahibi:
> Found an issue where you can't bring up the context menu for an image without having the text area to click. I plan on updating it so that a long press will bring up the context menu.

thanks, just ping us whenever it's ready for a review

mahibi:
Hi @anakin78z 
i copied your commit to https://github.com/nextcloud/talk-android/pull/6201 and fixed the longclick.
Did that so it's included in the RC1 for today, hope that okay for you. 
Thank you for your help! :+1: 
I will close the PR

PR #6087, opened 2026-04-16
fix: HPB partial updates no longer drop the call when a participant joins

### LINKED ISSUE #5452: Call started from talk android gets dropped when a participant joins
### Steps to reproduce

1. Start a group chat call from the talk-android app
2. In another account, join in the call

### Expected behaviour

Call is established between 2 participants

### Actual behaviour

Call is dropped by the talk android app, leaving only the other participant in the call

### Device brand and model

Samsung Galaxy A16 5G

### Android version

15

22.0.2

### Nextcloud server version

32

### Talk version

22

Yes (specify version in Additional Information)

Yes

No

### Android logs

_No response_

### Server log

```bash

```

### Additional information

_No response_

alauzon:
Fixes #5452

## Problem

With external signaling (HPB), `onParticipantsUpdate` delivers **partial updates** containing only changed participants. The previous code treated both `onUsersInRoom` (full list) and `onParticipantsUpdate` (partial update) identically: any participant absent from the list was moved to the `left` collection.

When a new participant joined via HPB, the local user was **not included** in the partial update (because the local user didn't change). This caused the local participant to appear in `left`, which made `CallActivity.handleCallParticipantsChanged` conclude:

```kotlin
isSelfInCall = false
// → "Most probably a moderator ended the call for all."
hangup(shutDownView = true, endCallForAll = false)
```

The call was dropped by the Android client as soon as another participant joined.

## Fix

`processParticipantList` now takes a `boolean fullList` parameter:

- `onUsersInRoom` passes `fullList=true` — absent participants have left (original behaviour)
- `onParticipantsUpdate` passes `fullList=false` — absent participants are still in the call and are moved to `unchanged` instead of `left`

`handleCallParticipantsChanged` builds `participantsInCall` from `joined + updated + unchanged`, so the local user stays present, `isSelfInCall` remains `true`, and the call is not dropped.


mahibi:
thank you @alauzon we will have a closer look soon

mahibi:
The PR is on our list.

mahibi:

mahibi:
Apologies for the delay, i was out sick for a week and now try to catch up

mahibi:
i was still not able to reproduce but the code makes sense to handle this. Thank you @alauzon :+1:

PR #6092, opened 2026-04-20
Fix call state timing regression causing delayed audio routing

alexey22p:
## Summary

This patch restores the original call initialization timing from 22.0.3.

In 22.1.x `prepareCall()` was moved later behind permission checks and foreground service startup.  
As a result `currentCallStatus = IN_CONVERSATION` may be set later than before.

On some devices this can delay audio routing, speakerphone activation, or other call startup behavior.

## Root cause

Regression introduced during call startup refactoring in 22.1.x.

## Fix

Move `prepareCall()` earlier to match the previous 22.0.3 behavior.

## Testing

Built locally and verified startup sequence changes.

Full push-notification based incoming-call testing could not be completed in the local build environment because Google push services were unavailable.

Regression comparison performed between:
- v22.0.3
- v22.1.0rc1

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -1016,6 +1016,12 @@ class CallActivity : CallBaseActivity() {
         checkRecordingConsentAndInitiateCall()
 
         if (permissionUtil!!.isMicrophonePermissionGranted()) {
+            // Fix regression introduced with ForegroundService flow:
+            // ensure any incoming ringtone/calling sound is stopped
+            // BEFORE promoting call to foreground service.
+            // Prevents ringtone continuing after answering call.
+            stopCallingSound()
+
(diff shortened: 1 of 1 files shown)


mahibi:
Thank you @alexey22p we will have a closer look

mahibi:
Thank you @alexey22p 
It makes sense codewise to call the stopCallingSound() in prepareCall()
I am not able to reproduce the bug but codewise it makes sense and the change will be taken.

However i created https://github.com/nextcloud/talk-android/pull/6103 and moved the stopCallingSound() outside of the `if (permissionUtil!!.isMicrophonePermissionGranted())` check.

Hope that's okay for you, it's just done in a separate PR as i want to move quickly forward.
Thanks again for the pointer! :+1:

alexey22p:
Thanks, Mahibi, that makes sense.

Moving stopCallingSound() separately is a cleaner minimal fix.
Glad the investigation helped identify the right spot.

PR #6108, opened 2026-04-23
Add message Markdown support

### LINKED ISSUE #6051: markdown lists aren't rendered

AndyScherzinger:
Resolves #6051

### 🚧 TODO

- [ ] review

### 🏁 Checklist

- [x] 📅 Milestone is set


mahibi:
I will merge this as it's working and improves the UX.

However there need to be followup improvements:

The code

```
@Composable
fun EnrichedText(message: ChatMessageUi, modifier: Modifier, maxLines: Int = Int.MAX_VALUE) {
    if (message.renderMarkdown) {
        MarkdownText(
             ...
        )
    } else {
        MentionEnrichedText(
               ...
        )
    }
}
```

currently acts as a switch between two worlds:

- On one hand MentionEnrichedText which uses Compose only but does not support markdown
- On the other hand MentionEnrichedText which uses AndroidView (which should be avoided) and supports Markdown 

Both Composables solve the same things (like chip mention etc) in a redundant way.
This should be unified. A switch may not be necessary in the end.

AndyScherzinger:
Yeah, I agree extracting the chip handling would be good. The current issue is that markwon is not having composable support, hence the typical work around.

Best case would be if they have composables at some point and we could just switch to that 🙏

AndyScherzinger:
On the other hand we could also just have the markwon variant for both cases but than have to live with the non-composable on the inside 😔

PR #6109, opened 2026-04-23
Improving voice messages

### LINKED ISSUE #6093: further improve voice messages
followup to https://github.com/nextcloud/talk-android/pull/6033

- [ ] it would be nice if the middle center of the circle would be the marker to color the vertical lines. Right now it seems to be the right border of the circle

   grafik

- [ ] would be nice if the message time is shown in the same line as the playback time. 
In theory this should be possible by passing `forceTimeBelow = false ` to MessageScaffold, but it seems this is not working. Please set it to false and check why it's not working.

- [ ] when starting to play another voice message while the first one is played back, the pause button of the first one should be replayed with the play button

   grafik

- [ ] for the old chatkit code, it was implemented that voice messages play automatically after the prior one was finished. Not a high priority and maybe we should discuss this in the team to align with all platforms. Just mentioning this is not implemented for now.

rapterjet2004:
- fixes #6093 

### 🚧 TODO

> 
> - [x] when starting to play another voice message while the first one is played back, the pause button of the first one should be replayed with the play button
> - [x] it would be nice if the middle center of the circle would be the marker to color the vertical lines. Right now it seems to be the right border of the circle
> 

- [ ] debug playButton state issue

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi (changes requested):
i rebased on master an d fixes the conflicts.

I see this behavior:
sometimes it happens that the play button of one message controls the playback in another message

mahibi (approved):

Issue #6081, opened 2026-04-15
WebSocket and REST calls receive different results for chat messages with files


anakin78z:
### Steps to reproduce

1. Upload a file or receive a media message
2. Check the logs with the WebSocketInstance TAG for the received message, starting with Receiving : 
3. Add a log to the insurance request and compare the result. 

### Expected behaviour

The data received should be identical

### Actual behaviour

The file `link` and `path` parameters are different. 
This causes the app to be unable to load the media until the insurance request completes.

<img width="671" height="521" alt="image" src="https://github.com/user-attachments/assets/a06803e5-2d8d-4ce2-a0d3-39a7cfe83110" />

### Device brand and model

any

### Android version

16

### Nextcloud Talk app version

master branch as of 4/15/2026

### Nextcloud server version

33.0.0

### Talk version

23.0.3

### Custom Signaling server configured

None

### Custom TURN server configured

None

### Custom STUN server configured

None

### Server log

```bash

```

nickvergessen:
This is expected behaviour, just lacking documentation at this point.
The WebSocket is sent 1-to-many and therefore the file data needs to be generic so it somewhat works for everyone.
Rest is 1-to-1 and therefore knows if a user has direct access and knows the local path for each user.

anakin78z:
@nickvergessen I realize now I didn't add this info to the issue, but the app is unable to display media from the WebSocket data. E.g. an image request will fail, and the user will see the placeholder image until the insurance request completes. 

So either the WebSocket data needs to be processed (adding `Talk/` to the path), or the data is used incorrectly somehow.

nickvergessen:
> E.g. an image request will fail, and the user will see the placeholder image until the insurance request completes.

That is the intended idea for now

> So either the WebSocket data needs to be processed (adding Talk/ to the path)

While true for most shares, this is not true for all files you receive as a share in Talk. Not every item is in the Talk attachment folder

anakin78z:
@nickvergessen I'm sure there's parts of the architecture that I don't yet understand, but at the moment this behavior causes issues with #6047 and #6080. If you could have a look at those and chime in, I'd appreciate it.

mahibi:
I guess the confusion is this:
Whenever a websocket message about a file comes in, the insurance request must be triggered (instead to wait for the regular timer based insurance request). The other platforms already do it like this.
This is not yet implemented and needs to be done (i was not aware of this behavior).

mahibi:
[...]

Issue #6064, opened 2026-04-15
modify system messages handling for chat signaling


mahibi:
there needs to be a special system messages handling regarding signaling

- [ ] handle and translate the important ones

- [ ] request server via insurance request for others instead to handle the message

ref https://github.com/nextcloud/spreed/blob/fd1807d0017d736b3d058a830f720defc670a6a0/lib/Signaling/Listener.php#L80-L101
ref https://github.com/nextcloud/spreed/blob/bf0f5f05e42b9d94449970c09836978650b1a514/lib/Chat/Parser/SystemMessage.php#L118

anakin78z:
Partially implemented in #6124

PR #6124, opened 2026-04-25
Update WebSocket data handling for files and system messages

anakin78z:
See #6081 

When receiving a WebSocket chat comment with a file attachment, fetch messages via REST call to get user specific file paths. 
This allows the user to display media without needing to wait on the insurance call to trigger. 

Also checks for system messages that need to be localized for the user. This follows the logic of the web app. However, since it doesn't look like we do localization in the Android app, it's set to fetch new data for any system message that is not on the list of untranslated system messages. Partially implements #6064

### 🏁 Checklist

- [ ] 📅 Milestone is set


mahibi:
will have a look soon!

mahibi:
still on our list. Review is not forgotten

mahibi:

mahibi (approved):

mahibi:
Thank you @anakin78z :+1:

PR #6128, opened 2026-04-27
add fullscreen permission setting

mahibi:
Add option for fullscreen permission in settings.

---

Necessary that the call notification screen is shown without to unlock the phone.

I encountered the following behavior:

- USE_FULL_SCREEN_INTENT is granted when running via debug in Android Studio.

- It is not granted when building the app and running the apk. So the unlock screen is shown instead when a call comes in.

- However google play console has a setting to allow USE_FULL_SCREEN_INTENT by default for this app, as it is a call app. Hopefully this will work in v23.0.1 and USE_FULL_SCREEN_INTENT is granted when installed via Google Play.

The setting is especially useful for users who don't install via gplay.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ConversationsListActivity.kt
@@ -8,6 +8,7 @@ package com.nextcloud.talk.conversationlist
 
 import android.Manifest
 import android.annotation.SuppressLint
+import android.app.NotificationManager
 import android.content.ActivityNotFoundException
 import android.content.Intent
 import android.content.pm.PackageManager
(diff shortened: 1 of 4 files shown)


mahibi:
Created this PR when i was testing v23.0.1 and was wondering why the lockscreen still appeared despite of it being changed in https://github.com/nextcloud/talk-android/pull/6085/commits/207f0a117ee3fa79dada954f470183fd356f4d5e.

mahibi:
Tested with 23.0.1 and confirm USE_FULL_SCREEN_INTENT is granted by default via play store install.

Anyway we might have to use the fullscreen setting for the generic build (as the permission might not be granted by default when not installing via playstore).

mahibi:
Also keep in mind when testing this with a fresh install via gplay: For the initial first time call the lockscreen may appear. This is not because of a missing fullscreen permission but for the other permissions that are required for the call. Only needs to be done one time when granting this for future uses.

PR #6129, opened 2026-04-27
prevent scroll jump to newest message while loading older history

### LINKED ISSUE #6022: sometimes jumping to newest message when scrolling up
- known bug from https://github.com/nextcloud/talk-android/pull/5635

The chat sometimes jumps to the newest message when scrolling up

random ideas: 
- Check the flows
- empty list passed?
- ViewCompositionStrategy ??

mahibi:
fix #6022

When older messages were loaded, connected chat blocks were merged by deleting old blocks and inserting a new one in separate DAO operations. This created a state with no latest chat block, so the message observer emitted an empty list (only for a very short time) and the UI jumped to the newest message because it lost the anchor.

To fix this chat blocks are now replaced atomically in a single transaction. This removes the transient empty state and keeps the current scroll position stable.

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/talk/data/database/dao/ChatBlocksDaoTest.kt
+++ b/app/src/androidTest/java/com/nextcloud/talk/data/database/dao/ChatBlocksDaoTest.kt
@@ -22,6 +22,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.test.runTest
 import org.junit.After
 import org.junit.Assert.assertEquals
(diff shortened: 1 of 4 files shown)


mahibi:
please test by scrolling far into the the past. -> chat must never jump to the newest message

mahibi:
i successfully scrolled multiple very long conversations up to the "created the conversation" system message

sowjanyakch (approved):

PR #6152, opened 2026-05-02
fix(upload): fix attachment sharing failure due to chunked upload filecache bug

IIMacGyverII:
## Summary

Sending images/files in Talk chat silently fails — the upload completes but the file never appears in the conversation for either the sender or recipient. This affects all self-hosted Nextcloud servers and has been reported across multiple server versions.

## Steps to Reproduce

1. Open any Talk conversation
2. Tap the attachment button and select a photo (> 1 MB) from the gallery
3. Observe: no image appears in chat; no error is shown to the user

## Expected Behavior

The image is uploaded to Nextcloud files and shared into the conversation as a message attachment.

## Actual Behavior

The upload silently succeeds internally but the file never appears in chat for sender or recipient.

## Root Cause

Nextcloud's chunked upload assembly is broken on affected server versions. The final WebDAV `MOVE` (from `/remote.php/dav/uploads/<user>/<uuid>/.file` to `/remote.php/dav/files/<user>/Talk/<filename>`) returns HTTP **201 Created**, but the server silently fails to register the file in its `oc_filecache` database. A `HEAD` request immediately after the "successful" MOVE returns **404** — the file doesn't exist as far as Nextcloud's APIs are concerned.

When the app then calls the OCS sharing API to share the file into the Talk room, it gets:
```
HTTP 404 — "Wrong path, file/folder does not exist"
```

The bug only manifests via chunked upload. Direct `PUT` upload correctly updates the filecache.

Network trace confirming the issue:
```
MKCOL .../uploads/<user>/<uuid>                              → 201
PUT   .../uploads/<user>/<uuid>/0*                           → 201 (all chunks)
MOVE  .../uploads/.../.file → .../files/<user>/Talk/photo.jpg → 201 ← claims success
HEAD  .../files/<user>/Talk/photo.jpg                        → 404 ← not in filecache
POST  /ocs/v2.php/.../shares  path=/Talk/photo.jpg           → 404 "Wrong path"
```

## Fix

Raise `CHUNK_UPLOAD_THRESHOLD_SIZE` from 1 MB to 20 MB in `UploadAndShareFilesWorker`. Files under 20 MB (virtually all phone photos) now go through direct `PUT` upload, which correctly updates the filecache. Chunked upload is preserved for files over 20 MB.

Additional hardening:

- **`ChunkedFileUploader`**: Build a fresh `OkHttpClient` that does not inherit `OCS-APIRequest: true` / `Accept: application/json` from the app's singleton client. Those headers were causing sabre/dav to return JSON `200` instead of XML `207 Multi-Status` for `PROPFIND`, breaking dav4jvm. `PROPFIND` failures now fall back gracefully to re-uploading all chunks.
- **`FileUploader`**: Fix `createRequestBody()` to use `readBytes()` instead of `available()`, which returns 0 for `content://` URIs and caused the PUT body to be empty.
- **`ShareOperationWorker`**: Retry share creation up to 4 times with a 2 s delay on 404, as a safety net for large files still going through chunked upload.

## Affected Versions

Reproducible on self-hosted Nextcloud regardless of server version (confirmed on multiple servers). iOS and web Talk clients are not affected because they do not use chunked upload for typical-sized attachments.

### 🖼️ Screenshots

Not applicable — the fix is upload/share logic with no UI changes.

### 🏁 Checklist

- [x] ⛑️ Tests — not needed (bug is server-version-specific; requires a live Nextcloud server to reproduce)
- [x] 🔖 Capability — not needed (no capability gating required)
- [ ] 📅 Milestone is set
- [x] 🌸 PR title is meaningful

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/jobs/ShareOperationWorker.kt
+++ b/app/src/main/java/com/nextcloud/talk/jobs/ShareOperationWorker.kt
@@ -26,6 +26,7 @@ import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_INTERNAL_USER_ID
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_META_DATA
 import com.nextcloud.talk.utils.bundle.BundleKeys.KEY_ROOM_TOKEN
 import io.reactivex.schedulers.Schedulers
+import retrofit2.HttpException
 import javax.inject.Inject
 
 @AutoInjector(NextcloudTalkApplication::class)
@@ -46,6 +47,16 @@ class ShareOperationWorker(context: Context, workerParams: WorkerParameters) : W
 
     override fun doWork(): Result {
         for (filePath in filesArray) {
+            tryCreateShare(filePath)
+        }
+        return Result.success()
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun tryCreateShare(filePath: String?) {
+        for (attempt in 1..SHARE_MAX_ATTEMPTS) {
+            var succeeded = false
+            var shouldRetry = false
             ncApi.createRemoteShare(
                 credentials,
                 ApiUtils.getSharingUrl(baseUrl!!),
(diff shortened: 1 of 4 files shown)


IIMacGyverII:
Just to clairfy, this bug has been preventing me from uploading photos or files on my android devices to other Nextcloud Talk apps and the web browser for years across multiple server installs.  I was never able to figure it out until now.  iOS and the web browser with Nextcloud Talk always worked fine but android never worked for me.  This fixes it.

mahibi:

mahibi:
Thank you @IIMacGyverII  :+1: 

While reviewing i stumbled over some more things regarding the uploads and also needed to rebase on latest master ...so this led to the fact that i branched off and put more changes on top of your PR. 
The new PR is https://github.com/nextcloud/talk-android/pull/6183
Hope that's okay for you. The commit from you is included but some things are changed afterwards (Esp. there is no new OkHttpClient.Builder() introduced but the existing one is modified to only include `OCS-APIRequest: true / Accept: application/json` when it is necessary).

Thank you for contributing!
I suggest to close this PR. Feel free to comment on https://github.com/nextcloud/talk-android/pull/6183

IIMacGyverII:
Completely fine with me!  Thanks!

PR #6182, opened 2026-05-12
Bugfix/noid/fix system message design

mahibi:
fix system message design with mention chips

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🏁 Checklist

- [ ] 📅 Milestone is set


AndyScherzinger (approved):
Works for me an dis an improvement. Implementing the copilot remarks seems to make is worse not better

PR #6213, opened 2026-05-17
Improve mark as read logic

### LINKED ISSUE #6059: After leaving a conversation (while it was scolled to the bottom), it isn't always marked as read in the conversation list

AndyScherzinger:
Resolves: #6059

According to our friend Claude:

The mark-as-read effect had two independent bugs:
  
- If the chat opened with the scroll position already at the bottom (e.g. only one unread message), or a new message arrived while the user was already at the bottom, the effect never re-ran since it only triggered on a  change of the "at newest" state — leaving messages unmarked as read.
- If the first item in the list was a non-message entry (such as a date separator), the index-based lookup silently bailed out without advancing the read position at all.
  
The fix makes the effect re-evaluate whenever the message list changes, and replaces the index-based item lookup with a search that skips non-message entries and reliably resolves the newest actual message.

### 🚧 TODO

- [x] test & review for a while

### 🏁 Checklist

- [ ] 📅 Milestone is set

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/ui/chat/ChatView.kt
+++ b/app/src/main/java/com/nextcloud/talk/ui/chat/ChatView.kt
@@ -302,15 +302,13 @@ fun ChatView(
         }
     }
 
-    LaunchedEffect(isAtNewest) {
+    LaunchedEffect(isAtNewest, state.chatItems) {
         if (!isAtNewest) return@LaunchedEffect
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Merging in alignment, to have it tested, works well for me, maybe a bit to well in marking as read without having scrolled to the bottom completely.

PR #5660, opened 2025-12-20
Implement persistent foreground service to keep calls active in background, with notification controls for managing the call

tareko:
This PR is in response to Issue #1688 . In summary, if PIP is not available, the call ends and this should not happen. I used ChatGPT Codex, Claude, and localAI via Qwen3-coder to generate the code in this PR. Within my personal programming abilities, I still believe the code is elegant. I've used the test app for several calls with no issues on a P6P with latest android.

This PR implements two key functions:
1. It allows the call to remain active even if PIP is not enabled and it is in the background
2. It creates a persistent notification that:
   a. Allows users to return to the call
   b. Allows users to end an active call directly from the persistent notification without needing to return to the CallActivity

## Features
- Add CallForegroundService with persistent notification
- Support calls in background without requiring picture-in-picture mode
- Add "Return to call" and "End call" action buttons to CallForegroundService notification with corresponding PendingIntent
- Handle proper foreground service types for microphone/camera permissions
- Add notification permission and fallback messaging.
- Add EndCallReceiver to handle end call broadcasts from notification action
- Use existing ic_baseline_close_24 drawable for end call action icon
- Register broadcast receiver in CallActivity to handle end call requests from notification using ReceiverFlag.NotExported for Android 14+ compatibility
- Add proper cleanup flow: notification action → EndCallReceiver → CallActivity → proper hangup sequence
- Track intentional call leaving to prevent unwanted service restarts
- Release proximity sensor lock properly during notification-triggered hangup
- Add diagnostic logging throughout the end call flow for debugging

The implementation follows Android best practices:
- Uses NotExported receiver flag for internal app-only broadcasts
- Properly unregisters receivers in onDestroy to prevent leaks
- Uses immutable PendingIntents for security
- Maintains proper state management during call termination

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 🔖 Capability is checked or not needed 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #6334, opened 2026-06-11
Bugfix/1688/fix call background death

### LINKED ISSUE #1688: continue call when minimizing app and PIP is not available/enabled

mahibi:
- continuation of https://github.com/nextcloud/talk-android/pull/5660 (credits go to @tareko )

- fix https://github.com/nextcloud/talk-android/issues/1688

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


backfield79:
> APK file: https://github.com/nextcloud/talk-android/actions/runs/27347821297/artifacts/7570066996
> To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.
> qrcode (please click on link to get QR code displayed)

Hello! 

Tried to download APK to test, but there is no page behind the link: "Didn’t find anything here!"

mahibi:
> Hello!
> 
> Tried to download APK to test, but there is no page behind the link: "Didn’t find anything here!"

Thanks for the hint, will have a look at the CI process and check whats going wrong

mahibi:
I rebased on master and fixed merge conflicts.
The PR is not forgotten but giving it a quick test i immediately ended up with a disturbing bug: Starting a call ended up in two app instances. I heared the ring (coming from the other instance, which started the calll) but the instance that i came from did noting after clicking the call icon.
If this branch gets picked up: Be careful and test/fix! Do not yolo merge!

tareko:
@mahibi I picked this up and addressed the bug you found plus the remaining review points with the commit I just made. As with all my other work, this is heavily AI assisted, though this time with Kimi K3.

The "two app instances" bug had three root causes:
1. excludeFromRecents had been removed from CallActivity, so the call task (which is singleTask with its own taskAffinity) showed up as a second Talk instance in recents. Restored it — the way back to a backgrounded call is the persistent notification from CallForegroundService, which is exactly what this PR adds.
2. CallActivity never overrode onNewIntent. That was harmless before because the activity was always finished when you left it, but now that moveTaskToBack() keeps it alive, tapping the call icon again routed the intent to the stale backgrounded instance and dropped it — that's why the instance you were looking at "did nothing" while the other one kept ringing. onNewIntent now handles this: same room → bring the existing call to the front; different room → hang up and hand off to a fresh instance so no stale state is reused.
3. onDestroy kept the mic stream, signaling listeners and receivers alive on "system-initiated destroy" while simultaneously stopping the foreground service — a half-dead ghost call. Since WebRTC state can't survive the activity being destroyed anyway, onDestroy now always cleans up and hangs up. Background survival still works because moveTaskToBack/PiP don't destroy the activity.
Also fixed while in there: the orphaned voice-room navigation branch in hangupNetworkCalls, a dead `else` that would have called prepareCall() twice, the notification's "Return to call" losing its extras during timer updates, the misleading "permission was denied" logs from your earlier review (camera/bluetooth/notification this time), and a stuck isLeavingRoom flag in ChatActivity if a leave-room request failed.

Builds clean, `detekt/ktlintCheck` pass, and the existing `CallBaseActivityPipTest`/`ChatActivityLeaveRoomLifecycleTest` pass. I tested on-device with the QA build.

I'm happy to adjust anything, and hopefully this helps get us closer to the finish line.

mahibi:
> @mahibi I picked this up and addressed the bug you found plus the remaining review points with the commit I just made. As with all my other work, this is heavily AI assisted, though this time with Kimi K3. ...

Thank you @tareko 
The PR won't be forgotten although there is a delay with reviewing. It's quite important for me to improve the behavior this PR addresses, thank you to keep it up :+1: 
Once things calm down with the upcoming release i plan to continue here!

mahibi:
I finally found the time to test and enhance the PR.
I tested a lot and it works perfectly for me now.
As solving the merge conflicts was a challenge and i don't want to do this again i will merge before this happens another time (25.0.0 will be released next week without this PR. So there is plenty of time to fix things until this goes out into the wild).

Thank you that you kept improving and rebasing the PR @tareko :+1: :100:

tareko:
Thank you so much for your hard work on this! It was a really tough bug to work on that is at the same time very important. Well done! :tada:

backfield79:
This fix is so much wanted... Hope it will be done some day

PR #6373, opened 2026-06-19
Feature/1376/logging and reporting

mahibi:
This PR will add:
- file logging
- view logfiles in settings
- option to enable advanced logging (Level.DEBUG) or disable advanced logging (Level.INFO)
- crash screen to catch unhandled exceptions
- Share logfile / crash report via email or download as ZIP file

This PR does not yet replace all Log command with the new logger. This should be done in separate PRs while also cleaning up unnecessary Log commands. 

#### Settings screen

grafik

grafik

#### Logs screen

grafik

grafik

#### Crash reporting screen

grafik

### Testing

To test the crash screen, just add

`throw RuntimeException("Test crash")`

somewhere in the code and trigger it.

### 🚧 TODO

- [ ] set predefined mail address by configs
- [ ] create own email address for talk android
   - [ ] set it as predefined mail 

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/logger/FileLogHandler.kt
+++ b/app/src/main/java/com/nextcloud/talk/logger/FileLogHandler.kt
@@ -1,0 +1,96 @@
+/*
+ * Nextcloud Talk - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package com.nextcloud.talk.logger
+
+import java.io.File
+import java.io.FileNotFoundException
+import java.io.FileOutputStream
+import java.io.IOException
+
+class FileLogHandler(private val logDir: File, private val logFilename: String, private val maxSize: Long) {
+    data class RawLogs(val lines: List<String>, val logSize: Long)
+
+    companion object {
+        const val ROTATED_LOGS_COUNT = 3
+    }
+
+    private var writer: FileOutputStream? = null
+    private var size: Long = 0
+
+    private val rotationList = listOf(
+        "$logFilename.2",
+        "$logFilename.1",
+        "$logFilename.0",
+        logFilename
+    )
+
+    val logFile: File get() = File(logDir, logFilename)
+
+    val isOpened: Boolean get() = writer != null
+
+    fun open() {
+        try {
+            writer = FileOutputStream(logFile, true)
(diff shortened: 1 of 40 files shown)


app/src/main/java/com/nextcloud/talk/logger/FileLogHandler.kt
```diff
@@ -0,0 +1,96 @@
...
+    private var size: Long = 0
+
+    private val rotationList = listOf(
+        "$logFilename.2",
+        "$logFilename.1",
+        "$logFilename.0",
+        logFilename
+    )
+
+    val logFile: File get() = File(logDir, logFilename)
```
rapterjet2004, line 31: This works right now, as FileLogHandler is only instantiated in the modules and logFilename is always "nc_talk_log.txt", but if we want the fileNames to be mutable or be custom then this should be sanitized to always be within the `context.filesDir` to avoid fileNames with '..' or '../' 

Alternatively, if we know that the logFilename and logDir will never change, then they should be removed from the parameters of the class constructor and hardcoded in a companion object instead.

rapterjet2004 (changes requested):
clicking on Invalid URLs don't lead to the crash screen

Image

Also text and logged text can grow large really quickly. I didn't look at the growth rate too much, but is there any limits to how much text can be logged before it gets overwritten with new logged text?

mahibi:
> clicking on Invalid URLs don't lead to the crash screen

works fine for me. Could you try to debug why it fails for catch it for you?
(btw the crash itself is fixed now in https://github.com/nextcloud/talk-android/pull/6392)

> Also text and logged text can grow large really quickly. I didn't look at the growth rate too much, but is there any limits to how much text can be logged before it gets overwritten with new logged text?

LOG_FILE_MAX_SIZE is 1MB and we have 4 files which rotate
I suggest we go with this and still can modify in the future if necessary

app/src/main/java/com/nextcloud/talk/logger/FileLogHandler.kt
mahibi: extracted the filename to companion object :+1:

rapterjet2004 (approved):
I'm getting a build error in NotificationWorker:171

`logger.d(TAG, "pushMessage.notificationIds: " + pushMessage.notificationIds.contentToString())` contentToString not recognized, I changed it toString and everything works

Issue #111, opened 2018-03-01
Screens overlap


fantasio31:
Hi,

I have an issue when navigating between scrren. screens are overlapped.
Here is the scenario:
 1) Open Nextcloud Talk android App
 2) go to Parameters view
 3) click on "private life" (i have it in french so it is "Vie privée") ==> a popup ask me to choose the application to open the link
 4) Click on "Cancel" (it's "Annuler" on my French Xiaomi phone) ==> the popup is closed
 5) click on "Contacts" to open the contact view ==> Parameters and Contacts screens are ovelapped.

My configuration:
 - phone: Xiaomi Redmi 3 Pro with MIUI 9 (8.2.8.1 Beta) / Android 5.1.1
 - NextCloud version : 1.0.14

mario:
Hm? Can you show me the screenshot where this "private life" is?

fantasio31:
Here are the screens:
 - One with "Vie Privée"
 - The other one showing the overlap
![screenshot_2018-03-02-11-03-35-665_com nextcloud talk2](https://user-images.githubusercontent.com/23166621/36893924-16b02a84-1e0a-11e8-95cc-5f9262d99fc0.png)
![screenshot_2018-03-02-11-03-46-140_com nextcloud talk2](https://user-images.githubusercontent.com/23166621/36893925-16e31142-1e0a-11e8-9166-3bc47e9ec894.png)

mario:
Thanks, I'll investigate :)

mario:
Could you try the beta please? I couldn't reproduce there: https://play.google.com/apps/testing/com.nextcloud.talk2

fantasio31:
Same behavior. 

Le 3 mars 2018 19:19:00 GMT+01:00, "Mario Đanić" <notifications@github.com> a écrit :
>Could you try the beta please? I couldn't reproduce there:
>https://play.google.com/apps/testing/com.nextcloud.talk2
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/nextcloud/talk-android/issues/111#issuecomment-370168460

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

mario:
I'll need to test this on Android 5, since I can't reproduce at all :(

mario:
@xXSTrikeXx can you reproduce this?

xXSTrikeXx:
[...]

PR #6375, opened 2026-06-21
Fix #6169: elect call offerer using same-namespace session IDs

### LINKED ISSUE #6169: Call deadlocks in HPB mode when local & remote session IDs are compared across namespaces
### Bug summary

In `PeerConnectionWrapper`, the offer/offeree role is decided by a lexicographic
string comparison between the remote session ID (received over the signaling
channel) and the local session ID. In HPB (external signaling) mode these two
IDs are not in the same namespace, which makes the comparison meaningless and
can cause both peers to compute "I am NOT the offerer" simultaneously. No offer
is ever sent, no answer is ever produced, and the call hangs silently.

### Steps to reproduce

1. Configure a Nextcloud instance with the standalone signaling server (HPB).
2. Start a P2P call (no MCU) between two Talk Android clients.
3. Repeat until both peers happen to compute the same direction for `compareTo`
   (depends on the random session IDs generated by NC and HPB on each join).

The deadlock is intermittent because session IDs are random — but the
underlying logic guarantees a non-zero failure rate per call.

### Expected behaviour

For every successful peer pairing, exactly one side becomes the offerer and
sends a WebRTC offer; the other side answers; media flows.

### Actual behaviour

In HPB mode, both peers can simultaneously decide they are the offeree. No
offer is sent. The call stays in "connecting" state until the user gives up.

### Versions

- **Talk Android app**: `master` — bug present in this code path since commit
  [`b98bb93e4`](https://github.com/nextcloud/talk-android/commit/b98bb93e4)
  (2017-11-29). All released versions are affected.
- **Nextcloud server**: 33.0.3 (also affects all earlier versions where HPB is
  supported).
- **Talk (spreed) server app**: 24.0.0-dev.3.
- **Standalone signaling server (HPB)**: v2.1.1.
- **Custom Signaling server configured**: Yes.
- **Custom TURN server configured**: Yes.
- **Custom STUN server configured**: Yes.

### Code references

`PeerConnectionWrapper.java` decides offerer/offeree at construction:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java#L119

```java
boolean hasInitiated = sessionId.compareTo(localSession) < 0;
```

The `localSession` argument comes from `CallActivity.createPeerConnectionWrapper`:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt#L2390

```kotlin
return PeerConnectionWrapper(
    ...
    sessionId,
    callSession,    // ← Nextcloud session ID, even in HPB mode
    ...
)
```

`callSession` is set from `ApplicationWideCurrentRoomHolder.session`
(NC session ID, ~16 char hex token):

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt#L1534

The remote `sessionId` parameter, however, is the **HPB session ID** in HPB
mode. It is set from the WebSocket envelope `senderWebSocketMessage.sessionId`:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt#L184

```kotlin
ncSignalingMessage.from = callWebSocketMessage.senderWebSocketMessage!!.sessionId
```

The HPB session ID is a 58–70 char base64url string set in the
`hello` response:

https://github.com/nextcloud/talk-android/blob/master/app/src/main/java/com/nextcloud/talk/webrtc/WebSocketInstance.kt#L324

### Why the comparison can deadlock

In standard internal signaling mode, both `sessionId` and `localSession` are
NC session IDs (same namespace, same length, same character set), so
`compareTo` defines a total order — exactly one side computes `< 0`.

In HPB mode, `sessionId` is the HPB session ID (base64url, 58–70 chars) and
`localSession` is still the NC session ID (~16 char hex). The two strings live
in different namespaces with different alphabets and lengths. Lexicographic
ordering between them carries no semantic meaning. There exist pairs of IDs
where:

- Peer A computes `remote_HPB.compareTo(local_NC_A) >= 0` ("A is not initiator")
- Peer B computes `remote_HPB.compareTo(local_NC_B) >= 0` ("B is not initiator")

Both peers wait for an offer that never arrives. The fix `processParticipantList`
in nextcloud/spreed#6087 prevents the call from being torn down on the partial-update path,
but it does not unblock the offer/answer deadlock — that is a separate symptom
of the same underlying contract violation.

### Suggested fix

Pass the HPB session ID (i.e. `webSocketClient.sessionId`) instead of
`callSession` to `PeerConnectionWrapper` when external signaling is enabled.
Both sides then compare in the same namespace.

A more defensive approach would be to rename the parameter to
`localSignalingSessionId` and document that it MUST be in the same namespace
as the remote `sessionId`.

### Related

- nextcloud/spreed#2679 — similar symptom (silent deadlock, no offer sent), different root
  cause (peer not publishing media on the receiving side).

alauzon:
# Fix #6169: elect call offerer using same-namespace session IDs

Fixes #6169.

## Problem

In **High-Performance Backend** (external signaling) mode, 2-party calls can deadlock:
neither client creates the WebRTC offer, no media flows.

Root cause: the offerer is elected by a lexicographic comparison of the local and remote
session IDs (`remoteSession.compareTo(localSession) < 0`). This yields exactly one offerer
**only if both peers compare the same ordered pair of strings**. But at the
`PeerConnectionWrapper` construction site, the **remote** `sessionId` is a *signaling-layer*
session ID while the **local** session passed in (`callSession`) is the *Nextcloud* session ID.
The two peers therefore compare unrelated strings and can both compute the same role:

- both "not offerer" → nobody offers → **deadlock**, or
- both "offerer" → offer collision.

The `requestoffer` fallback only fires in MCU mode, so a direct (no-MCU) 2-party call deadlocks
outright. Because the session IDs are effectively random per call, the failure is intermittent.

## Fix

Pass the **local signaling session** to the wrapper (the same `webSocketClient.sessionId` that is
already used for `currentSessionId` a few lines above), so both peers compare IDs from the same
namespace. In internal signaling mode `webSocketClient == null`, so `callSession` is used as
before (unchanged behaviour).

```kotlin
// before
callSession,
// after
if (webSocketClient != null) webSocketClient!!.sessionId else callSession,
```

The election is also extracted into `PeerConnectionWrapper.isOfferer()` to make it unit-testable.

## How it was verified

Measured directly via a debug build logging the two compared session IDs, on a test NC instance
using the same HPB (no MCU):

- **Before the fix** — `localSession` is a Nextcloud session ID (`omX5MFK/…`, base64), `remoteSession`
  is an HPB session ID (`73gsILAp…-g_…`). A↔A automated calls connected **1 / 3** (deadlock 2/3).
- **After the fix** — `localSession` is now an HPB session ID, same namespace as `remoteSession`; both
  peers compare the same pair. A↔A automated calls connected **6 / 6** (0 deadlock).

New unit tests (`PeerConnectionWrapperOffererRoleTest`) pin the contract (exactly one offerer for a
consistent pair) and the cross-namespace failure mode as a regression guard. They are
mutation-verified (breaking `isOfferer` turns them red).

## Not covered / scope

- Web and desktop clients use a different offerer-election path and are unaffected.
- This changes only the offerer election; media/ICE setup is unchanged.

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/activities/CallActivity.kt
@@ -2485,7 +2485,11 @@ class CallActivity : CallBaseActivity() {
             iceServers,
             tempSdpConstraints,
             sessionId,
-            callSession,
+            // #6169: the remote sessionId is a signaling-layer session ID when external signaling
+            // is used, so the offerer election must compare same-namespace IDs. Pass the local
+            // signaling session (matching the currentSessionId resolution used elsewhere); with
+            // internal signaling webSocketClient == null, so callSession is used as before.
+            if (webSocketClient != null) webSocketClient!!.sessionId else callSession,
             tempLocalStream,
             tempIsMCUPublisher,
             tempHasMCU,
--- a/app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java
+++ b/app/src/main/java/com/nextcloud/talk/webrtc/PeerConnectionWrapper.java
@@ -116,7 +116,7 @@ public class PeerConnectionWrapper {
         this.mediaConstraints = mediaConstraints;
 
         sdpObserver = new SdpObserver();
-        boolean hasInitiated = sessionId.compareTo(localSession) < 0;
+        boolean hasInitiated = isOfferer(localSession, sessionId);
         this.isMCUPublisher = isMCUPublisher;
 
         PeerConnection.RTCConfiguration configuration = new PeerConnection.RTCConfiguration(iceServerList);
@@ -163,6 +163,30 @@ public class PeerConnectionWrapper {
                 }
             }
         }
+    }
+
+    /**
+     * Decides whether this client takes the offerer role for a peer connection.
+     *
+     * <p>The role is resolved purely by lexicographic comparison of the two session IDs:
+     * the client whose remote session ID sorts before its own local session ID becomes the
+     * offerer. This yields exactly one offerer per pair <em>only if both peers compare the
+     * same ordered pair of strings</em>.
+     *
+     * <p>Bug #6169: with the High-Performance Backend the remote {@code sessionId} is a
+     * signaling-layer session ID. Callers must therefore pass a {@code localSession} from the
+     * <em>same</em> namespace (the local signaling session, not the Nextcloud session); otherwise
+     * the two peers compare unrelated strings and can both compute {@code false} (neither offers —
+     * deadlock) or both {@code true} (offer collision). The {@code requestoffer} fallback only
+     * applies in MCU mode, so a direct (no-MCU) 2-party call deadlocks outright. The deadlock is
+     * exercised by {@code PeerConnectionWrapperOffererRoleTest}.
+     *
+     * @param localSession  this client's own session ID (same namespace as {@code remoteSession})
+     * @param remoteSession the remote peer's session ID
+     * @return true if this client should create the initial offer
+     */
+    static boolean isOfferer(String localSession, String remoteSession) {
+        return remoteSession.compareTo(localSession) < 0;
     }
 
     public void raiseHand(Boolean raise) {
--- a/app/src/test/java/com/nextcloud/talk/webrtc/PeerConnectionWrapperOffererRoleTest.kt
+++ b/app/src/test/java/com/nextcloud/talk/webrtc/PeerConnectionWrapperOffererRoleTest.kt
@@ -1,0 +1,64 @@
+/*
+ * Nextcloud Talk - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package com.nextcloud.talk.webrtc
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+
+/**
+ * Tests for the offerer-role election ([PeerConnectionWrapper.isOfferer]) — bug #6169.
+ *
+ * Contract: for a pair of peers, exactly one must create the initial offer. The role is resolved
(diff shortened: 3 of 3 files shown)


alauzon:
Since this fixes a call-breaking regression (2-party Android calls deadlock intermittently in HPB mode, no media at all), would it be possible to backport it to the 24.0.2 patch (#111) rather than waiting for 24.1.0?

The change is small and self-contained (just the offerer election at the `PeerConnectionWrapper` construction site + a unit test), it's mergeable with no conflicts, and it's validated with automated 2-party Android↔Android calls on the same HPB (no MCU): ~1/3 connected before, 6/6 after. Happy to rebase onto the stable branch if that helps.

mahibi:
Thank you for contributing @alauzon i will review this soon!

alauzon:
Please hold this — I'm converting it back to draft.

On real-device testing (two phones on different networks), with this change the call now correctly **rings** and an **offer is created** where previously the offerer election deadlocked — so the cross-namespace offerer bug itself is addressed. However, after the callee answers, the call **does not establish media** (it keeps "connecting"). So there is a separate problem in the post-answer ICE/media negotiation that this change does not resolve, and I don't want to give the impression #6169 is fully fixed by this commit alone.

My earlier "connects 6/6" validation was on emulators sharing the same host network, which isn't representative of two real-world NAT'd devices — that's on me. I'll characterize the remaining media-connection issue (TURN relay / answer handling) on real devices before marking this ready again.

mahibi:
Thanks to let me know and for the further testing @alauzon :+1:

mahibi:
it's on my list

mahibi:
hope to find time this week

mahibi:
I'm back from vacation and have this on my list to have a close look

mahibi:
planned for tuesday/wednesday

mahibi (approved):

mahibi:
great catch @alauzon :+1: Thank you for contributing and please excuse the delay for the review!

PR #6396, opened 2026-06-29
Implementing Media Controls for Voice Messages

### LINKED ISSUE #6066: control playback of voice messages in androids notification panel
followup of https://github.com/nextcloud/talk-android/pull/5635 (Playback in ConversationList was removed)

There must be an overall solution to play voice messages in OS notification panel (implement a service?).

rapterjet2004:
- fixes #6066 

Harder than I thought, this required refactoring the entire media player io logic to use MediaSessionService as it's source of truth. While this simplifies the architecture a bit, it also means I have to go about and fix some old bugs once again.

Further follow up PR's would be to move more logic into the view model and simplify chat activity and changing how MediaPlayerManager is used.

Screenshot 2026-07-06 at 11 43 04 AM

### 🚧 TODO

- [x] Get it working
- [x] Fix Pause/Play btn state
- [x] Allow seeking in UI
    - [x] Seeking another messages seekbar also seeks the message next to it, when you play a another message before pausing the currently playing message.
- [x] Add Message MetaData to Control Notification
- [x] Allow waveforms
    - [x] SeekBar speed should match Thumb, might need to do some math here
- [x] Make played seconds update -_-
    - [x] Seconds played should count down from duration

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -195,6 +195,7 @@ configurations.configureEach {
 }
 
 dependencies {
+    implementation("androidx.media3:media3-session:1.10.1")
     kapt("org.jetbrains.kotlin:kotlin-metadata-jvm:$kotlinVersion")
     implementation("androidx.room:room-testing-android:$roomVersion")
     implementation("androidx.compose.foundation:foundation-layout:1.11.2")
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -194,6 +194,7 @@ configurations.configureEach {
 }
 
 dependencies {
+    implementation("androidx.media3:media3-session:1.10.1")
     kapt("org.jetbrains.kotlin:kotlin-metadata-jvm:$kotlinVersion")
     implementation("androidx.room:room-testing-android:$roomVersion")
     implementation("androidx.compose.foundation:foundation-layout:1.12.0")
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -47,6 +47,7 @@
     <uses-permission android:name="android.permission.READ_CONTACTS" />
     <uses-permission android:name="android.permission.WRITE_CONTACTS" />
     <uses-permission android:name="android.permission.READ_PROFILE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK"/>
 
     <uses-permission android:name="android.permission.READ_SYNC_SETTINGS" />
     <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS" />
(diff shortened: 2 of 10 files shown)


mahibi (changes requested):
Works good so far :+1:  Some findings to improve in this or followup PRs:

- there should be the talk icon the the status bar when the talk app is playing a voice message

- clicking on the playback panel should open the chat where the message is played (best would be to scroll to the played back message of course)

- during playing one voice message, i can't start playback of some other voice message. I need to pause the prior message first

- playback speed is not yet re-implemented (was also missing before, since the chatkit removal maybe)

- there is quite some dead code now in ChatViewModel? Like this:

  ```
      fun setPlayBack(speed: PlaybackSpeed) {
          mediaPlayerManager.setPlayBackSpeed(speed)
          viewModelScope.launch {
              _voiceMessagePlayBackUIFlow.emit(speed)
          }
      }
  
      fun startMediaPlayer(path: String) {
          audioRequest(true) {
              mediaPlayerManager.start(path)
          }
      }
  
      fun startCyclingMediaPlayer() = audioRequest(true, mediaPlayerManager::startCycling)
  
      fun pauseMediaPlayer(notifyUI: Boolean) {
          audioRequest(false) {
              mediaPlayerManager.pause(notifyUI)
          }
      }
  
      fun seekToMediaPlayer(progress: Int) = mediaPlayerManager.seekTo(progress)
  
      fun stopMediaPlayer() = audioRequest(false, mediaPlayerManager::stop)
  
      fun queueInMediaPlayer(path: String, msg: ChatMessage) = mediaPlayerManager.addToPlayList(path, msg)
  
      fun clearMediaPlayerQueue() = mediaPlayerManager.clearPlayList()
  ```
   and probably more.

mahibi:
ChatActivity.kt:877:61 Unresolved reference 'currentlyPlayedMessageId'.

i set it to developing. @rapterjet2004 please update the label when i should review

rapterjet2004:
> there should be the talk icon the the status bar when the talk app is playing a voice message
> clicking on the playback panel should open the chat where the message is played (best would be to scroll to the played back message of course)
> playback speed is not yet re-implemented (was also missing before, since the chatkit removal maybe)

I think both of these would be best handled in another PR, as to reduce the scope of this one which involves a fairly significant change to the architecture regarding media playback

> during playing one voice message, i can't start playback of some other voice message. I need to pause the prior message first

This is intentional behavior so we don't have to keep track of more state in the chat list, to synchronize playing with the media controls panels

> there is quite some dead code now in ChatViewModel? Like this:

Should be removed now 👍🏾

mahibi (approved):

PR #6416, opened 2026-07-17
📣 Adding Announcements and Channels

### LINKED ISSUE #6415: 📣 Android: Announcement and Channel presets
- Overview https://github.com/nextcloud/spreed/issues/18683

---

## Overview

**Channel** (`channel` preset, attribute `8`) — broadcasting to a potentially huge audience:
- Only moderators can post, everybody else can only react.
- Permissions are a default, so individuals can still be granted the chat permission.
- Calls are impossible for everyone, including moderators.
- The participants list is only available to moderators.
- System messages are not shown (client-side only).
- Listable for users by default.

**Announcement** (`announcement` preset, attributes `8|16` = `24`) — a channel with extra rules:
- Everything a channel does, unless contradicted below.
- Not listable by default.
- Non-moderators can not leave; moderators can, and can still remove others.
- Notifies about all messages by default (changeable afterwards).

Capability `announcement-preset` indicates the existence and support of both presets. A conversation is a **channel** when `TalkRoom.attributes & 8` (the `CHANNEL` bit) is set, and an **announcement** when `TalkRoom.attributes & 16` (the `ANNOUNCEMENT` bit) is set. An announcement is a channel with additional restrictions, so it **always has both bits set** — every channel behavior below therefore applies to announcements as well, and `attributes & 8` is the only check needed for the shared parts.

### Detection
- [ ] Detect channels via `TalkRoom.attributes & 8` and announcements via `TalkRoom.attributes & 16`; only apply this behavior when the `announcement-preset` capability is present.
- [ ] Add `channel` and `announcement` to `CONVERSATION.PRESET` in `src/constants.ts` and give them an icon in `presetIcons` (`NewConversationSetupPage.vue`).

### Channels (also applies to announcements)
- [ ] **Do not render system messages** in channel conversations. *(frontend-only, no backend block)*
- [ ] Hide the participants list / tab for non-moderators — the list can be enormous. *(backend returns 403 on `GET /room/{token}/participants`, so do not call it)*
- [ ] Hide the call buttons and the whole call UI. Calls are impossible for **everyone**, including moderators: the backend strips `PERMISSIONS_CALL_START`/`PERMISSIONS_CALL_JOIN` from `permissions` and reports `canStartCall: false`, and joining a call returns 403.
- [ ] Show that only moderators can post: non-moderators get `permissions` without `PERMISSIONS_CHAT` (128) but with `PERMISSIONS_REACT` (256), so the message composer should be replaced by a reaction-only affordance. Individual participants can still be granted the chat permission, so always follow the actual `permissions` value rather than assuming.

### Announcements only
- [ ] Hide the "Leave conversation" action for non-moderators — `canLeaveConversation` is `false` for them and `DELETE /room/{token}/participants/self` returns 400 `announcement`. Moderators can still leave and can still remove other participants.
- [ ] The notification level defaults to "all messages" (`1`) for users added to an announcement. It is only a default, so the notification settings must stay editable.

rapterjet2004:
- fixes #6415 

Screenshot 2026-08-03 at 11 29 58 AM

### 🚧 TODO

- [x] test channel usage
- [x] test channel creation - doesn't seem to be working right. (Didn't set permissions right)
- [ ] test announcements

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationcreation/viewmodel/ConversationCreationViewModel.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationcreation/viewmodel/ConversationCreationViewModel.kt
@@ -14,15 +14,12 @@ import androidx.core.net.toFile
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.nextcloud.talk.conversationcreation.data.ConversationCreationRepository
+import com.nextcloud.talk.conversationinfo.CreateRoomRequest
 import com.nextcloud.talk.data.user.model.User
-import com.nextcloud.talk.models.RetrofitBucket
 import com.nextcloud.talk.models.json.autocomplete.AutocompleteUser
 import com.nextcloud.talk.models.json.conversations.Conversation
-import com.nextcloud.talk.models.json.generic.GenericMeta
-import com.nextcloud.talk.repositories.conversations.ConversationsRepositoryImpl.Companion.STATUS_CODE_OK
 import com.nextcloud.talk.utils.ApiUtils
-import com.nextcloud.talk.utils.ApiUtils.getRetrofitBucketForAddParticipant
-import com.nextcloud.talk.utils.ApiUtils.getRetrofitBucketForAddParticipantWithSource
+import com.nextcloud.talk.utils.ParticipantPermissions
 import com.nextcloud.talk.utils.database.user.CurrentUserProviderOld
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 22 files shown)


mahibi:
rebased, resolved some merge conflicts and added some commits to unify how/where the bitwise and capabilities checks are done.

Not yet finished with reviewing but i for now i got these points:

- [x] "admin only" for Announcements means that only admins of the instance should be able to create it. I was logged in without admin rights and could see the option in the "new conversation" screen.

- [x] According to https://github.com/nextcloud/spreed/issues/18683#issuecomment-5140352856 please remove the admin label and change the description instead

- [x] if i was granted the chat permission, wrote something, and then the permission was revoked, do not offer the edit message option.

app/src/main/java/com/nextcloud/talk/conversationcreation/viewmodel/ConversationCreationViewModel.kt
```diff
@@ -82,6 +78,20 @@ class ConversationCreationViewModel @Inject constructor(
         _conversationDescription.value = conversationDescription
     }
 
+    fun updateConversationPreset(preset: String) {
```
mahibi, line 109: - [x] default and voice room also need to be handled here to set the correct values.

Otherwise when switching between the presets the values get sometimes set and sometimes not.

mahibi (changes requested):

mahibi:
@rapterjet2004 please have a look at my commits and let me know if something is wrong/unclear in your opinion.

Besides the points that i mentioned above it look good.

mahibi (approved):

Issue #2375, opened 2022-09-16
show uploading files in conversation


mahibi:
uploading files should be shown in chat, including a progress indicator and a cancel button.

right now for files >1MB there is a notification with progress, for files <1MB there is no notification. for both big and small files, there should be a preview message in the chat while uploading.
maybe it makes sense to postpone this until https://github.com/nextcloud/talk-android/issues/1552 is done.

woolfg:
Fully agree, there is a progress bar in the native Android Notification Panel (if you swipe down) but this is not visible in the app by default. UX wise this is quite confusing if you send a message with an attachment and you don't see any response in the UI after clicking the send button.

woolfg:
May be fixed by #3443

ariedov:
I don't think that is fixed, my wife sent me 3 cute pictures of my son instead of 1 because she did not understand that the picture was already sent.

While I appreciate more cute pictures, it is also clear that a progress bar for uploads is missing.

AndyScherzinger:
Definitely still open, yes, @ariedov

ariedov:
I have looked into it, and there are a few things to consider. It's a big one.

### We treat the upload as a message (Telegram style)

1. The upload status should be another chat message, but local to the device, no one else knows an upload is happening
2. They have to survive state/activity changes
3. This means that we should "fake" the storage of the upload messages, but also
4. In case multiple uploads are happening and other messages appear, the order has to be preserved.
5. Disabled replies to the upload message, which would be crazy to support.

### All uploads are a single popup.

1. This is somewhat similar to the proposed solution before
2. We do not block the UI and allow navigation
3. Except we show a little snackbar at the top of the chat window, indicating the upload progress and the number of uploads
4. Possibly a dialog shows up on the snackbar click, showing all pending uploads.

One is not necessarily easier than the other, maybe a little, but they do have a very different feel. Not sure if I would be able to pull this off, it requires messing with a number of different systems of the app, potentially causing trouble. The app does feel incomplete without the feature though...


### Implementing upload-as-message

I would do it as follows, mainly to avoid altering the existing data storage

1. Create a new table for uploads
2. Create a table entry for every new upload
3. Alter the upload status from the `UploadWorker`
4. Reflect the updated status from the ChatViewModel/ChatActivity/ChatRepository
5. Sort the upload messages into the chat messages based on time (probably) in the ChatViewModel
6. Remove the upload entry from the new table after the upload completes
[...]

PR #6429, opened 2026-07-28
Compress images and videos before upload + new upload/preview dialog + use system camera

### LINKED ISSUE #2461: replace photo camera code with intent to open system camera
With https://github.com/nextcloud/talk-android/pull/2351 we introduced the possibility to record a video directly from the attachments menu.

For photos this is already possible as there is an own implemented camera. However we would like to replace that code with an intent based solution like it's done for the videorecording.

If there are any reasons against this, it can be discussed in this issue. Otherwise we might replace the feature in the future..

Also pinging @ASerbinski to let you know, as you contributed code to the TakePhotoActivity..

### LINKED ISSUE #5190: Use compression for images and files selected via gallery
Human beings use images and memes on a regular basis to communicate "1000 words" (a picture is worth 1000 words). Other messaging platforms prioritize rapid image sending to keep communication flow moving. They do this using several methods, but by far the most impactful method is image (and video) compression on the client. Some also send extremely small and highly compressed thumbnails first, then automatically update the image with a compressed but higher quality image once it's been transferred. Talk does not modify the image what-so-ever before uploading, and with modern phones taking high res photos that result in large file sizes, image transfer is very slow via the Talk client.

When the user selects media (image or video) to send from the "Gallery" option (Android), automatically apply compression techniques to significantly speed the transfer of selected media. Should the user with to send the unaltered, full resolution media file, without compression, they can opt to send it via the "Upload from device" option.

Image

I considered offering compression as a user selectable option under settings, however the question then becomes; is compression enabled by default or not? One would hope it would be on by default, but then this would be a perpetual annoyance for people like myself who routinely sends photos taken with my phone to communicate important information which are fine if compressed, but also often works from my phone to send media meant for print which should not be compressed. I realized since there were two ways to upload media files - gallery and upload - I already had an easy way to choose compressed media from the Gallery or unaltered raw file upload via Upload dialogs.

mahibi:
- resolve #5190
- resolve https://github.com/nextcloud/talk-android/issues/2461


  Adds automatic compression of images and videos before upload in chat, replaces the old attachment picker/preview dialog with a new Compose-based    
  one, and removes the app's built-in camera capture screen in favor of the system camera.                                                             
                                                                                                                                                       
- Compress images and videos before upload                                                                                                           
  - New ImageCompressor/VideoCompressor utilities: images are downscaled and re-encoded as JPEG; videos are downscaled to 540p (short side) and      
  transcoded via Media3 Transformer on a dedicated HandlerThread, keeping the main thread unblocked.                                                   
  - BitmapShrinker modernized to use ImageDecoder on API 28+ (single-pass decode+scale with automatic EXIF rotation for JPEG/HEIC), falling back to  
  the previous BitmapFactory path on API 26/27.                                                                                                        
  - UploadAndShareFilesWorker shows live "Compressing… X%" progress (via Transformer.getProgress()) before switching to the upload progress notification.                                                                                                                                        
    --> this is just a temporary solution until in chat upload previews are implemented.
  - Fixes MIME-type resolution for file:// URIs (e.g. camera-captured photos/videos), which previously broke dialog/thumbnail detection since        
  ContentResolver.getType() only handles content:// URIs.                                                                                              
- New attachment preview/upload dialog                                                                                                               
  - FileAttachmentPreviewFragment/FileAttachmentPreviewDialogCompose rewritten in Jetpack Compose, with thumbnails and per-file dimensions/size shown
  before sending.                                                                                                                                      
- Media quality setting                                                                                                                              
  - The compression toggle is now a "Sent media quality" setting under Settings, presented as a dialog with Original quality / Reduced quality       
  options (previously a simple switch), persisted via AppPreferences.                                                                                  
- Removed built-in camera                                                                                                                            
    - Deleted TakePhotoActivity, TakePictureViewModel, and the associated layout/drawables. Photo/video capture now goes through the system camera;    
  PickImage, ConversationCreationActivity, ConversationInfoEditActivity, and ProfileActivity updated accordingly. 



Assisted-by: Claude Code 2.1.199:claude-sonnet-5

### 🖼️ Screenshots/Videos

🏚️ Before | 🏡 After
---|---
<img width="300" alt="grafik" src="https://github.com/user-attachments/assets/3a7556c5-3c33-46d0-a468-a5512fa75b2d" /> | <img width="400" alt="grafik" src="https://github.com/user-attachments/assets/e5da6b35-6f85-44b3-910b-3bc25de6ecc1" />

Demonstration video

https://github.com/user-attachments/assets/4f83d72a-527c-4420-9950-0d28c6da08a3

Settings

<img width="300" alt="grafik" src="https://github.com/user-attachments/assets/41199ef1-f9b8-4058-98f0-0386f16c1ad5" />

<img width="300" alt="grafik" src="https://github.com/user-attachments/assets/707d5f29-d5bb-4e3c-a41d-c3d20a87083f" />



### TODO for followup PR's
As not to go even further off-topic with this PR it only focuses on the upload/preview dialog and not on the upload itself. So
- https://github.com/nextcloud/talk-android/issues/2375 

is not part of this PR but should be a next step soon.
Although the order of the previews can be modified by drag and drop in this PR, the related code to upload and share does not yet respect this order. This must be part of a followup improvement.

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


jancborchardt (changes requested):
It would be nice to formulate and design it a bit less technical. Signal e.g. specifically calls it "Sent media quality" with an easy to understand "Low" and "High" setting.

I'd say we could have a single-select segmented button in the send flow with:
- Original quality
- Reduced quality

And last selected setting remembered. What do you think @nimishavijay

mahibi:
For reference: thats how signal does it in the send flow (in the left pic it is the (crossed) HD button):

grafik

grafik

nimishavijay:
If we are able to have a fullscreen preview of the image before sending like in whatsapp,telegram, signal, etc, the `HD/SD` icon-only button is pretty standard. if that is not possible then the radio-style buttons in the current UI make sense :)

mahibi:
@jancborchardt @nimishavijay which solution of the HQ selection do you prefer? Do you have more feedback for this current state?:

HQ_button
HQ_text
HQ_segmentedButton

jancborchardt (commented):
@mahibi I’d say the last one with the toggle on the bottom that has actual text. Makes it very clear every time.
It’s best to do it clear & safe in the beginning, we can still make it more obscure later if we notice it’s too much. Important is also that the last preference is remembered.

(Another option to merge the approaches and save space could of course be to have the icon in the top right, but then make it open a menu with 2 entries which have clear wording "Original quality" and "Reduced quality".)

mahibi:
> @mahibi I’d say the last one with the toggle on the bottom that has actual text. Makes it very clear every time. 

alright thanks for the quick feedback.
For the "remember" behavior: The default can be set in the settings. This will be the state that will be set whenever the dialog is opened. Same behavior as for other messengers.

jancborchardt (approved):

rapterjet2004 (changes requested):
Can you remove HighQualityIcons.kt and replace it with the xml equivalent? Found here. 

https://fonts.google.com/icons?selected=Material+Symbols+Outlined:high_quality:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=hq&icon.size=24&icon.color=%23222222&icon.platform=android

Which can be used in Compose with `painterResource`. That way all the resources are in one place. If we keep them in code in it might be a pain to track them all later.

mahibi:
> Can you remove HighQualityIcons.kt and replace it with the xml equivalent? Found here.
> 
> https://fonts.google.com/icons?selected=Material+Symbols+Outlined:high_quality:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=hq&icon.size=24&icon.color=%23222222&icon.platform=android
> 
> Which can be used in Compose with `painterResource`. That way all the resources are in one place. If we keep them in code in it might be a pain to track them all later.

Done :+1: . I was thinking "let's use the compose equivalent from now on" (but yeah, should have chosen a better place to save it of course) but it seems the official documentation from google still recommends to just download the XML:

> ...we recommend using [Google Font Icons](https://fonts.google.com/icons) and download the XML file from the Android Tab to create an up-to-date Material Symbols style Icon.

https://developer.android.com/develop/ui/compose/graphics/images/material

RoboPolaris:
A great first step — thank you so much! This is a really important improvement for using Talk as an everyday messenger.

A few things that would make the experience even better:

1. **Visible compression progress**
   It would be very helpful to have a clear visual indicator while a video is being compressed. At the moment, there is no %-display to see in my version and therefore it is not obvious that a compression process is running.

2. **Video preview in received messages**
   When receiving a video, it would be great to see a muted preview directly in the chat without having to open the player first.

3. **Better fullscreen video player**
   The player would benefit from a proper fullscreen option, automatic fitting to the screen, and automatic rotation/orientation handling.

4. **Higher video quality or selectable presets**
   540p is quite low for normal everyday videos. 720p or 1080p (like Telegram) would be much more useful. Ideally with selectable presets (which would be better than any other messanger) such as:

   * Data saver: 540p
   * Standard: 720p
   * High quality: 1080p
   * Original

5. **Higher image resolution**
   768 px is also quite low for photos. In my opinion, the short side should be at least 1080 px, while still keeping the option to send the original file.

Overall, this is a fantastic and very important step. With a little more flexibility in media quality and a few improvements to the video experience, Talk could become much more suitable as a real replacement for mainstream family and friends messengers.

PR #6454, opened 2026-08-04
🔄️ Chat message pre-fetching

AndyScherzinger:
- New `ChatMessageSyncer`: fetch-and-persist core extracted from `OfflineFirstChatRepository` into a stateless singleton usable without an open chat; both paths share one write path.
- Room-list prefetch: rooms with advanced `lastActivity` (matching iOS) and unread rooms without cached messages get a background catch-up - delta fetch, or initial fetch creating the first chat block for never-opened rooms (beyond iOS). Capped at 20 rooms / 3 concurrent; skipped in battery saver / restricted background data.
- Push-triggered single-room catch-up after the notification is shown (best effort, never delays the notification).
- Notification-safe fetching: `markNotificationsAsRead=0`, read marker untouched, gated on the `chat-keep-notifications` capability.
- Chat open trusts the cache: zero network when cached messages reach `lastMessage.id`; proportional delta fetch otherwise; initial load never waits for the websocket.
- Hardening: chat-block reconciliation on message expiry; empty `/room` responses no longer cascade-delete the cache; insurance fetches anchor on the DB.
- Tests: syncer unit tests + Robolectric integration test; debug logging for logcat verification.

---

**How to verify:** check logcat -> expect "Catching up messages for N rooms", per-room "Background catch-up … fetched N message(s)", and the "Initial online request is skipped … until the conversation's last message" line on open. Most visible on cold opens from push notifications and throttled networks. (for logging, see commit https://github.com/nextcloud/talk-android/pull/6454/commits/64a3c8564c3a78f28490b15e703cb335d032d57c). Alternatively, deploy it on the phone, wait for a while, put the phone in flight-mode, then open the chats. You should then see the unread messages.

---

**Assumptions to be aware of**
1. **`markNotificationsAsRead=0` server semantics** - assumed the server keeps push notifications for background-fetched messages whenever `chat-keep-notifications` is announced; taken from iOS usage, not server code. If wrong, prefetching silently dismisses notifications. *Most critical - please confirm.*
2. **Federated conversations use one consistent message-ID space** - `lastMessage.id` from the room list is compared against local chat-block IDs; divergence via the federation proxy could wrongly skip the initial fetch (unread tail delayed to polling) or fetch redundantly (harmless). Needs a federated-room test.
3. **The chat relay never replays a backlog on (re)connect** - sole justification for the delta fetch at open on HPB; based on code comments/observed design, not signaling-server sources. Partial replay could leave gaps until the 2-min insurance request.
4. **`GET /chat` long polling still works against HPB servers** - the fallback when the websocket is slow/mis-detected; pre-existing behavior, unverified server-side.
5. **`lastActivity` advances for new chat messages** and its persisted copy is a valid cross-restart baseline; missed candidates degrade to pre-PR open behavior, non-message bumps cost one empty delta.
6. **Message IDs strictly increase per conversation** (non-contiguous is fine) - inherited chat-block premise, now leaned on harder for delta anchors.
7. **Message expiration is roughly server-synchronized** - block trimming tolerates clock skew (early: harmless re-fetch; late: briefly stale block, as before).
8. **`pushMessage.id` is always a fetchable room token for TYPE_CHAT pushes** (incl. federated/invite edge cases); wrong token costs one failed, caught request; the notification is unaffected.
9. **`NotificationWorker` lifetime accommodates the synchronous catch-up** - notification dispatched first, but ordering isn't formally guaranteed; an expedited separate work item is the fallback design.
10. **Concurrent persists are race-safe** - open-path delta, first poll, signaling and background catch-up may interleave; upserts + `getConnectedChatBlocks` merging + insurance requests are the intended self-healing, without formal proof. Transaction boundaries deserve review.
11. **Capability cache freshness suffices for the gate** - 12 h `CapabilitiesWorker` window; a server downgrade inside it exposes assumption 1's impact until refresh.
12. **Chat API stays at v1** - `CHAT_API_VERSION = 1` hardcoded in prefetch/push paths, mirroring `ChatViewModel`; compile-time visible, low risk.
13. **20 rooms / 3 concurrent caps are server-friendly** - no rate-limit guidance existed, iOS is uncapped; too aggressive risks throttling, too conservative leaves rooms un-prefetched until the next sync (logged). Tunable constants.

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


app/src/main/java/com/nextcloud/talk/jobs/NotificationWorker.kt
```diff
@@ -211,6 +216,47 @@ class NotificationWorker(context: Context, workerParams: WorkerParameters) : Wor
...
+        }
+
+        // the user from the push signature verification may carry stale capabilities, so resolve
+        // the current state before the capability check in catchUpRoom
+        val currentUser = userManager.getUserWithId(user.id!!).blockingGet() ?: return
+
+        val target = ChatMessageSyncer.SyncTarget(
+            user = currentUser,
+            roomToken = roomToken,
+            threadId = null,
```
mahibi: threadId must not be null.

Try to apply
parseThreadId(ncNotification.objectId)

app/src/main/java/com/nextcloud/talk/chat/ui/VoiceRecordingLockFab.kt
```diff
@@ -0,0 +1,87 @@
+/*
```
mahibi: this class accidentally made it into this PR?

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -278,31 +322,35 @@ class OfflineFirstChatRepository @Inject constructor(
...
         cleanupExpiredMessages()
+
+        val newestMessageIdFromDb = chatBlocksDao.getNewestMessageIdFromChatBlocks(internalConversationId, threadId)
+
         val fieldMap = getFieldMap(
             lookIntoFuture = true,
             timeout = 0,
             includeLastKnown = false,
-            lastKnown = latestKnownMessageIdFromSync.toInt(),
+            lastKnown = newestMessageIdFromDb.toInt(),
```
mahibi: Replacing latestKnownMessageIdFromSync with newestMessageIdFromDb is a major problem.

newestMessageIdFromDb can't be used because it will destroy the insurance request. The insurance request has to know the lastMessage from the last sync (omitting the messages that came in via signaling!).

But chatBlocksDao.getNewestMessageIdFromChatBlocks will include the messages that came in via signaling. So  the insurance request will load messages newer than the last known one from signaling -> This will create persistent gaps!


To understand this also see the comment in onSignalingChatMessageReceived:
        // we assume that the signaling messages are on top of the latest chatblock and include them inside it.
        // If for whatever reason the assumption was not correct and there would be messages in between, the
        // insurance request should fix this by adding the missing messages and updating the chatblocks.

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
```diff
@@ -165,53 +188,74 @@ class OfflineFirstChatRepository @Inject constructor(
...
-                }
-            } else {
-                Log.d(
-                    TAG,
-                    "An online request for newest 100 messages is made because we don't have the lastReadMessage " +
-                        "(gaps could be closed by scrolling up to merge the chatblocks)"
+            }
+
+            weAlreadyHaveSomeOfflineMessages && weHaveAtLeastTheLastReadMessage -> {
+                // Close the backlog since the newest offline message with a delta fetch. This is
```
mahibi: This is a major problem:

it reads like the gap will definitely be closed but keep in mind that only happens when the message limit is not exceeded. So if the gap is >limit it narrows the gap but it will not close it. 

So what happens when the gap is higher than the limit..:

- For long polling this is not a problem. Long polling will just load all messages until it reached the newest, so it "self heals".

- But for chat relay this is a problem:
   - combined with the change in line 353 (replacing latestKnownMessageIdFromSync with newestMessageIdFromDb) this will immediately create permanent gaps when a signaling message comes in before the gaps are closed.
        - but even without replacing latestKnownMessageIdFromSync with newestMessageIdFromDb this could result in 2 minute stop and go fetching by the insurance request until the newest message is reached. 
        
        
So in short, the `weAlreadyHaveSomeOfflineMessages && weHaveAtLeastTheLastReadMessage -> {` block might have to be reverted or modified.

mahibi (changes requested):
Most of it works really good and it's really promising :+1: 

However there are some major problems with the chatrelay and insurance request (see my comments).
These are edge cases what won't happen too often, but they would create permanent gaps.
For now i just identified this by code reviewing (not yet reproduced by testing).
Fixing it should not be too hard, might be just reverting some lines..

I did not finish the code review yet, so there might be more on monday (wont finish it today).

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
AndyScherzinger: Fixed in https://github.com/nextcloud/talk-android/pull/6454/commits/ee89b74aef7a2fd10b6c7b712fa67a94e010f8ee: the insurance anchor is again tracked from HTTP syncs only, but now lives in the singleton ChatMessageSyncer (keyed per conversation/thread) instead of the unscoped repository, so it survives reopening a chat — that reset-to-zero on reopen was the reason I had (wrongly) replaced it with the block id.

Two design decisions go beyond a plain revert, and your explicit 👍/👎 on both might be good / or just to bring them to your awareness:

1. Seeding in the skip branch: when the initial fetch is skipped because the local chat block already reaches conversation.lastMessage.id, I seed the anchor with that lastMessage.id. Rationale: it comes from the room list sync (HTTP, server truth) and the skip condition proves the block covers it — so it should be a valid "known via HTTP" anchor. Without a seed, the first insurance request of a session would query with lastKnownMessageId=0 again.
2. 304/empty confirmation: when a lookIntoFuture request returns 304 or an empty list, I record the queried lastKnownMessageId as the anchor — the server just confirmed nothing newer than it exists, which makes it HTTP-verified even though no message was transferred. This covers the case where the open-path delta fetch finds the cache already current.

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
AndyScherzinger: Fixed in https://github.com/nextcloud/talk-android/pull/6454/commits/a6be0038dc25fa1cba9dea08de60a5a0a33e5553 by introducing ChatMessageSyncer.closeBacklog(): the delta fetch is repeated (anchor advancing to each round's newest persisted message) until the server returns fewer messages than the limit, i.e. until the backlog is provably closed. All three delta paths use it now — the chat-open catch-up, the background prefetch, and the insurance request. For the insurance request this also fixes the "2 minute stop and go" you mentioned: a large backlog is caught up within a single insurance cycle.

Two decisions I want to bring to your awareness:

1. Round cap: the loop is bounded at 5 rounds (≤500 messages at open, ≤1000 for insurance). If the server still returns full pages after that, I stop closing bottom-up and fetch the newest page with includeLastKnown instead. That deliberately creates a separate top chat block, so the remaining gap stays visible in the block structure and gets closed by the normal scroll-up block merging — the invariant being: a chat block never claims a range that wasn't actually fetched. The alternative (loop without bound) seemed worse for rooms with a huge backlog on metered connections.
2. Interplay with the insurance anchor (from the other thread): the fallback fetch advances the http-synced anchor to the newest message. I think that's correct — the anchor's job is verifying the signaling assumption on top of the newest block, and the skipped middle is represented honestly as disconnected blocks rather than being silently claimed — but if you see a case where the insurance request should instead retro-fill that middle range, I'd rather learn it now.

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
mahibi: > Seeding in the skip branch ...

👎, and I removed it in 519c08e27.
The problem isn't the seeding mechanism itself, it's the skip condition it was attached to.                             
`conversationModel.lastMessage` is only as fresh as the last room-list `GET /room` sync — it can already be behind the server by the time the chat is opened 
e.g. a message sent while the conversation list was open. 

If push notifications don't work to update the chat while being in the conversation list, gaps will be created if a server request is skipped when opening a chat.
I reproduced to create gaps with these steps:

1. have genricDebug flavor installed (so notifications do not update the chat DB)
2. open conversation list
3. send a message to android
4. open the chat on android
5. -> the message is missing because lastMessageIdFromServer has an old state
6. send another message to android to this chat while it is opened on android
7. -> see there is a gap


So the skip branch decided "we're already caught up" based on that stale value, skipped the fetch entirely, and the message was genuinely missing from the chat — not just delayed. 
Since the seed was only ever used to feed that same stale value into the anchor, removing the skip branch made `seedHttpSyncedMessageId` dead code too, so I dropped it along with it. 

I simplified the code to this now:

        val weAlreadyHaveSomeOfflineMessages = newestMessageIdFromDb > 0
        val weHaveAtLeastTheLastReadMessage = newestMessageIdFromDb >= conversationModel.lastReadMessage.toLong()
        val weLikelyOnlyHaveASmallBacklog = weAlreadyHaveSomeOfflineMessages && weHaveAtLeastTheLastReadMessage

        Log.d(TAG, "weAlreadyHaveSomeOfflineMessages:$weAlreadyHaveSomeOfflineMessages")
        Log.d(TAG, "weHaveAtLeastTheLastReadMessage:$weHaveAtLeastTheLastReadMessage")
        Log.d(TAG, "weLikelyOnlyHaveASmallBacklog:$weLikelyOnlyHaveASmallBacklog")

        if (weLikelyOnlyHaveASmallBacklog) {
            tryCloseBacklogFromNewestOfflineMessage(newestMessageIdFromDb)
        } else {
            fetchNewestMessagesForInitialLoad(
                withNetworkParams,
                weAlreadyHaveSomeOfflineMessages,
                weHaveAtLeastTheLastReadMessage
            )
        }
        
So i removed the possibility to skip the request entirely. Skipping the request entirely just won't work as we never can be sure the conversationModel is up to date.
We now just rely on
weAlreadyHaveSomeOfflineMessages && weHaveAtLeastTheLastReadMessage
as a vague indicator that the backlog might not be too huge. It only decides between trying to close the backlog vs fetching the newest messages. 
It's a pity we can not just skip a request entirely but in my opinion this is the only way we can be sure to not create gaps. UX wise this is not a problem: The prefetched messages will be shown immediately and nothing is blocked. With the prefetching we just assure that it happens way less that new messages have a short delay to be shown. 



> 304/empty confirmation ...

:+1:  that's perfectly fine

app/src/main/java/com/nextcloud/talk/chat/data/network/OfflineFirstChatRepository.kt
mahibi: > Round cap ...

that works fine and is the best solution (As far as Joas is okay with the traffic it will produce).
However in 9c9db41b i changed the output of the SyncOutcome: Mixing the fallbackOutcome with the newestPersisted reported a misleadingly contiguous oldest..newest
span for two unrelated ranges with an unclosed gap in between.
This was not yet a problem as it was never relied on this outcome. But for future code changes this would have been a pitfall.


> Interplay with the insurance anchor ...

Yes the fallback fetch must advance the http-synced anchor. That's correct.
It's not about avoiding gaps. It's about avoiding gaps inside chatBlocks. So gaps between chatBlocks are okay as they are handled on the fly when scrolling (chatBlocks just grow or get merged).

mahibi:
With the current changes that i made i think the great dangerous pitfalls are avoided. 
However i would like to test this branch even more tomorrow..

AndyScherzinger:
> With the current changes that i made i think the great dangerous pitfalls are avoided.
> However i would like to test this branch even more tomorrow.

@mahibi sounds good to me. I'd also like to further test and added some more commits to improve the unread marker behavior while opening talk rooms with a lot of unread messages - a thing I experienced i.e. opening rooms with more than 200 unread messages.

mahibi:
Regarding the **Assumptions to be aware of**:

TLDR: point 9 and 10 needs improvements in my opinion, but could be done in followup PRs.

---


1. **`markNotificationsAsRead=0` server semantics**

Yes, the server keeps push notifications for background-fetched messages whenever `chat-keep-notifications` 

2. **Federated conversations use one consistent message-ID space**

Yes, they do. All fine

3. **The chat relay never replays a backlog on (re)connect** - sole justification for the delta fetch at open on HPB; based on code comments/observed design, not signaling-server sources. Partial replay could leave gaps until the 2-min insurance request.

All fine. Signaling messages are gone if they are not received for whatever reason. Only the insurance request will catch up (going forward from the latest message form initial loading).

4. **`GET /chat` long polling still works against HPB servers** - the fallback when the websocket is slow/mis-detected; pre-existing behavior, unverified server-side.

yes

5. **`lastActivity` advances for new chat messages** and its persisted copy is a valid cross-restart baseline

yes

6. **Message IDs strictly increase per conversation** (non-contiguous is fine) - inherited chat-block premise, now leaned on harder for delta anchors.

yes


7. **Message expiration is roughly server-synchronized** - block trimming tolerates clock skew (early: harmless re-fetch; late: briefly stale block, as before).

yes, deleteExpiredMessages compares expirationTimestamp (server-set, from the message JSON) against System.currentTimeMillis() / 1000 


8. **`pushMessage.id` is always a fetchable room token for TYPE_CHAT pushes** (incl. federated/invite edge cases); wrong token costs one failed, caught request; the notification is unaffected.

yes

9. **`NotificationWorker` lifetime accommodates the synchronous catch-up** - notification dispatched first, but ordering isn't formally guaranteed; an expedited separate work item is the fallback design.

This should be improved. Fetching messages could be done in a separate worker, including a better workers retry success/failure/retry handling.


10. **Concurrent persists are race-safe** - open-path delta, first poll, signaling and background catch-up may interleave; upserts + `getConnectedChatBlocks` merging + insurance requests are the intended self-healing, without formal proof. Transaction boundaries deserve review.

This should indeed be checked. 

updateBlocks (ChatMessageSyncer.kt:821) does:
upsertChatBlock  
getConnectedChatBlocks 
replaceConnectedChatBlocks (at least this one is marked as Transaction)

The sequence across these methods is not atomic. With the open-path delta, long-poll/insurance, and signaling all able to call into this concurrently for the same conversation, two near-simultaneous callers could each upsert a block and then query connectivity before seeing the other's write. 


11. **Capability cache freshness suffices for the gate** - 12 h `CapabilitiesWorker` window; a server downgrade inside it exposes assumption 1's impact until refresh.

yes the capabilitis worker is set to 12h. Not sure how this is relevant here. Server downgrades should be an edge case and i don't see problems here.


12. **Chat API stays at v1** - `CHAT_API_VERSION = 1`

yes, can stay as is


13. **20 rooms / 3 concurrent caps are server-friendly** - no rate-limit guidance existed, iOS is uncapped; too aggressive risks throttling, too conservative leaves rooms un-prefetched until the next sync (logged). Tunable constants.

These are realistic values in my opinion

mahibi (approved):

mahibi:
reviewed and tested a lot. It's ready to merge in my opinion.
Great improvements :+1:  :rocket:

PR #6327, opened 2026-06-09
Show file upload progress and placeholder media message in chat.

mahibi:
resolve https://github.com/nextcloud/talk-android/issues/2375

copied the PR https://github.com/nextcloud/talk-android/pull/6047 from @anakin78z to the nextcloud repo as he won't find the time to continue.
I will

- [x] first try to resolve the conflicts 
- [x] and whenever time allows continue implementation. 

Thank you @anakin78z for this initial code!

### 🏁 Checklist

- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [x] 🔖 Capability is checked or not needed 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set
- [x] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

PR #6486, opened 2026-08-12
fix(chat): guard localLastReadMessage against implausible message ids

mahibi:
Prevents unplausible high message ids from being sent to the server as the read marker, which was marking conversations read far beyond their real last message and causing new chat notifications to be immediately deleted.

Assisted-by: Claude Code 2.1.199:claude-sonnet-5

This should ensure unrealistic high messageIds for lastReadMessage never make it to the server
- see https://github.com/nextcloud/talk-android/pull/6327/files#r3766002280
- see https://github.com/nextcloud/talk-android/pull/6330

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


rapterjet2004 (approved):
It remains important to make sure that the conversation model is synchronized with the server. We don't want to end up missing notification messages because the `conversation?.lastMessage` is out of date.

mahibi:
> It remains important to make sure that the conversation model is synchronized with the server. We don't want to end up missing notification messages because the `conversation?.lastMessage` is out of date.

Sure. As far as i see an outdated lastMessage would not be a problem regarding the notifications. It's just about gating unrealistic high messageIds (that could be caused by bugs) from becoming the lastReadMessage on server

PR #6498, opened 2026-08-14
🔄️ Improve conversation list performance

AndyScherzinger:
For the user interface, rely on the database, make action where useful optimistic with self-healing in place based on the regular server-fetches. Sagefuard the changes via actions when competing with server fetches being in-flight, likely impacting the conversation list data.

### 🏁 Checklist

- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt
@@ -2790,20 +2790,7 @@ class ChatActivity :
     }
 
     private fun updateRemoteLastReadMessageIfNeeded() {
-        if (this::spreedCapabilities.isInitialized) {
-            spreedCapabilities?.let {
-                val url = ApiUtils.getUrlForChatReadMarker(
-                    ApiUtils.getChatApiVersion(it, intArrayOf(ApiUtils.API_V1)),
-                    conversationUser.baseUrl!!,
-                    roomToken
-                )
-
-                chatViewModel.updateRemoteLastReadMessageIfNeeded(
-                    credentials = credentials!!,
-                    url = url
-                )
-            }
-        }
+        chatViewModel.updateRemoteLastReadMessageIfNeeded()
     }
 
     private fun isActivityNotChangingConfigurations(): Boolean = !isChangingConfigurations
@@ -3487,15 +3474,7 @@ class ChatActivity :
     }
 
     private fun markAsRead(messageId: Int) {
-        chatViewModel.setChatReadMessage(
-            credentials!!,
-            ApiUtils.getUrlForChatReadMarker(
-                ApiUtils.getChatApiVersion(spreedCapabilities, intArrayOf(ApiUtils.API_V1)),
-                conversationUser?.baseUrl!!,
-                roomToken
-            ),
-            messageId
-        )
+        chatViewModel.setChatReadMessage(messageId)
     }
 
     fun markAsUnread(chatMessage: ChatMessage) {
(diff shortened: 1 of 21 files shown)


mahibi:
Not for this PR, but an idea for a followup PR. Summarized by claude:


# Allow favorite/archive/mark-unread/tags while offline                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                   
## Problem                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                   
  The conversation list's quick actions (favorite, archive, mark as unread, tags) are hidden while offline — gated by isOnline in ConversationList.kt:392 (swipe) and networkMonitor.isOnline.value in ConversationsListActivity.handleConversationLongClick (long-press menu). This predates the optimistic-write + pending-guard mechanism added in feat/noid/convoListUpdate, which now makes it safe to write these changes locally and reconcile with the server later.                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                   
  Mark-as-read/unread already gets this treatment for the read marker specifically: ChatViewModel.setChatReadMessage writes the local state immediately and hands delivery to ReadMarkerSyncWorker — a NetworkType.CONNECTED-constrained, exponential-backoff, disk-persisted WorkManager job. Favorite, archive, mark-unread (from the list), and tags do not have this — they  call withRetry(1) inline in the viewmodel coroutine (ConversationsListViewModel.addConversationToFavorites, toggleConversationArchive, markConversationAsUnread, ConversationTagsViewModel) and revert the optimistic write immediately on any failure, including "currently offline." Removing the UI gate today would just make those actions flash and then flip back with an  error while offline — not the offline-first behavior we want.                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                   
## Proposal                                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                   
  1. Give favorite, archive, mark-unread, and tags their own durable delivery, mirroring ReadMarkerSyncWorker: either one worker per action or a single parameterized worker keyed by action type + room. Each enqueued with NetworkType.CONNECTED, exponential backoff, and unique-per-room-per-action work names (ExistingWorkPolicy.REPLACE) so a later action supersedes an earlier queued one.                                                                                                                                                                                                                                                                                                                                                              
  2. Point each viewmodel action (ConversationsListViewModel, ConversationTagsViewModel) at its worker instead of the inline withRetry(1) call.                                                                                                                                                                                                                                    
  3. The existing pendingFavorites / pendingArchivedFlags / pendingUnreadFlags / pendingTagIds guards in ConversationListUpdater need no changes — they already keep the optimistic value alive in the merge regardless of how long delivery takes.                                                                                                                                
  4. Relax the isOnline gates: ConversationList.kt's SwipeableConversationItem(enabled = isOnline) and ConversationsListActivity.handleConversationLongClick's online check, once the actions behind them are durable.                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                   
## Out of scope                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                   
  - The FAB / "create new conversation" and anything that genuinely needs a round trip to exist (room creation) — leave those gated on connectivity.                                                                                                                                                                                                                               
  - No changes needed to the pending-guard merge logic itself.                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                   
## Risks / open questions                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                   
  - Precedent already exists in the app (UploadAndShareFilesWorker, ShareOperationWorker, ContactAddressBookWorker), so this is consistent with existing architecture, not a new pattern.                                                                                                                                                                                          
  - Decide whether one generic worker or four action-specific workers is less code — probably one worker with an enum/string action type, given how similar ReadMarkerSyncWorker would be to the others.                                                                                                                                                                           
  - Consider UX for "action queued, will sync when back online" — e.g. does the star/archive icon need a subtle pending indicator, or is silent optimistic-until-confirmed enough (matches current mark-as-read behavior, which has none).

mahibi (approved):
Great improvements again!  :rocket: :+1: 

during debugging i was sometimes still able to reproduce this:

1. got to chat with unread messages
2. read all
3. go back
4. conversation is marked as read
5. but a second later it's marked as unread
6. only after pull it is marked as read again

With the two commits i added this seems to be fixed. I was not able to reproduce it anymore.

I might create more followup PR's but this one should be merged.

AndyScherzinger:
Nice improvements @mahibi 🎉

PR #6499, opened 2026-08-16
🔄️ Improve conversation avatar caching

AndyScherzinger:
Conversation avatars somtimes flashed the app-shipped fallback avatar (`account_circle_96dp`) while scrolling / when loading before rendering the actual avatar - espesially after cold starts, sporadically during a session, and while scrolling. Three changes rework how avatars are fetched, cached and rendered.

So this PR changes the avatar rendering berhavior and loading/caching logic.

<details>

### 1. Versioned avatars are cached immutably

The avatar endpoints send `no-cache` headers, so Coil revalidated avatars over the network on every memory-cache miss — every cold start refetched all list avatars. Since the `avatarVersion` URL parameter already is the invalidation token, versioned avatar URLs are immutable content: 
a dedicated avatar `ImageLoader` (sharing memory/disk caches with the default one) now ignores the cache headers. Cached avatars never expire or revalidate; a new `avatarVersion` from the room list sync changes the URL and forces the fetch. Cold starts and offline usage render avatars straight from the disk cache.

One-to-one conversations previously used the unversioned user-avatar endpoint, outside the version scheme. They now use the versioned conversation-avatar endpoint (which natively serves the peer's avatar), so their refresh is also driven purely by `avatarVersion` changes.

**Server compatibility:** both changes are gated on the `avatar` capability (Talk 17+). On older servers, one-to-one rooms keep the unversioned user-avatar endpoint with header-driven revalidation, and group/public rooms show themed default icons — which also fixes the previous behavior of unconditionally requesting the conversation-avatar endpoint that does not exist there.

### 2. A known avatar is never replaced by the fallback

Every successful avatar load also stores its bitmap under a stable per-room alias in the shared memory cache. Requests use that alias as placeholder *and* error image: an `avatarVersion` bump or theme switch crossfades old → new, and transient load failures keep the last known avatar.

### 3. The fallback icon is a fallback, not a loading indicator

While an avatar loads and no previous one is known (e.g. rows entering the viewport for the first time in a session while scrolling), nothing is rendered — the slot keeps its size and the avatar crossfades in. The fallback icon only appears when a load fails and no avatar was ever shown.

</details>

### Resulting behavior

| Scenario | Rendered avatar |
|---|---|
| Memory-cache hit (most scroll re-entries) | avatar instantly, no intermediate state |
| Loading, avatar was shown before in this session | last shown avatar, crossfade if it changed |
| Loading, not shown yet this session | nothing (blank slot), avatar crossfades in from disk/network |
| Load failed, previous avatar known | previous avatar |
| Load failed, nothing ever shown | fallback icon (its only remaining appearance) |
| Server without `avatar` capability (pre Talk 17) | 1:1 via user-avatar endpoint, group/public rooms show themed default icons |

### Testing

- Unit tests cover the endpoint/capability selection (`ConversationAvatarContentTest`), incl.   the fallback matrix for servers without the `avatar` capability.
- Manual: cold start with warm cache (no flash, avatars from disk), avatar change on another   client (old → new crossfade after the next room list sync), airplane-mode cold start (avatars   render offline), fast scrolling (blank slots instead of fallback flashes), theme switch.

### 🏁 Checklist

- [x] 📅 Milestone is set

## 🤖 AI (if applicable)

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationlist/ui/AvatarContent.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationlist/ui/AvatarContent.kt
@@ -1,0 +1,100 @@
+/*
+ * Nextcloud Talk - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package com.nextcloud.talk.conversationlist.ui
+
+import androidx.annotation.DrawableRes
+import com.nextcloud.talk.R
+import com.nextcloud.talk.data.user.model.User
+import com.nextcloud.talk.models.domain.ConversationModel
+import com.nextcloud.talk.models.json.conversations.ConversationEnums
+import com.nextcloud.talk.utils.ApiUtils
+import com.nextcloud.talk.utils.SpreedFeatures
+
+internal sealed class AvatarContent {
+    /**
+     * [versioned] marks an immutable avatar URL: the avatarVersion parameter is the invalidation
+     * token, so the image may be cached without expiry - a new version changes the URL. Only
+     * conversation avatars are versioned; a one-to-one room's avatar is the peer's user avatar,
+     * which is outside the version scheme and must revalidate via the response cache headers.
+     */
+    data class Url(val url: String, val versioned: Boolean) : AvatarContent()
+    data class Res(@param:DrawableRes val resId: Int) : AvatarContent()
+    object System : AvatarContent()
+    object NoteToSelf : AvatarContent()
+}
+
+/**
+ * Resolves what to show as a conversation's avatar.
+ *
+ * On servers with the avatar capability (Talk 17+) rooms use the conversation-avatar endpoint.
+ * For everything but one-to-one rooms the URL carries the avatarVersion as invalidation token,
+ * so those avatars are immutable content refreshed solely by version changes from the room list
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/talk/conversationlist/ui/AvatarContent.kt
```diff
@@ -0,0 +1,84 @@
...
+    data class Url(val url: String, val versioned: Boolean) : AvatarContent()
+    data class Res(@param:DrawableRes val resId: Int) : AvatarContent()
+    object System : AvatarContent()
+    object NoteToSelf : AvatarContent()
+}
+
+/**
+ * Resolves what to show as a conversation's avatar.
+ *
+ * On servers with the avatar capability (Talk 17+) every room type — one-to-one included, the
```
SystemKeeper: I am pretty sure this is not entirely true. While we do serve a version for 1:1 rooms, they don’t change when a new avatar is uploaded. I wanted to have a stable versioning for ages, but it’s quite tricky. 

I can check in more depth later, but for ref: https://github.com/nextcloud/talk-ios/blob/b5b137bd3145c93bfc870f1069b7f9a27428b4ea/NextcloudTalk/Network/NCAPIController.swift#L3650-L3651

AndyScherzinger: You are right 😭

mahibi (approved):

PR #6597, opened 2026-08-27
fix(scheduled-messages): use unique message id as LazyColumn item key

mahibi:
fix the crash:

```
Exception java.lang.IllegalArgumentException:
  at androidx.compose.ui.internal.InlineClassHelperKt.throwIllegalArgumentException (InlineClassHelper.kt:36)
  at androidx.compose.ui.layout.LayoutNodeSubcompositionsState.subcompose (SubcomposeLayout.kt:1592)
  at androidx.compose.ui.layout.LayoutNodeSubcompositionsState$Scope.subcompose (SubcomposeLayout.kt:1353)
  at androidx.compose.foundation.lazy.layout.LazyLayoutMeasureScopeImpl.compose (LazyLayoutMeasureScope.kt:94)
  at androidx.compose.foundation.lazy.layout.LazyLayoutMeasuredItemProvider.getPlaceables-3p2s80s (LazyLayoutMeasuredItem.kt:118)
  at androidx.compose.foundation.lazy.LazyListMeasuredItemProvider.getAndMeasure-0kLqBqw (LazyListMeasuredItemProvider.kt:54)
  at androidx.compose.foundation.lazy.LazyListMeasuredItemProvider.getAndMeasure-0kLqBqw$default (LazyListMeasuredItemProvider.kt:48)
```

however i assume there is still some other bug in the app which produces the same issue (given the amount of reports its maybe chatview itself. to be fixed in a separate PR).

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/chat/ScheduledMessagesActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/chat/ScheduledMessagesActivity.kt
@@ -545,7 +545,7 @@ class ScheduledMessagesActivity : BaseActivity() {
 
                                         items(
                                             items = grouped[date].orEmpty(),
-                                            key = { it.token ?: "" }
+                                            key = { it.jsonMessageId }
                                         ) { message ->
 
                                             val parentId = message.parentMessageId


mahibi:
> however i assume there is still some other bug in the app which produces the same issue (given the amount of reports its maybe chatview itself. to be fixed in a separate PR).

most probably commit 1b71ecb0 from https://github.com/nextcloud/talk-android/pull/6327
Using the referenceId when replacing temporary messages with actual messages might have some problems..
-> "fixed" with a9d1533a

rapterjet2004 (approved):

PR #6649, opened 2026-09-04
fix(conversations): apply the selected preset's parameters when creating a conversation

AndyScherzinger:
The conversation creation screen sent only the preset identifier to the creation endpoint. That
endpoint derives just the conversation attributes from it and stores whatever parameters the request
carried, so the parameters of the selected conversation type never arrived:

* Voice rooms were created neither listable nor expiring.
* Channels and announcements granted call and media permissions (372) instead of reactions only
  (257), which contradicts our own `nc_channel_description` string.
* The "allow guests" switch had no effect whatsoever: `roomType` was hardcoded to group, so no
  public conversation was ever created, and the password was then set on a private one.
* The administrator configured defaults never reached a created conversation.

On top of that the redundant `POST /listable` after creation is gone, the voice room type is hidden
where calls are disabled server wide, and servers without `conversation-creation-all` are served by
follow-up requests instead of silently losing the participants, the description and the listable
state.

This is the first of three stacked PRs; the conversation types themselves move to the presets
endpoint in the follow-up.

### 🖼️ Screenshots

No visual change other than the voice room type being hidden where calls are disabled server wide,
so no before/after shots.

### 🚧 TODO

- [x] nothing outstanding

### 🏁 Checklist

- [x] 📅 Milestone is set

Capabilities checked: `conversation-creation-all`, `conversation-creation-password` (added here),
`conversation-presets`, `announcement-preset` and `config => call => enabled`.

## 🤖 AI (if applicable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014YnDswTbCmRrVETnwE4twB

### DIFF
--- a/app/src/main/java/com/nextcloud/talk/conversationcreation/ConversationCreationActivity.kt
+++ b/app/src/main/java/com/nextcloud/talk/conversationcreation/ConversationCreationActivity.kt
@@ -107,7 +107,6 @@ import com.nextcloud.talk.models.json.autocomplete.AutocompleteUser
 import com.nextcloud.talk.utils.CapabilitiesUtil
 import com.nextcloud.talk.utils.DisplayUtils
 import com.nextcloud.talk.utils.PickImage
-import com.nextcloud.talk.utils.SpreedFeatures
 import com.nextcloud.talk.utils.bundle.BundleKeys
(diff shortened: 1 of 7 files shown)


mahibi (approved):
added one commit.
lgtm+tested it

PR #6697, opened 2026-09-14
fix: make pinned message text tappable to scroll to message

mahibi:
The pinned message bubble's Card had an onClick to scroll to the message, but the embedded MarkdownText's TextView became clickable via LinkMovementMethod whenever it contained a link/mention, silently swallowing taps on the text itself since no listener was attached. Only taps on the icon/padding around the text reached the Card.

Assisted-by: Claude Code:claude-sonnet-5

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


rapterjet2004 (changes requested):
Upon tapping a pinned message, the highlighted UI remains, instead of disappearing. 

Image

In addition, I noticed that re-entering ChatActivity removes the pinned UI, despite it remaining under the shared items dialog. This might be unrelated to the PR.

mahibi:
> Upon tapping a pinned message, the highlighted UI remains, instead of disappearing.

That's intentional (and was also before this PR), same behavior as when opening a message search result from the conversation list search.

> In addition, I noticed that re-entering ChatActivity removes the pinned UI, despite it remaining under the shared items dialog. This might be unrelated to the PR.

Can't reproduce

mahibi:
However in you screenshot, there is an empty string for "pinned by". Should be fixed in another PR

rapterjet2004 (approved):

PR #6701, opened 2026-09-15
fix: only show the answer button matching the caller's call type

### LINKED ISSUE #6402: 📞 Distinguish video and voice calls
When someone calls me, the video call and voice calls are seperate. When an iPhone user video called me, I accepted as a voice call and neither of us saw each others video feed. (I didn't see the iPhones because I was in voice mode, and the iPhone user didn't see mine, because well, voice call) The point is, that I don't know the intention of the person calling.

Instead of two buttons for both voice and video call, show only one for the mode, the person calling chose.

Always picking up as video call, but it's more user friendly to have only one button and I have to turn off my camera after joining if we only want voice call.

**Additional context**
How to solve for when calling from desktop? Maybe display only video call button as the user doesn't have to have the camera turned on and if accepted as voice call, you cannot turn on the camera afterwards.

rapterjet2004:
- fixes #6402 

Incoming voice calls showed both a voice-only and a video answer button, so it was easy to accept a video call as voice-only and vice versa, leaving both sides without a video feed. Now the ringing screen hides whichever button doesn't match the known call type (falls back to showing both when the server doesn't expose conversation-call-flags). Also collapses the now-redundant duplicate answer PendingIntent construction in NotificationWorker.

Assisted-by: Claude Code:claude-sonnet-5

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


mahibi:
How does the iOS app handle this nowadays @Ivansss @SystemKeeper ?

SystemKeeper:
When we receive an incoming call it is audio only by default, since that information is not provided by the push payload (I still think we should do that...). We then fetch the information from the server and in case the callFlags indicate that the call is with video, we set that internally (though, depending on the request duration it could be racy). 
Other than that we can downgrade/upgrade to voice/video from a menu.

Not sure that answers your question?

SystemKeeper:
https://github.com/nextcloud/talk-ios/blob/0231f1c6b6ce51aa13dd9bde5008ae0839e1f8d1/NextcloudTalk/Calls/CallKitManager.swift#L275-L303

mahibi:
it's fine for me to merge, however be aware this only fixes an edge case: 
it only works reliable for calls from phone to phone. Because calls from web always come with video.

The call flags can't tell us whether the caller actually has video on. WITH_VIDEO only means "joined able to send video", not "camera is on". A call from web always comes in as  a video call, even with the camera off. So with this PR, you'd get only the video button for every web call, and the person being called would expect video that never shows up. In group calls the flag is a mix of everyone's flags anyway. 
   
iOS doesn't have better info either. It also fetches the room and checks callFlag for WITH_VIDEO (CallKitManager.swift). But there it only changes how the incoming call looks. There's a single answer button, and you can switch between voice and video during the call.

So the real important issue that must be addressed is #708: switching voice/video during the call. 
See comment https://github.com/nextcloud/talk-android/issues/708#issuecomment-5816914649
Then picking the wrong button doesn't matter anymore.

mahibi (approved):

PR #6720, opened 2026-09-18
Send text messages by worker

### LINKED ISSUE #4596: Worker to send temporary messages
followup to https://github.com/nextcloud/talk-android/pull/4422

Problems:
- When writing in chat A in offline mode and go to chat B, that messages in chat A are not sent automatically when device is online again.
- When writing in chat A in offline mode and then closing app, messages are not sent

Plan:
introduce a Worker to handle these cases.

mahibi:
- resolve https://github.com/nextcloud/talk-android/issues/4596

What changed (4 commits on this branch):                                                                                                                                
                                                                                                                                                                          
  1. 1fc75dbb3 — Text messages were sent from a plain viewModelScope coroutine tied to ChatActivity's ViewModel. If you left the chat while offline, that coroutine (and  
     the only reconnect-listener that retried sending) died with it, so the message never went out. Added SendMessageWorker, a CoroutineWorker enqueued via WorkManager   
     and chained per conversation (ExistingWorkPolicy.APPEND_OR_REPLACE), which survives leaving the chat, backgrounding, and process death, and retries with backoff.    
     Both the normal send path and the manual "resend failed message" action now go through it.                                                                           
  2. d4377ea07 — Deleting a message (via the message actions sheet) while it was still queued offline didn't stop the already-enqueued worker; it still posted the message
     once connectivity returned. The worker now re-checks that the temporary message row still exists at the start of every attempt (including retries) and skips the     
     send if it's gone.                                                                                                                                                   
  3. 0858d09dd — Editing a still-queued offline message only updated the local temp-message row; the worker kept sending the original text it had captured into           
     WorkManager's input data at enqueue time. The worker now reads the message text and other send parameters fresh from that same row at send time instead of from stale
     captured data, so edits (and everything else already persisted on that row) are always picked up.                                                                    
  4. d07bdd603 — Sending several messages in quick succession could deliver them to the server out of order (visible as a mismatch between mobile and web client ordering,
     until the periodic resync corrected it), because each send's local DB write and worker-enqueue ran as an independent, unordered coroutine. A Mutex now serializes    
     that step per conversation so sends reach the server in the order they were actually sent.                                                                           
                                                                                                                                                                          
  Suggested manual test steps:                                                                                                                                            
                                                                                                                                                                          
  - Offline send survives leaving the chat: Turn off network. Send a message. Immediately leave the chat (back out of ChatActivity, don't just background it). Turn       
    network back on. Reopen the chat — the message should now show as sent (not stuck pending/failed).                                                                    
  - Delete while offline: Turn off network, send a message, long-press it → Delete. Turn network back on. Confirm it never appears on the server side (e.g. check via web 
    client) — it should not have been sent.                                                                                                                               
  - Edit while offline: Turn off network, send a message, long-press it → Edit, change the text, submit the edit. Turn network back on. Confirm the message that arrives  
    on the server has the edited text, not the original.                                                                                                                                                                                                                       
  - Ordering under rapid sends: With a stable connection, type and send several short messages as fast as possible in a row. Compare the order they appear in on this     
    device vs. a second client (web or another device) — they should match with no reordering.                                                                            
  - Regression check: normal online sending, editing, deleting, and resending should behave exactly as before.

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


rapterjet2004 (approved):

rapterjet2004 (approved):
Just fix the merge conflicts, shouldn't be a problem.

PR #6727, opened 2026-09-19
fix(chat): make the push message prefetch actually run on a flaky connection

AndyScherzinger:
### 🖼️ Screenshots

No UI change, this is background fetch behaviour only.

Push notifications prefetch a pushed room's messages so the chat is already populated when it is opened from the notification. For a number of reporters that stopped working, typically on a flaky wifi, and the chat opens empty. Three causes, all on the fetch side — the chat screen renders straight from the database, so an empty chat means nothing was persisted.

**A failed chat pull was retried with a smaller page each time** (100 → 50 → 10 → 5). A connection that fails a few times before succeeding therefore persists a five message chat block, and since the chat shows the latest block the conversation opens nearly empty even though the prefetch reported success. The limit is now kept across retries and the retry backs off in time instead, which also stops four immediate requests being fired at a server that is already struggling or rate limiting us.

**The background catch-up refused to fetch whenever `NetworkMonitor` reported the device as offline.** That flag is a `StateFlow` shared while subscribed and every collector in the app is UI scoped, so a worker started by a push reads whatever value was cached when the monitor was constructed and no callback ever updates it — a process backgrounded while offline skips every catch-up for as long as it lives. On top of that the check requires `NET_CAPABILITY_VALIDATED`, so it also skips on a connected but unvalidated network, which is exactly the flaky wifi case the prefetch exists for. The pre-check is gone: callers are already gated by a WorkManager `NetworkType.CONNECTED` constraint, and a request that fails because we really are offline is reported as a failed sync and retried with backoff.

**`NetworkMonitor.isOnline` now shares its upstream eagerly**, so reading its value from a background context returns the current state rather than a snapshot from process start. The scope already lived for the whole process, so this replaces repeated register/unregister cycles with one long lived network callback.

Battery saver still disables the prefetch outright (`ChatMessageCatchUpWorker`), and the catch-up is still only triggered for `type == "chat"` pushes, so reminders and room pushes don't prefetch. Both are deliberate and left for a follow-up.

### 🚧 TODO

- [ ] Confirm against a reporter log that the catch-up now runs (`Background catch-up for room …: fetched N message(s)` instead of `Device is offline, skipping catch-up …`)

### 🏁 Checklist

- [x] 📅 Milestone is set

## 🤖 AI (if applicable)


AndyScherzinger:
D:/Tools/Git/backport to stable-25.0.x

mahibi (approved):

PR #6730, opened 2026-09-19
Fix no supported api exception

mahibi:
avoid the crash                                                                                                                                                                                                                   
                                                                                                                                                                                                                                   
```
  Exception com.nextcloud.talk.utils.NoSupportedApiException:                                                                                                                                                                      
    at com.nextcloud.talk.utils.ApiUtils.getConversationApiVersion (ApiUtils.kt:139)                                                                                                                                               
    at com.nextcloud.talk.chat.data.network.RetrofitChatNetwork.getRoom (RetrofitChatNetwork.kt:32)                                                                                                                                
    at com.nextcloud.talk.conversationlist.data.network.OfflineFirstConversationsRepository$getRoom$1.invokeSuspend (OfflineFirstConversationsRepository.kt:122)       
```                                                            

RetrofitChatNetwork.getRoom resolved the conversation API version while building the RxJava Observable, before it was ever subscribed to. When a user's spreed capabilities aren't loaded/available yet,                         
  ApiUtils.getConversationApiVersion throws NoSupportedApiException synchronously — bypassing RxJava's onError entirely, since the exception happens before .subscribeOn()/.subscribe() runs. In                                   
  OfflineFirstConversationsRepository.getRoom this call sat inside a bare coroutine launch with no surrounding try/catch, so the exception escaped uncaught and crashed the app. ConversationInfoViewModel.getRoom and             
  NotificationWorker.handleCallPushMessage had the same eager-throw shape.                                                                                                                                                         
                                                                                                                                                                                                                                   
  Changes                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  - ChatNetworkDataSource.getRoom / RetrofitChatNetwork.getRoom: converted from Observable<ConversationModel> to a suspend fun, backed by the existing ncApiCoroutines.getRoom. A plain try/catch around a suspend call has no     
    window for an exception to bypass error handling, so this removes the bug class rather than papering over one call site.                                                                                                       
  - OfflineFirstConversationsRepository.getRoom: replaced the Observer/Disposable boilerplate and nested runBlocking with a sequential try/catch; on failure (network offline, request fails, or no supported API version) it falls
    back to the locally cached conversation, same as before.                                                                                                                                                                       
  - ConversationInfoViewModel.getRoom: same conversion; removed the now-unused GetRoomObserver inner class.                                                                                                                        
  - NotificationWorker.handleCallPushMessage: converted its Rx subscription to runBlocking { chatNetworkDataSource?.getRoom(...) }, consistent with the rest of that method, which already blocks synchronously.                   
  - Updated the two tests that stubbed getRoom as an Observable to stub it as a suspend function.                                                                                                                                  
                                                                                                                                                                                                                                   
  Follow-up (not addressed here)                                                                                                                                                                                                   
                                                                                                                                                                                                                                   
This PR only makes the failure non-fatal — it doesn't address why a user's capabilities can end up null/empty when getRoom is called (e.g. a freshly added/switched account, a capabilities fetch that failed silently, or a race
  where getRoom fires before the capabilities sync completes). A follow-up should trace when User.capabilities gets populated relative to getRoom being triggered, and either wait for capabilities before resolving the API       
  version or fetch them lazily as a fallback inside getConversationApiVersion.  



related old draft (wont be merged): https://github.com/nextcloud/talk-android/pull/6317



### 🚧 TODO

  - [ ] Follow-up PR: investigate root cause of missing/null capabilities that triggers this fallback path 

### 🏁 Checklist

- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


AndyScherzinger (approved):
👍

